PR #14, opened 2026-08-04
Add Linux (AppImage) packaging target

kalyanar:
The shell packaged only for mac and win, so there was no way to produce a Linux artifact even though the tree already builds and runs there: npm install resolves linux-arm64 for every native optional dep, the xlsx sidecar compiles natively via cargo, and the full unit suite passes.

Adds a linux target alongside the existing two. Four Linux-only details the mac/win config never had to handle:

- executableName: linux derives the binary name from package.json "name", and "@genoffice/shell" sanitizes to the invalid "@genofficeshell", which fails the AppImage build outright. mac/win use productName.
- desktopName + syncDesktopName: otherwise StartupWMClass falls back to productName ("GenOffice") while the window reports WM_CLASS "genoffice", and X11 compares case-sensitively, so the taskbar shows the window unlinked from its launcher entry.
- fileAssociations mimeType: the Linux desktop entry builds its MimeType list from this field and drops associations without it. mac and win ignore the field and key off ext, so the existing targets are unaffected.
- The sidecar is taken from cargo's host-native target/release/, the same source mac uses, so no arch is pinned; electron-builder defaults to the build host's.

AppImage only. deb/rpm additionally require a Debian maintainer identity and homepage in the repo metadata.

Verified on arm64: builds, and the packaged AppImage launches with the aarch64 sidecar bundled at resources/native/xlsx-sidecar.

## Summary

- What changed?
- Why is this change needed?

## Related issue

Closes #

## Validation

- [ ] `npm run format:check`
- [ ] `npm run lint`
- [ ] `npm run typecheck`
- [ ] `npm test`

## Screenshots or recordings

Include before/after evidence for visible changes, or write "Not applicable."

## Contributor checklist

- [ ] The change is focused and does not include unrelated reformatting or refactoring.
- [ ] User-facing strings use the existing i18n resources.
- [ ] File open/save changes include an appropriate round-trip or fidelity test.

### DIFF
--- a/apps/shell/electron-builder.cjs
+++ b/apps/shell/electron-builder.cjs
@@ -82,36 +82,45 @@ const config = {
       to: 'gsk/node_modules/ws',
     },
   ],
+  // `mimeType` is read only by the Linux target, where it becomes the
+  // desktop entry's MimeType= list; associations without it are dropped
+  // there. macOS and Windows ignore the field and key off `ext`.
   fileAssociations: [
     {
       ext: 'docx',
       name: 'Word Document',
       role: 'Editor',
+      mimeType: 'application/vnd.openxmlformats-officedocument.wordprocessingml.document',
     },
     {
       ext: 'xlsx',
       name: 'Excel Workbook',
       role: 'Editor',
+      mimeType: 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
     },
     {
       ext: 'pptx',
       name: 'PowerPoint Presentation',
       role: 'Editor',
+      mimeType: 'application/vnd.openxmlformats-officedocument.presentationml.presentation',
     },
     {
       ext: 'xls',
       name: 'Excel 97-2003 Workbook',
       role: 'Editor',
+      mimeType: 'application/vnd.ms-excel',
     },
     {
       ext: 'csv',
       name: 'CSV Document',
       role: 'Editor',
+      mimeType: 'text/csv',
     },
     {
       ext: 'pdf',
       name: 'PDF Document',
       role: 'Editor',
+      mimeType: 'application/pdf',
     },
   ],
   npmRebuild: false,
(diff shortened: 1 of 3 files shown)


Tsurgcom:
+1

3x3xX3N0N:
Verified on x64 (WSL2 Ubuntu, electron-builder 26.15.3), complementing the arm64 verification above:

- AppImage builds; `genoffice.desktop` comes out with `StartupWMClass=genoffice` matching the executable basename, and the MimeType list carries all six associations.
- `resources/modules/{docs,sheets,slides,pdf}` and `resources/native/xlsx-sidecar` (x86-64 ELF, exec bit preserved) are all bundled; sidecar path matches the runtime lookup in shell/sheets main.
- The packaged app launches and stays up (WSLg).

One adjacent footgun this verification surfaced (pre-existing, not introduced here): when an `extraResources` source is missing — e.g. a module `out/` after a partial `build:all` — electron-builder only logs `file source doesn't exist` and still exits 0. A first run with a broken sheets build produced a working-looking AppImage that was silently missing three of the four modules. Fail-fast guard extending the existing gsk `existsSync` check, in case it's useful: https://github.com/genspark-ai/genoffice/compare/main...3x3xX3N0N:genoffice:fail-fast-missing-modules

nazimsaifullah:
This exactly what i was looking. 
I built this locally in Linux Mint 22.3 - Cinnamon 64-bit and it works perfectly fine.

merrick-2002:
Thanks — the Linux-only details are exactly right, and great cross-verification from @3x3xX3N0N and @nazimsaifullah .

MarkRijckenberg:
Thanks for creating GenOffice. I have been able to get GenOffice running on NixOS unstable by running following bash shell script:
https://gitlab.com/Rijckenberg/shell-scripts/-/raw/master/NixOS/scripts/install-nixos-genoffice.bash?ref_type=heads

PR #20, opened 2026-08-04
fix(shell): relayout active tab after X11 maximize bounds settle

### LINKED ISSUE #15: Linux (X11): maximizing the window leaves the active tab's view at its pre-maximize size
### App
Shell / home screen

### Version and OS
0.1.0 (commit `4da673d4dfa994bd0b4a9bc43430e4a058a17c61`) on Linux Mint 22.2 (X11), Electron 41.7.1

### What happened?

GenOffice isn't officially distributed for Linux, but it builds and runs from source there without any code changes (Rust `xlsx-sidecar` compiles clean, no native deps are platform-locked). While verifying that, I hit a real Linux-only rendering bug.

**Steps to reproduce:**
1. Build and launch the app on Linux (X11): `npm run shell`
2. Open any tab (e.g. a spreadsheet)
3. Maximize the window (double-click the titlebar, or the maximize button)

**Expected:** the tab's `WebContentsView` fills the maximized window.

**Actual:** the view keeps the window's *pre-maximize* size — a blank white strip is left along the right and bottom edges, matching exactly the difference between the old and new window size.

**Root cause:** `TabManager` re-fits the active view's bounds only on the window's `resize` event (`apps/shell/src/main/tab-manager.ts:65`):

```ts
shellWindow.on('resize', () => this.layout())
```

`layout()` reads `shellWindow.getContentBounds()` at that moment. On X11, `resize` fires *before* the window manager has actually applied the new bounds — `getContentBounds()` still returns the old size inside the handler, and there's no follow-up event to correct it. I confirmed this with an isolated Electron repro (below): calling `win.maximize()` fires `resize` while `getContentBounds()` still reports the pre-maximize size; the real size is only available ~1ms later, on the next tick.

This is Linux/X11-specific — macOS and Windows report the updated bounds synchronously inside the `resize` handler, so the bug doesn't reproduce there.

**Fix (happy to open a PR if useful):** lay out once synchronously (keeps the no-flicker behavior on macOS/Windows) and once more via `setImmediate`, once the window manager has applied the real bounds, guarding `layout()` against a destroyed window since the deferred call can land after close:

```ts
shellWindow.on('resize', () => {
  this.layout()
  setImmediate(() => this.layout())
})
```

```ts
layout(): void {
  if (this.shellWindow.isDestroyed()) return
  const active = this.tabs.find((t) => t.id === this.activeId)
  if (active?.view) active.view.setBounds(this.contentBounds())
}
```

Verified this resolves it: after the fix, the grid fills the full maximized width (columns visible up to Q instead of stopping at K), with no blank strip. Added a regression test to `apps/shell/tests/tab-manager.test.ts` that fails without the fix and passes with it; full shell suite (114 tests) still passes.

### Logs / screenshots

Isolated repro confirming the stale-bounds timing on X11:

```
--- calling maximize() ---
[maximize] contentBounds=900x574
[resize]   contentBounds=900x574   <- still stale inside the resize handler
[+1ms]     contentBounds=1920x986  <- real size, one tick later
```

Before/after screenshots (maximized window, spreadsheet tab open):
- Before: grid stops around column K, blank white area fills the rest of the window.
- After: grid fills the full width, columns visible up to Q, no blank area.

edenbuilds:
## Summary

- Fixes the Linux/X11 maximize bug where `TabManager` re-fits the active `WebContentsView` only on `resize`, but X11 fires that event before the window manager applies the new size — so `getContentBounds()` is still the pre-maximize size and a blank strip remains along the right/bottom.
- Layouts once synchronously (unchanged behavior on macOS/Windows) and once more via `setImmediate` after bounds settle; guards deferred `layout()` when the shell window is already destroyed.
- Adds regression tests that fail without the deferred re-layout and cover the destroyed-window guard.

Closes #15

Credit: root cause, repro, and fix shape reported on #15 by @skaka.

## Related issue

#15

## Validation

- [x] `npm run format:check`
- [x] `npm test -w @genoffice/shell` — 116 passed (was 114 + 2 new)
- [ ] `npm run lint` (not re-run full monorepo; change is shell-only)
- [ ] `npm run typecheck` (shell-only change; no new types)

## Screenshots or recordings

Not re-verified on a Linux X11 host in this environment (macOS). Behavior matches the isolated Electron repro and unit regression in #15.

## Contributor checklist

- [x] User-facing strings use the existing i18n resources. (N/A — no UI strings.)
- [x] File open/save changes include an appropriate round-trip or fidelity test. (N/A — shell layout only.)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/apps/shell/src/main/tab-manager.ts
+++ b/apps/shell/src/main/tab-manager.ts
@@ -63,7 +63,15 @@ export class TabManager {
     /** localized placeholder title for a tab that has no file yet */
     private readonly untitledTitleFor?: (kind: TabKind) => string,
   ) {
-    shellWindow.on('resize', () => this.layout())
+    // Layout once synchronously for macOS/Windows (bounds are already correct),
+    // then once more on the next tick. On Linux/X11, `resize` fires before the
+    // window manager applies the new size, so getContentBounds() is still the
+    // pre-maximize size inside the handler and a follow-up layout is required.
+    // See https://github.com/genspark-ai/genoffice/issues/15
+    shellWindow.on('resize', () => {
+      this.layout()
+      setImmediate(() => this.layout())
+    })
   }
 
   private untitled(kind: TabKind, fallback: string): string {
@@ -96,6 +104,8 @@ export class TabManager {
 
   /** re-fit the active tab's view after a window resize */
   layout(): void {
+    // Deferred resize layouts can land after the shell window was closed.
+    if (this.shellWindow.isDestroyed()) return
     const active = this.tabs.find((t) => t.id === this.activeId)
     if (active?.view) active.view.setBounds(this.contentBounds())
   }
--- a/apps/shell/tests/tab-manager.test.ts
+++ b/apps/shell/tests/tab-manager.test.ts
@@ -100,6 +100,7 @@ const WINDOW_HEIGHT = 600
 
 interface FakeShellWindow {
   on: ReturnType<typeof vi.fn>
+  isDestroyed: ReturnType<typeof vi.fn>
   getContentBounds: () => { x: number; y: number; width: number; height: number }
   contentView: {
     addChildView: ReturnType<typeof vi.fn>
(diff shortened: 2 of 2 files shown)


merrick-2002:
Thanks — the analysis matches the code exactly, and the fix shape is right: the deferred re-layout is a no-op on macOS/Windows, and the destroyed-window guard plus regression tests cover the edge cases. Also appreciate the credit to @skaka's investigation on #15.

edenbuilds:
let's make this even better :) loving it

PR #24, opened 2026-08-05
Add ChatGPT Codex workflows across GenOffice

sallliisa:
## Summary

Add ChatGPT Codex as a first-class, account-backed AI provider across GenOffice.

- Share provider, model, reasoning, and service-tier controls across Docs, Sheets, Slides, PDF, and Markdown.
- Keep OAuth credentials in GenOffice-owned main-process storage and fail closed on insecure POSIX credential files.
- Preserve app-owned agents and tools while reusing the common provider transport and UI.

## Problem

GenOffice had no ChatGPT-account-backed provider path, so subscribers could not use Codex without replacing each app's existing AI workflow. The initial Docs integration also left cross-app controls, credential storage hardening, and durable test coverage incomplete.

## UX Flow

```text
Current flow
└─ AI sidebar → key-backed provider only

Desired flow
└─ AI sidebar → select ChatGPT Codex → sign in → choose available model/reasoning/tier → stream the existing app-owned workflow
```

## Solution

Implement GenOffice-owned PKCE OAuth and Codex Responses streaming, with app-owned IPC validation and only redacted account/capability state exposed to renderers. Persist credentials in a dedicated JSON file with atomic `0600` writes; on POSIX, reject symlinks, non-regular files, and group/world-readable files before parsing. Extract shared sidebar controls and suite-global settings synchronization, then retain focused behavioral tests while removing brittle copy and layout assertions.

## Validation

- `npm test`
- `npm run lint` — 0 errors; 10 existing warnings
- `npm run typecheck`
- `npm run licenses`
- `npm run build:all`
- `git diff --check`
- Prettier check across every PR-diff file against `origin/main`

## Related issue

Not applicable.

## Screenshots or recordings

Not applicable — this reuses the existing sidebar layout rather than introducing a visual redesign.

## Contributor checklist

- Focused scope with no unrelated source changes.
- User-facing Codex strings use the shared i18n resources.
- No document open/save path changes.


merrick-2002:
Thanks for the effort here, @sallliisa. This is a substantial piece of work and we read through it carefully.

We are closing it and will not be adding a ChatGPT-account-backed provider. The integration signs in through the Codex CLI OAuth client and calls the `chatgpt.com/backend-api/codex` endpoints while identifying itself as Codex CLI. That routes a user's ChatGPT subscription through a third-party application in a way OpenAI's terms do not permit, and shipping it built in would put users' accounts at risk. That is not something we can carry in the product regardless of implementation quality.

OpenAI models are already available through the OpenAI provider in Settings → AI with an API key, and any OpenAI-compatible endpoint can be reached through the Custom provider slot.

The shared provider-controls refactor across the five apps is a separate topic. If you would like to propose that on its own, please open an issue first so we can agree on scope before code is written.

PR #26, opened 2026-08-05
feat: add BUILD_DIR environment variable to support electron-builder on non-APFS drives

Ac31415:
### Summary

When packaging Electron applications on macOS where the codebase or build directory resides on an exFAT/FAT32 volume (e.g. an external USB drive), `electron-builder` fails because:
1. The macOS kernel filesystem driver automatically generates hidden `._` AppleDouble metadata sidecar files when files are written to non-APFS volumes.
2. `electron-builder` matches `._app.asar` during ASAR header validation and fails with `RangeError: The value of "offset" is out of range`.
3. `electron-builder`'s file copy step fails on `chmod` for `._` sidecar files (`ENOENT`).

### Changes
- Updated `apps/shell/electron-builder.cjs` to allow overriding `directories.output` via `process.env.BUILD_DIR` (e.g., `BUILD_DIR=/tmp/genoffice-release`). This enables staging and packaging intermediate build files on the local APFS drive while keeping the source repository on an external drive.
- Documented usage and workaround in `README.md` and `CONTRIBUTING.md`.

### Verification
- Ran `BUILD_DIR=/tmp/genoffice-release npm run dist:mac` from an exFAT drive on macOS. The `.dmg` and `.zip` installers packaged successfully (exit code 0).
- Ran `npm run format:check` to ensure code formatting compliance.


merrick-2002:
Thanks for tracking this down, @Ac31415. The root cause and the fix are both right: macOS writes `._` AppleDouble sidecars on exFAT/FAT32 volumes and electron-builder trips over them, and letting `directories.output` point at an APFS path is the clean way around it.

We are landing the `BUILD_DIR` override and the CONTRIBUTING notes on `main` with you credited as co-author, so the change is going in. We are closing this PR rather than merging it because the branch also carries a full fork history (about 200 files, including some committed `._*` sidecar files and a very large `.gitignore` change) that we cannot take as-is.

If you hit any other packaging issues on external drives, please open an issue and we will look at it. Appreciate the contribution.

PR #31, opened 2026-08-05
add A2A transport to agent-core

talktogina-developer:
## Summary

- Adds `createA2ATransport` — an `AgentTransport` implementation that routes the GenOffice AI panel to any [A2A-protocol](https://github.com/a2aproject/A2A)-compatible remote agent over HTTP SSE (`tasks/sendSubscribe`)
- The full conversation context (system prompt + message history) is encoded into a single A2A user message per call; streamed artifact text parts map to `onDelta` events; cancellation uses `AbortController`
- Exports `createA2ATransport` and `A2ATransportOptions` from `@genoffice/agent-core` alongside the existing `createIpcTransport`

## Test plan

- `npm run test -w @genoffice/agent-core` — 11 new unit tests (57 total, all pass); covers JSON-RPC wire format, auth/session headers, SSE streaming deltas, status-completion, error paths (HTTP, JSON-RPC, task failed/canceled, network), abort, multi-turn history encoding, and URL normalisation
- `npm run typecheck -w @genoffice/agent-core` — clean
- `npm run format:check` — clean (Prettier)
- `npm run lint` — 0 errors (pre-existing warnings only, none in new files)

### DIFF
--- a/packages/agent-core/src/a2a-transport.ts
+++ b/packages/agent-core/src/a2a-transport.ts
@@ -1,0 +1,273 @@
+import type {
+  AgentMessage,
+  AgentStreamCallbacks,
+  AgentStreamRequest,
+  AgentTransport,
+} from './types'
+
+// ---- A2A protocol wire types (subset used by this transport) ----
+// Spec: https://github.com/a2aproject/A2A
+
+interface A2ATextPart {
+  type: 'text'
+  text: string
+}
+
+interface A2ADataPart {
+  type: 'data'
+  data: Record<string, unknown>
+}
+
+type A2APart = A2ATextPart | A2ADataPart
+
+interface A2AMessage {
+  role: 'user' | 'agent'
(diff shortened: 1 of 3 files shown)


merrick-2002:
Thanks for the contribution, @talktogina-developer. The transport is cleanly written and the test coverage is solid.

We are closing this one for now for two reasons:

1. Nothing in the app can select or configure a remote agent yet, so the transport would land as an exported module with no consumer. Remote-agent support needs a settings surface, credential handling, and a product decision about how it coexists with the built-in provider path, and that design should be discussed in an issue before code lands.
2. The implementation targets the `tasks/sendSubscribe` JSON-RPC method, which the A2A spec has since replaced with `message/stream`. Anything we ship would need to follow the current protocol.

If you are interested in pursuing this, please open an issue describing the intended user flow and we can work out the design together. Happy to revisit a PR on top of that.

PR #41, opened 2026-08-06
feat(shell): dark/light/system theme + performance optimizations

alfirus:
## Theme Support

Adds a theme toggle to the account popup menu with three options:
- **Light**: explicit light theme
- **Dark**: explicit dark theme
- **System**: follows OS prefers-color-scheme

The toggle lives between Language and Update Channel, using the same flyout pattern.

### Changes
- UiTheme type + getTheme()/setTheme() IPC with persistence
- Theme row + flyout in AccountEntry (same UX as Language)
- data-theme attribute applied before first paint (no flash)
- Dark mode CSS variables for all shell surfaces
- 40+ hardcoded color selectors overridden
- Onboarding modal dark mode treatment
- i18n strings in all 19 languages

---

### Cache settings reads
currentUpdateChannel() and currentTheme() now cache like currentLang(). Eliminates disk reads on every IPC call.

### Consolidate CSS dark mode selectors (-300 lines)
Converted 20+ hardcoded colors to CSS custom properties. Removed ~300 lines of duplicated element-level dark mode overrides. Single source of truth for the dark mode palette.

### Fix button hover performance
Replaced filter: brightness() with opacity transitions. Avoids GPU compositing overhead.


merrick-2002:
Thanks @alfirus, great work! 🎉 The implementation is clean and consistent with the existing settings flyout patterns, and we really appreciate the full i18n coverage and the CSS variable consolidation.

Issue #1, opened 2026-08-02
Pre-open-source audit: remaining items before public launch


wwwppp0801:
Pre-release audit of the repo ahead of the public Apache-2.0 launch. Three areas came back clean: **no debug output left in code** (remaining CJK strings are all functional — font tables, numbering, i18n dictionaries), **no secrets / internal hosts / PII** in the working tree or git history, and **all 238 production dependencies are Apache-2.0-compatible** (`tools/check-licenses.mjs` passes; only copyleft touchpoint is `jszip`'s dual license, already pinned to MIT in the notices generator).

The items below are what's left. File/line references are from the current `main` (177dbe3).

## Must fix before going public

- [x] **Add `.github/CODEOWNERS`** — the file doesn't exist, but `CONTRIBUTING.md:145` links to it and `ee/README.md:17-19` claims the `ee/` license boundary is "enforced via CODEOWNERS". Right now nothing protects `ee/` from external PRs. Minimum: a rule covering `ee/` owned by the maintainer team.
- [x] **Fix README.md:28-29 rendering** — "(with `docxIndex` anchors + passthrough)" got wrapped so line 29 starts with `- `, which renders as a nested list item and cuts the sentence in half on the repo landing page.
- [x] **Fix the installer instructions in CONTRIBUTING.md:56-58** — they tell contributors to run `npm run dist:mac -w @genoffice/shell`, which fails on a fresh clone: electron-builder packs `../{docs,sheets,slides,pdf}/out` as extraResources, which only exist after `npm run build:all`, and it also skips `npm run notices` (so THIRD-PARTY-NOTICES.txt is never generated, contradicting the NOTICE file). Point to the root `npm run dist:mac` instead.
- [x] **Enable private vulnerability reporting on this repo** — SECURITY.md and `.github/ISSUE_TEMPLATE/config.yml:4` both link to `security/advisories/new`; that URL 404s unless the feature is turned on in repo settings.

## Recommended — security hardening

These are the first things external security researchers will flag on an Electron app. Mitigating context: every renderer loads local files only, and all 14 `BrowserWindow`s already run with `sandbox: true` + `contextIsolation: true` + `nodeIntegration: false`.

[...]

PR #46, opened 2026-08-07
feat(ai): user-owned AI configuration — custom endpoint, Tavily, proxy, rules/skills, built-in browser

### LINKED ISSUE #45: User-owned AI configuration: custom model endpoint, Tavily search, proxy, agent rules/skills, and a built-in browser
### App

AI panel / providers

### Problem

Everything the AI panel does is currently fixed at build time: which model answers, which search backend it uses, how it reaches the network, and what it knows about how this user works. Five gaps follow from that, and they compound for the same people.

**1. The model cannot be changed.** `packages/ai-provider` already ships a `custom` provider with an OpenAI-compatible client, streaming and tool-calling — but all three main processes discard the stored selection:

```ts
settings.provider = 'genspark'   // docs-main.ts, slides/ai-ipc.ts, sheets-main.ts
```

so a custom endpoint can never take effect, hand-editing `ai-settings.json` included, and no UI exposes the fields. That rules out anyone who cannot create a Genspark account, anyone whose organisation will not let document contents leave for a third party, and anyone running a local model (Ollama, LM Studio, vLLM) to keep documents on the machine.

Underneath it, `temperature` is hardcoded to `0.3` in both request builders. Reasoning models reject *any* temperature but their own — I get `invalid temperature: only 1 is allowed for this model` from a live endpoint — so those models stay unusable even once the provider selection is fixed. The field has to be **omittable**, not merely settable. There is also no way to set a token ceiling or `reasoning_effort`.

**2. Search dies with the account.** `webSearch` / `imageSearch` try gsk first, then Serper, then DuckDuckGo. A user on their own model has no gsk session, and a user in mainland China without a proxy cannot reach any of the three. The AI panel keeps working but silently loses every fact-finding tool.

**3. The proxy cannot be configured.** Each app detects one from `HTTPS_PROXY`/`ALL_PROXY` or the system setting. A packaged app launched from the dock inherits no shell environment, and a user whose system proxy is off but who runs a local client on `127.0.0.1:7897` has no way to say so. There is also no way to say "no proxy" on a machine whose system proxy would otherwise be picked up.

**4. The agent cannot be told how this user works.** There is no place for standing instructions — house style, terminology, "never invent figures" — and no way to give the agent a reusable procedure for a recurring task. Everything has to be re-typed into each conversation.

**5. The agent cannot read a page.** It can search and get titles, links and snippets, but it cannot open a result and read it. Most of what it would be asked to read renders client-side, so a plain fetch returns an empty body.

### Proposed solution

A **Settings** window (sidebar: Model, Network, Rules, Skills, General) replacing the AI entry in the account menu, which had no room left. Everything below is stored once and shared by Docs, Sheets, Slides and PDF, including tabs that are already open.

**Model** — Genspark account (unchanged, still the default) or any OpenAI-compatible endpoint: base URL, model, optional API key (optional because local servers accept anonymous requests). Plus temperature, a max-token ceiling and `reasoning_effort`, where **blank means the field is omitted from the request** — that is the only thing that makes reasoning models work. A never-configured value keeps today's `0.3`, so existing installs see no change. A "Test connection" button dry-runs the endpoint with the same parameters it will run with.

**Network** — a Tavily key, and an explicit proxy (`http://`, `https://`, `socks5://`, or a bare `host:port`). Tavily becomes the preferred search backend when a key is present, because it returns extracted article text rather than snippets. The proxy applies to model calls, search, sign-in and the built-in browser, and wins over both environment variables and the system proxy.

**Rules** — always-on instructions: one global block plus one per surface (docx / pptx / sheets / pdf), concatenated into that app's system prompt.

**Skills** — markdown playbooks stored as real files under `userData/agent-skills/`, so they can be written in the app, dropped in, uploaded, or kept in git and edited outside. Each carries front matter (name, description, scope) and can be scoped to specific apps. **Only titles and descriptions reach the prompt**; a `load_skill` tool fetches a body when a task matches, so a large library does not crowd out the context window.

**Built-in browser** — `browse_page` renders a URL in a hidden, sandboxed Chromium window and returns the text, so JavaScript-rendered pages actually work. It gets its own throwaway session partition (the user's cookies never reach a model-chosen page), is SSRF-guarded, and honours the proxy. `extract_pages` is the cheaper bulk path through Tavily.

### Alternatives considered

- **Environment variables only.** Less discoverable, and useless for an app launched from a dock icon.
- **Hand-editing `ai-settings.json`.** Would need only the three `settings.provider = 'genspark'` lines removed, but leaves the feature undocumented and unvalidated, and does not address the other four gaps.
- **Exposing the whole existing provider matrix** (`anthropic` / `gemini` / `deepseek` / `openai`, already in `AI_PROVIDERS`). More surface to explain; one OpenAI-compatible slot covers those vendors and every self-hosted gateway through a single code path.
- **`fetch` + HTML-to-text instead of a real browser.** Cheaper, but returns an empty body for exactly the pages worth reading.
- **Pasting whole skills into the prompt.** Simple, but the context cost grows with the library, which is why loading is on demand.

### Before I take more of your review time

I have this implemented and tested, and #46 carries it. Two things are worth your opinion before you spend time on the diff:

1. **Is the direction welcome at all?** The README's "no model API key is stored locally" reads like a deliberate product stance, and routing model calls away from Genspark is more than an engineering decision. Happy to drop it if it cuts against where you want the product to go.
2. **One line in it is a judgement call I would not merge on my own authority**: the Model section ends with "GenOffice works best with Kimi K3." It was requested by the person I am building this for, but it is a competitor endorsement in your product's UI. It is isolated in the diff and I will remove it on request — please do not let it colour the other five features.

Screenshots of each section are attached below.
Image
Image
Image
Image
Image

NannaOlympicBroadcast:
## Summary

**What changed?** A **Settings** window (Model, Network, Rules, Skills, General) and the plumbing behind it, so five things that were fixed at build time become the user’s to configure — once, shared by Docs, Sheets, Slides and PDF, including tabs that are already open.

| | |
|---|---|
| **Model** | Genspark account (default, unchanged) or any OpenAI-compatible endpoint: base URL, model, optional API key. Plus temperature, a max-token ceiling and `reasoning_effort`. |
| **Search** | A Tavily key unlocks web search, image search and page extraction, and is preferred over gsk/Serper/DuckDuckGo when present. |
| **Proxy** | Explicit `http` / `https` / `socks5` / `socks4`, or a bare `host:port`. |
| **Rules** | Always-on instructions: one global block plus one per surface. |
| **Skills** | Markdown playbooks with front matter and scope, loaded on demand. |
| **Browser** | `browse_page` renders a URL in a hidden Chromium window; `extract_pages` is the bulk path. |
| **Sessions** | One file can hold several conversations, created and switched from the AI panel. |
| **Model switching** | Several saved endpoints, switched from the AI sidebar without opening Settings. |

**Why is this change needed?** Described in full in #45. The short version is five gaps that compound for the same people:

1. `packages/ai-provider` already ships a working `custom` provider, but all three main processes threw the stored selection away — `settings.provider = ’genspark’` in `docs-main.ts`, `slides/ai-ipc.ts` and `sheets-main.ts` — so no endpoint could ever take effect and no UI exposed the fields. Underneath, `temperature` was hardcoded to `0.3`, which reasoning models reject outright (`invalid temperature: only 1 is allowed for this model` from a live endpoint), so the field has to be **omittable**, not merely settable.
2. Search silently dies for anyone without a gsk session or without reach to gsk/Serper/DuckDuckGo.
3. The proxy could only come from env vars or the system setting — a packaged app launched from the dock sees neither.
4. Nowhere to put standing instructions or a reusable procedure.
5. The agent could search but not read; most pages worth reading render client-side.

### Design notes worth a reviewer’s eye

- **Blank means omit.** For the tuning fields an empty box drops the field from the request. A never-configured value keeps today’s `0.3`, so existing installs are byte-identical in behaviour.
- **Fall back, never break.** A half-filled custom endpoint resolves back to Genspark rather than silently disabling AI. A malformed proxy degrades to “no proxy” rather than poisoning every request.
- **The settings file is the source of truth.** `ai:stream` / `ai:chat` re-read it per request instead of trusting the renderer’s snapshot, which is what lets a change reach already-open tabs.
- **The proxy needs telling three times** — undici for main-process fetch, the Chromium session for renderer/sign-in/agent browser, and the gsk child process via its environment. An explicit setting wins over env and system, including when it is empty.
- **Skills advertise titles only.** Bodies arrive through `load_skill` when a task matches, so a large library does not crowd out the context window. They are real files under `userData/agent-skills/`, so they can be kept in git and edited outside the app.
- **The browser is locked down**: no Node, sandboxed, its own throwaway session partition so the user’s cookies never reach a model-chosen page, SSRF-guarded against loopback and private ranges, navigation pinned, destroyed on every exit path.

Two latent bugs surfaced while wiring this and are fixed here: `composeSkills` flattened tool lists eagerly, so a skill whose tools vary at runtime could never expose them; and the one-shot OpenAI-compatible path never sent `max_tokens`, so “test connection” was not exercising the ceiling it claimed to.

## Since this PR was opened

Feedback from using the branch produced six more commits. All of it builds on the
settings above rather than changing anything already described.

**The configured proxy did not reach the standalone editors.** `applyNetworkSettings` /
`applyProxy` / `bootstrapNetworkSettings` lived in `docs-main` and only the shell called the
bootstrap; `startSlidesStandalone` and `startSheetsStandalone` read env vars and the system
proxy and nothing else. So a proxy set in the new window worked in the shell but not in
standalone slides — where PPT generation runs — and `setGskProxyUrl()` was never called there,
leaving `gskChildEnv()` nothing to forward to the CLI child. The Tavily key had the same hole.
The bootstrap moved to `@genoffice/electron-utils` and all three apps call it first.

**SOCKS could not reach the CLI child.** A child is pointed at a proxy through
`NODE_USE_ENV_PROXY`, which is undici's env proxy support and speaks HTTP `CONNECT` only. A
loopback `CONNECT` server now performs the SOCKS5 handshake on the child's behalf —
`CONNECT`-only, bound to 127.0.0.1, `unref`'d, torn down when the proxy changes, and awaited by
`runGsk` so the first call cannot outrun the bind. socks4 falls back to no proxy rather than a
broken one.

**One file can hold several conversations.** "New chat" only reset React state while the store
kept appending to the same chatId, because `chatIdByPath` mapped a file to exactly one chat.
That key now means "the file's active chat", a new `chatsByPath` records every chat a file has
opened, and a picker in the panel header of docs, slides, sheets and markdown creates and loads
them. `fileRenamed` and `moveFileToProject` carry every chat rather than orphaning the inactive
ones.

**The sidebar switches models.** `providers.custom` stays the endpoint every request builder
reads; `customProfiles` + `activeProfileId` sit beside it and `syncActiveProfile` copies the
selection in on read, so `chatForProvider`, `streamForProvider`, `activeProvider`,
`isCustomConfigured` and all four main processes are untouched and an older settings file still
resolves. A file with a configured `providers.custom` and no library gets that endpoint as
profile #1, still live. `ai:set-active-model` re-reads the file rather than trusting the
renderer's snapshot. The Model section became a list with add / remove / rename.

**The AI can request icon sets and above/below-average conditional formatting.** The serializer
already wrote both and the rule panel already offered them; `cfRuleSchema` exposed neither, so
asking for a traffic-light icon set produced nothing. `aiIconConfigs` emits icon ids `0..n-1` or
fully reversed — the only two orderings the save path accepts — so an AI-built rule cannot fail
closed at save time, and a drift guard keeps the offered set equal to `OOXML_ICON_SETS`.

**Genspark Projects says why a sync failed.** Every cause collapsed into "load failed, try again
later" with a Retry button, including an out-of-credits account where retrying can never
succeed. `syncCloudProjects` now resolves with a classified reason and the cached list instead
of rejecting. The account-switch abort still rejects — it is a control-flow signal, not a
failure to explain.

**`npm run dist:win` works from a clean checkout.** `electron-builder.cjs` pinned the Windows
sidecar to the gnu triple directory, but a host build on an MSVC toolchain lands in
`target/release`; electron-builder only logs a missing `extraResources` source and still exits 0,
so the installer shipped without `xlsx-sidecar.exe` and every workbook open failed at runtime.
The path now resolves to whichever build exists — triple directory first, so a cross-compiling
pipeline is unaffected — and `beforePack` fails the build instead. `tools/cargo.mjs` also finds
cargo when `~/.cargo/bin` is not on PATH, which otherwise killed `npm test` with a message that
never mentioned Rust.

Fixed along the way: `applyModelSettings` rebuilt settings from scratch, so the first save from
the settings window would have deleted every model profile but the live one; and two
pre-existing Windows-only test bugs that stopped `npm test` reaching past the second workspace
(`dialog-memory.test.ts` hardcoded a POSIX path separator, `genoffice-auth.test.ts` asserted
POSIX mode bits).

## Related issue

Closes #45

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`
- [x] `npm run licenses` — CI gates on it although the template does not list it

Also: `npm run build:all` clean, and the e2e suite green except the four pre-existing `sheets-*` specs noted below.

Re-run clean after every follow-up commit, including the Rust `xlsx-engine` suite. The e2e suite
was **not** re-run for the follow-ups. A Windows installer built from this branch with
`npm run dist:win` installs and launches, and now builds with no manual setup step.

- **`npm run format:check` could not run as a script** on this Windows machine: `tools/format-changed.mjs` dies with `spawnSync ...\node_modules\.bin\prettier.cmd EINVAL` before reaching Prettier, with or without `--base`. This is Node’s Windows handling of `.cmd` in `spawnSync`, pre-existing and unrelated. Verified the CI-equivalent way instead — Prettier run directly over every file changed against `upstream/main`, all clean — so the box is ticked on the check, not the wrapper. Worth a glance.
- **Four `sheets-*` e2e specs fail on this machine** (a typed cell value never reaches the saved sheet). I confirmed these are **pre-existing**: with only `apps/sheets/src/renderer/App.tsx` reverted via `git stash`, the failure is byte-identical. Everything else — 16 workspaces typechecked, the Rust sidecar suite, and the other e2e specs — passes.

## Screenshots or recordings

All five sections are screenshotted in #45. They are also reproducible in one command, since the e2e spec captures them itself:

```bash
npx playwright test --config e2e/playwright.config.ts e2e/settings-window.spec.ts e2e/ai-model-settings.spec.ts
# → e2e/artifacts/screenshots/settings-{model,model-custom,network,rules,skills}.png
```

## Contributor checklist

- [x] User-facing strings use the existing i18n resources — 55 new keys across all 19 languages in the shell dictionary, which is compile-time checked, so a missing translation is a type error rather than a runtime fallback.
- [x] File open/save changes include an appropriate round-trip or fidelity test — no docx/xlsx/pptx path is touched. Settings persistence is covered by round-trip unit tests plus e2e assertions on the written `ai-settings.json`, `agent-rules.json` and `agent-skills/*.md`.
- [x] Nothing under `ee/` is modified.

## Notes for review

**Merged `upstream/main`** (through 1a1247a) into the branch: it now merges cleanly, and the settings window was updated to follow the light/dark/system theme that landed in the meantime — three hardcoded colours went through CSS variables, and a `--success` token was added beside the existing `--danger`.

**Tests.** `ai-provider` 111 and `agent-core` 67 unit tests cover provider resolution, the tuning knobs’ omit/send behaviour, proxy normalization, skill front-matter round-tripping and scope filtering. Two e2e specs drive the built Electron app: one against a stub OpenAI-compatible server, one through the settings window to disk. Two assertions carry most of the weight — an `ai:stream` turn issued **with a deliberately stale provider snapshot** still reaches the custom endpoint (that is what proves the open-tabs claim), and a stub model that 400s whenever `temperature` is present reproduces the reasoning-model failure end to end.

**Three things I would like your opinion on:**

1. **The Model section ends with “GenOffice works best with Kimi K3.”** This was requested by the person I am building this for, but it is a competitor endorsement in your product’s UI and I would not merge it on my own authority. It is one line and one i18n key set, isolated in the diff — say the word and I will strip it. Please do not let it colour the rest.
2. **The e2e specs reach handlers through `ipcMain._invokeHandlers`**, a private Electron API. Concise and stable in practice, but if you would rather tests only drive the real renderer path, I will rework them.
3. **The settings window absorbed language, update channel and account** from the account popup, which had no room left. If you would rather those stayed put, that is easy to undo.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


NannaOlympicBroadcast:
Follow-up to the packaging note above — fixed here rather than left for the release pipeline.

**`npm run dist:win` now works from a clean checkout with no manual steps.** Two things were in the way, both only visible when building outside CI.

`electron-builder.cjs` pinned the Windows sidecar to `target/x86_64-pc-windows-gnu/release`, but `npm run native:build` passes no `--target`, so a host build on an MSVC toolchain lands in `target/release`. Since electron-builder only *logs* a missing `extraResources` source and still exits 0, the installer shipped without `xlsx-sidecar.exe` and every workbook open failed at runtime — silently, until someone opened a spreadsheet. The path now resolves to whichever build exists, triple directory first so a cross-compiling pipeline is unaffected, and `beforePack` turns a missing sidecar into a build error. That closes the same gap the config already covers for its other four sources.

Separately, `cargo` is not always on PATH: rustup installs it into `~/.cargo/bin` and adds that to the *user* PATH, which a shell started before the install never picks up. `npm test` and `npm run build:all` then died on `'cargo' is not recognized`, which says nothing about the Rust toolchain being what is missing. `tools/cargo.mjs` probes PATH first — a rustup override or a CI toolchain still wins — then falls back to `CARGO_HOME` and the default location, and explains the fix when cargo genuinely is not installed.

Verified by removing things rather than only by the happy path: with the hand-copied binary deleted the resolution falls back to the host build; `npm test` and `npm run build:all` pass with cargo off PATH; `npm run dist:win` packages the sidecar with no manual step; and deleting the sidecar entirely fails the build with a message naming it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

NannaOlympicBroadcast:
## Summary (this replaces the eight progress comments below with one)

Nine items, twenty commits, branch `feat/custom-llm-provider-settings`. Five were the original scope; four more came from actually using the built app. All verified clean on every push: `npm run lint`, `npm run typecheck` (18 packages), `npm test`, `npm run build:all`, plus a packaged Windows installer (`npm run dist:win`) built and run from this branch.

### 1–5: original scope

1. **Proxy did not reach standalone editors.** Only the shell bootstrapped network settings; standalone slides (where PPT generation runs) and sheets ran their own proxy path that never called `setGskProxyUrl()`, so the gsk CLI child always dialled direct. Bootstrap moved to `@genoffice/electron-utils`; all three apps now call it, precedence settings → env → system.
2. **SOCKS could not reach the gsk CLI child.** A child process is only pointed at a proxy via `NODE_USE_ENV_PROXY`, which is undici's env-proxy support and speaks HTTP `CONNECT` only — so `socks5://` worked in the browser and main process but every CLI call dialled direct. A loopback `CONNECT`→SOCKS5 bridge now fronts it (`CONNECT`-only, bound to 127.0.0.1, `unref`'d, torn down on proxy change).
3. **One chat per file.** "New chat" reset React state while the store kept appending to the same `chatId`. `chatIdByPath` now means "the file's active chat"; a new `chatsByPath` records every chat a file has opened, with `newChatForFile` / `setActiveChatForFile` / `listChatsForFile` driving a picker in the panel header across docs, slides, sheets and markdown.
4. **Switching models meant opening settings**, one endpoint only. `customProfiles` + `activeProfileId` sit beside the existing `providers.custom`, which every request builder still reads unchanged (`syncActiveProfile` copies the active selection into it) — so an older settings file still resolves with no visible change. Settings → Model is now add/remove/rename.
5. **AI could not request icon sets or above/below-average conditional formatting**, though the save path and the manual rule panel both already supported them — `cfRuleSchema` just never exposed them to the model.

### 6–9: found while testing

6. **Genspark Projects gave no reason for a failed sync.** Every cause collapsed into "load failed, try again later" with a Retry button — including an out-of-credits free account, where Retry can never succeed. `syncCloudProjects` now resolves with a classified reason (`credits` / `signedOut` / `network` / `unknown`) instead of rejecting, keeps the cached list on a failed refresh, and swaps Retry for a Genspark link on the credits case.
7. **Google models rejected every tool call from the second round on** (`17a7a01`) — `HTTP 400 … missing a thought_signature`. Confirmed the field's actual location empirically against the live endpoint rather than the docs: `message.tool_calls[].extra_content.google.thought_signature` on the OpenAI-compatible shim, `thoughtSignature` beside `functionCall` on native Gemini. Both rebuild paths had the same hole; fixing only the shim would have moved the failure, not removed it. Captured at all four call-parsing sites, echoed on both rebuilds.
8. **Nothing the agent learned about you survived the conversation** (`deca409`, `54e0447`). `remember` / `forget` tools plus a memory section in the prompt, capped (300 chars/entry, 4000 total, 100 entries) so a chatty agent can't fill the budget. The trust boundary is explicit: the agent writes these from conversation, which includes whatever document is open, so a document can try to talk its way into permanent memory. The tool description says not to — but guidance is not a control; the list being visible and deletable under Settings → Rules is what actually limits the damage.
9. **The agent could not see what it was laying out** (`acb2c85`, `ed06e44`, `65082a8`). `view_page` renders the document and returns a PNG. Images ride a follow-up user turn rather than the tool result, because a tool result is text on every provider wire format this app supports (Anthropic allows image blocks in a `tool_result`, but OpenAI-compatible's `tool` message is a plain string and Gemini's `functionResponse` is a JSON struct — no shape works on all three; a user turn does). Slides renders any requested slide off-screen via the same path PPTX export uses, so it reaches slide 9 while slide 2 is on screen without touching the user's viewport; docs captures its current view only, since `.doc-page` is one continuous editor node with no page N in the DOM to target. That capability difference is expressed in the tool schema (`viewPageCount` gates whether `page` is offered at all), not papered over.

### Corrected along the way, on the record

- **A proxy race I reported and then disproved.** Suspected `applyProxy`'s fire-and-forget tail could undo the env/system fallback proxy. Could not write a failing test; the interleaving that would cause it cannot actually occur, because `applyProxy`'s first `await` always resolves before the fallback's dispatcher branch runs. No code change. What's real is only a log-ordering cosmetic (`[proxy] direct` can print after the fallback's line).
- **`view_page` shipped first as a screenshot, corrected to a renderer for slides.** The initial cut used `capturePage` for both apps and dropped the `page` argument because the renderer adapters had no navigation handle. True for docs; wrong for slides, which already had an off-screen deck renderer in the same file. Slides now renders directly instead of screenshotting — reaches any slide, never moves the viewport, shows the page as it will print. The `page` argument came back, but only where the bridge can honour it.
- **Packaging gap, fixed rather than left for the release pipeline:** `electron-builder.cjs` pinned the Windows sidecar path to a cross-compile target directory; a host MSVC build lands elsewhere, and electron-builder exits 0 on a missing `extraResources` source — so an installer could silently ship without `xlsx-sidecar.exe`. Path resolution now checks both locations and `beforePack` fails the build if neither has it. Also added `tools/cargo.mjs` so `cargo` off PATH (the common rustup state) fails with an actionable message instead of `'cargo' is not recognized`.
- **A pre-existing gap surfaced by the memory/view_page work:** standalone slides and sheets (not the shell) had none of the seven web-skill IPC channels registered, so browsing, skills, memory and page capture all silently failed there. Fixed in `ed06e44` by moving registration into a shared `registerAgentToolIpc` in `electron-utils`. Still open, deliberately not folded in here: pdf and markdown standalone register no `ai:*` channels at all.

The three questions in the original PR description are still open and unaffected by any of the above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

NannaOlympicBroadcast:
@merrick-2002 这个 PR（#46）现已完成九项修复，五关验证均通过，想请你空时审核一下。

merrick-2002:
Thanks @NannaOlympicBroadcast for the substantial work here, and apologies for taking this long to respond.

We're closing this PR, mostly because the landscape has moved since it was opened:

- Custom endpoints are now covered by the built-in BYOK settings (Settings → AI Model), with a Custom slot for any OpenAI-compatible server.
- Custom models for the built-in image and tool paths remain tracked in #45, which stays open.
- The app already honors `HTTPS_PROXY` / `HTTP_PROXY` / `ALL_PROXY` and falls back to the system proxy for main-process requests.
- Tavily search, always-on rules, markdown skills and a built-in browser are each sizeable product features. We'd want a separate issue and design discussion for any of them before code, rather than taking them in one 107-file change.

If one of those pieces matters most to you, please open a focused issue for it and we'll take it from there. Thanks again for pushing on this.

Issue #36, opened 2026-08-06
Support .doc


MoonMao42:
### App

GenOffice Docs (docs)

### Problem

please

lucamosca1:
But also older office files such as xls and ppt ... and the whole open document format ...

merrick-2002:
Closing in favour of #579, which describes the same request (native .doc import with formatting-preserving conversion) with concrete scope. Please follow along there.

PR #50, opened 2026-08-07
feat(doc): Support opening legacy .doc files as plain text

alfirus:
Implements hybrid approach for .doc file support, closing #36. Adds word-extractor for text extraction, shows warning dialog, opens in docs editor. User can save as .docx for full editing.


merrick-2002:
Thanks for taking a shot at #36 — the effort is appreciated, and `word-extractor` is a reasonable pure-JS choice for text extraction (we verified it handles Word 97–2003 files fine).

That said, we've decided not to take the plain-text route for .doc files. Legacy .doc documents are typically formatting-heavy (forms, tables, contracts), and opening them as flattened text tends to read as "broken" rather than "degraded". We'd rather keep the explicit "format not supported" dialog than ship a half-open experience.

Two notes on the implementation as it stands, for reference:

- The extracted text is written to a temp `.txt` and handed to the docs editor, but the docs loader only accepts `.docx`, so the tab comes up empty. Edits would also be saved into the temp directory rather than anywhere the user expects.
- The branch carries ~9,400 lines of deletions that revert the current markdown app and other recent changes on `main`, which makes it unreviewable as-is. For future contributions, please branch from a clean `genspark-ai:main` and keep the diff scoped to a single feature.

We're going to pass on this one. Thanks again for contributing!

PR #51, opened 2026-08-07
fix(file-parse): keep xlsx and pptx attachment text verbatim

Waiel5:
## Summary

**What changed?**

`packages/file-parse` builds its two `fast-xml-parser` instances without `trimValues: false` and `parseTagValue: false`, so text pulled out of an xlsx or pptx attachment is trimmed and numerically coerced before it reaches the model. This sets both options on those parsers, matching `packages/docx-engine/src/xml-utils.ts:6` and `packages/pptx-engine/src/parse.ts:55`, which already set them for exactly this reason, and adjusts the two read sites that were relying on the parser to normalise for them.

`parseFileToText` is the attachment path into the AI panel in docs, sheets and slides. Two fast-xml-parser defaults corrupt the text on the way.

`parseTagValue` (default `true`) arithmetically evaluates anything number-shaped:

| in the file | reached the model as |
| --- | --- |
| `<t>02139</t>` | `2139` |
| `<is><t>0x1A</t></is>` | `26` |
| `<t>+15551234567</t>` | `15551234567` |
| `<a:t>1e3</a:t>` | `1000` |

`trimValues` (default `true`) is the larger problem: it strips the spaces carried by `xml:space="preserve"` runs, which is where PowerPoint and Excel put the space whenever formatting splits a sentence into separate runs. A single bold word is enough:

```xml
<a:r><a:t xml:space="preserve">Quarterly </a:t></a:r>
<a:r><a:rPr b="1"/><a:t>revenue</a:t></a:r>
<a:r><a:t xml:space="preserve"> grew </a:t></a:r>
<a:r><a:rPr b="1"/><a:t>12%</a:t></a:r>
<a:r><a:t xml:space="preserve"> in </a:t></a:r>
<a:r><a:t>Q3</a:t></a:r>
```

before: `Quarterlyrevenuegrew12%inQ3`
after: `Quarterly revenue grew 12% in Q3`

The same happens to xlsx rich-text cells, where `Total ` + `due` + ` today` arrived as `Totalduetoday`. So the model was receiving attachment prose with its word boundaries removed, which is not an edge case: it is any deck or sheet with mixed formatting inside a paragraph or cell. `packages/pptx-engine/src/parse.ts:58-61` already carries the rationale in a comment; `file-parse` is the one text-reading package that missed it.

**Why two other things change with it**

`trimValues` is per parser, not per element, so turning it off also stops trimming things that are not user text. Two places were quietly relying on it:

- **Attributes.** `r:id`, `Id`, `Target` and the cell ref `r` are structural, and their schema types collapse whitespace (`Target` is `xsd:anyURI`). Untrimmed, a conforming `Target=" worksheets/sheet1.xml "` builds the wrong ZIP path and the sheet is skipped, so a whole sheet silently disappears from the attachment. The xlsx parser therefore keeps trimming attributes via `attributeValueProcessor`, and the fixture now carries a padded `Target` so removing that line fails three tests.
- **`collectText`.** It pushed every primitive it reached, not only `a:t` content, so an `<a:br>` written as an open/close pair contributed its formatting whitespace to the paragraph. It now pushes a primitive only when that primitive came from an `a:t`, which is what its docstring always claimed.

`<v>` itself is left alone. It is `ST_Xstring`, so it reaches the caller verbatim for every cell type, the same way the sheets xlsx sidecar reads it (`apps/sheets/native/xlsx-engine/src/lib.rs:2314`). The one exception is the `t="b"` comparison, which needs the scalar and emits `TRUE`/`FALSE` rather than whitespace; `t="s"` needs no change because `Number()` already tolerates it.

Verified by building the `origin/main` modules from `git show` and diffing real output across every shape I could construct: `t` in `{b, s, str, e, n, d, inlineStr, absent}`, each plain, padded, newline-wrapped and CDATA-wrapped; empty and missing `<v>`; float and big-integer values; padded `Target`, `Id`, `r:id`, sheet `name` and cell `r`; self-closing and expanded `<a:br>`; CJK and emoji runs. Every one is identical to `main` except the `<t>` and `<a:t>` fidelity fixes, plus whitespace inside a `<v>` now being preserved rather than trimmed, which is the same `ST_Xstring` rule.

Three of the four new tests fail against `main`. The fourth, the `<a:br>` one, fails if the parser options are applied without the `collectText` guard.

One hunk in `xlsx.ts` is not mine: the `for (const row of ...)` line is 103 characters on `main`, over the 100 `printWidth`, so `npm run format:check` wraps it once the file is part of a change. That is the incremental formatting `CONTRIBUTING.md` describes, not a reformat I chose.

## Related issue

None. Found while reading the attachment parse path.

## Validation

Run from a clean tree rebased on `origin/main`.

- [x] `npm run format:check -- --base origin/main`: `All matched files use Prettier code style!`
- [x] `npm run lint`: `7 problems (0 errors, 7 warnings)`, all pre-existing
- [x] `npm run typecheck`: clean across every workspace
- [x] `npm test`: exits 0, no failures, including the Rust sidecar tests (`52 passed; 0 failed`). `@genoffice/file-parse` goes 21 to 25 tests.
- [x] `npm run licenses`: `All production npm dependency licenses are within the allowlist.`
- [x] `npm run fixtures && git diff --exit-code -- fixtures/generated`: no drift
- [x] `npm run fixtures -w @genoffice/sheets && npm run compat -w @genoffice/sheets`: every fixture `"passed": true`, no unexpected changes
- [x] `npm run build:all && npm run test:e2e`: 22 Playwright tests passed

## Screenshots or recordings

Not applicable. Attachment text extraction, no UI surface.

## Contributor checklist

- [x] User-facing strings use the existing i18n resources. (Not applicable, no UI strings.)
- [x] File open/save changes include an appropriate round-trip or fidelity test. (Not an open/save path: `file-parse` is read-only text extraction and never writes a document back. The four new tests are the fidelity equivalent, asserting the extracted text matches the source runs verbatim.)

### DIFF
--- a/packages/file-parse/src/pptx.ts
+++ b/packages/file-parse/src/pptx.ts
@@ -1,7 +1,9 @@
 import JSZip from 'jszip'
 import { XMLParser } from 'fast-xml-parser'
 
-const parser = new XMLParser({ ignoreAttributes: true })
+// Text fidelity: no trim (xml:space="preserve" runs carry the spaces between words),
+// no numeric coercion of tag values (otherwise <a:t>02139</a:t> becomes a number and loses characters)
+const parser = new XMLParser({ ignoreAttributes: true, trimValues: false, parseTagValue: false })
 
 function slideNumber(path: string): number {
   const m = /slide(\d+)\.xml$/.exec(path)
@@ -9,19 +11,20 @@ function slideNumber(path: string): number {
 }
 
 /** collect the text of all a:t descendants of one node */
-function collectText(node: unknown, out: string[]): void {
+function collectText(node: unknown, out: string[], isText = false): void {
   if (node == null) return
   if (typeof node === 'string' || typeof node === 'number') {
-    out.push(String(node))
+    // untrimmed, a non-a:t element written across lines carries its whitespace as a value
+    if (isText) out.push(String(node))
     return
   }
   if (Array.isArray(node)) {
-    for (const item of node) collectText(item, out)
+    for (const item of node) collectText(item, out, isText)
     return
   }
   if (typeof node === 'object') {
     for (const [key, value] of Object.entries(node)) {
-      if (key === 'a:t') collectText(value, out)
+      if (key === 'a:t') collectText(value, out, true)
       else if (typeof value === 'object') collectText(value, out)
     }
   }
--- a/packages/file-parse/src/xlsx.ts
+++ b/packages/file-parse/src/xlsx.ts
@@ -1,7 +1,18 @@
 import JSZip from 'jszip'
 import { XMLParser } from 'fast-xml-parser'
 
-const parser = new XMLParser({ ignoreAttributes: false, attributeNamePrefix: '@_' })
+// Text fidelity: no trim (xml:space="preserve" runs carry the spaces between words),
+// no numeric coercion of tag values (otherwise <t>02139</t> becomes a number and loses characters)
+const parser = new XMLParser({
+  ignoreAttributes: false,
+  attributeNamePrefix: '@_',
+  trimValues: false,
+  parseTagValue: false,
+  // trimValues also governs attributes; nothing read here (r:id, Target, cell ref, sheet
+  // name) carries meaningful edge whitespace, and an untrimmed Target builds the wrong zip
+  // path, which silently drops the whole sheet
+  attributeValueProcessor: (_name, value) => value.trim(),
+})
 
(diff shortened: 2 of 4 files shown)


merrick-2002:
Merged — thank you! The lost word boundaries in xlsx/pptx attachment text were quietly hurting AI answer quality across all three editors, and the level of verification here (diffing real output against main across every cell-type shape) made this an easy review. Great catch.

PR #56, opened 2026-08-07
Add TinyFish web fetch to the docs AI agent

rhea-pandya:
## What

Adds a `fetch_url` tool to the docs AI agent that returns a web page's full readable text, via TinyFish's hosted MCP server.

Today the agent can `web_search`, but it only gets titles/links/snippets back — it can't read the pages it finds. `fetch_url` closes that gap: search finds a page, fetch reads it, and the model can write the real content into the document.

## How

- **`packages/ai-search`** — a `fetchUrl` worker and a `tinyfish-mcp` client. The client connects to `agent.tinyfish.ai/mcp` and calls its `fetch_content` tool. Auth is browser OAuth (authorization code + PKCE + dynamic client registration) via the official MCP SDK's `OAuthClientProvider`; tokens are stored under `~/.genoffice/` (mirroring `genoffice-auth.ts`), so **no API key is stored locally** — the user signs in once. A loopback listener catches the OAuth redirect.
- **`apps/docs`** — the `fetch_url` tool definition + async dispatch (mirrors `web_search`), a "Connect TinyFish" control in the AI panel header with status polling, and the renderer → preload → main IPC wiring plus handlers (`ai:tinyfish-signin` / `-status` / `-signout`), mirroring the existing `gsk-login` / `gsk-status` handlers.
- **Safety** — the model-supplied URL is validated with the existing `isSafeRemoteUrl` SSRF guard before use.
- **i18n** — new keys across all 19 locales.

## Testing

- Type-checks clean; unit tests for the SSRF gate and the connect-first path pass.
- Verified end-to-end against the live TinyFish MCP server: sign-in (OAuth loopback) completes, and `fetch_content` returns clean page text.

## Notes / follow-ups

- Translations are machine-assisted — happy to have your team refine the wording.
- Adds one runtime dependency: `@modelcontextprotocol/sdk` (plus its peer `zod`).


merrick-2002:
Thanks for the well-structured PR. Reading the full text of a page the agent found is a real gap, and the search-then-fetch flow you describe is the right shape.

We're closing this one, though, because it ties the tool to a specific hosted third-party service (TinyFish's MCP endpoint with browser OAuth). The URLs a user asks about and the fetched content would be routed through that service by default, and we don't add default dependencies on individual third-party gateways in the core agent.

If you'd like to pursue a `fetch_url` tool, please open an issue so we can discuss providing it through the existing search backend. Thanks again.

PR #60, opened 2026-08-08
fix(sheets): allow Save As without edits

Jabberwocky238:
## Thinking

Save As should be independent of whether the workbook has pending edits. A clean workbook may have no mutations to apply, but the user should still be able to save the current workbook to a new path. Previously, Sheets treated “no edits” as “nothing to save” and returned before opening the Save As dialog. This change allows empty Save As requests to pass through the existing XLSX save pipeline, while ordinary Save remains a no-op for unchanged workbooks.

## Summary

- make explicit Save As independent of whether the workbook has pending edits
- allow an empty Save As request through the preload boundary
- let the existing XLSX save pipeline write an unchanged workbook to the selected path

Docs and Slides already perform explicit Save As regardless of dirty state, so they require no code changes.

## Verification

- `npm run typecheck -w @genoffice/sheets`
- `npx eslint apps/sheets/src/preload/index.ts apps/sheets/src/renderer/save-actions.ts apps/sheets/src/gateway/xlsx-gateway.ts`

### DIFF
--- a/apps/sheets/src/gateway/xlsx-gateway.ts
+++ b/apps/sheets/src/gateway/xlsx-gateway.ts
@@ -584,29 +584,6 @@ export async function planCellEditsToXlsx(
   sparklineAdditions: readonly SheetSparklineAddition[] = [],
   formulaValues: readonly SheetFormulaValues[] = [],
 ): Promise<MutationPlan> {
-  if (
-    edits.length === 0 &&
-    structuralOps.length === 0 &&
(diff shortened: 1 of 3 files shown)


merrick-2002:
Merged — thank you! Save As on a clean workbook is exactly the kind of paper cut that's easy to miss; nice that Docs and Slides already behaved correctly so this brings Sheets in line.

Issue #13, opened 2026-08-04
Comprehensive Right-to-Left (RTL) Support for Arabic Across the Entire Office Suite


Wyetgf:
### App

Whole suite

### Problem

The current editing experience for Arabic is not practical for daily use because right-to-left (RTL) languages are only partially supported, or not supported at all.

The most important issues are:

1. There is no option to switch document writing direction to Right-to-Left (RTL).
2. There is no setting to make RTL the default writing direction for new documents.
3. There is no option to make right alignment the default paragraph alignment.
4. The AI sidebar always behaves as a left-to-right interface, even when writing Arabic. Text input, cursor movement, alignment, and generated responses should automatically follow RTL conventions whenever the user’s language is Arabic.
5. Lists, indentation, numbering, and bullet points do not follow RTL behavior correctly.
6. Tables do not provide RTL layout support (column order, cell alignment, and text flow).
7. Headers, footers, comments, and text boxes should fully support RTL.
8. Keyboard shortcuts and formatting behavior should work naturally for RTL documents.
9. Mixed Arabic and English text should be handled correctly without breaking punctuation, numbers, or cursor movement.
10. Overall, creating professional Arabic documents is currently difficult compared to Microsoft Office or LibreOffice.

Arabic is one of the world’s most widely used languages, and complete RTL support is essential for users in the Middle East and many other regions.

### Proposed solution

Implement full RTL support consistently across the entire office suite.

Suggested improvements include:

* Add a Writing Direction button (LTR / RTL) in the toolbar.
* Add an application setting to make RTL the default for new documents.
* Add an option to make right alignment the default paragraph alignment.
* Automatically detect Arabic, Persian, Hebrew, and Urdu text and switch writing direction intelligently.
* Ensure all formatting features (lists, numbering, indentation, tables, comments, text boxes, headers, footers, page layout, etc.) fully support RTL.
* Make the AI sidebar automatically adapt to RTL:
    * Right-aligned text.
    * RTL cursor behavior.
    * RTL prompt input.
    * RTL formatting in generated responses.
    * Mirror the sidebar layout when the interface language is RTL.
* Preserve correct bidirectional text rendering for documents containing both Arabic and English.
* Provide full RTL support across Documents, Presentations, and Spreadsheets for a consistent user experience.

This would significantly improve usability for millions of RTL-language users and make the suite a viable alternative to Microsoft Office for Arabic-speaking professionals, students, and organizations.

[...]

PR #62, opened 2026-08-08
fix(docx): keep header and footer paragraphs right-to-left on save

Waiel5:
## Summary

**What changed?**

`headerFooterPartXml` hand-builds its paragraph properties and emits only `w:jc`, so saving a document with a right-to-left header rewrites that header as left-to-right. It now builds them with `mergePPrFormat`, the same model the rest of the engine uses.

Two faults compound in `packages/docx-engine/src/patch.ts`:

```js
const jc = para.align
  ? `<w:pPr><w:jc w:val="${para.align === 'justify' ? 'both' : para.align}"/></w:pPr>`
  : ''
```

1. **`w:bidi` is dropped.** The only child ever emitted is `w:jc`. `HfParagraph extends ParaFormat`, so the parsed header carries `bidi`, `spacing`, `indent`, `borders` and `tabStops`, and all of them are discarded on save.
2. **The alignment is written back unswapped.** `parse.ts:1321` converts Word's logical `w:jc` to a visual direction on read for bidi paragraphs ("Word quirk: in bidi paragraphs w:jc left/right are logical values (start/end); convert to visual direction"). `para.align` therefore holds the *visual* value, and writing it straight out inverts it.

Together:

```
before  <w:pPr><w:bidi/><w:jc w:val="right"/></w:pPr>
after   <w:pPr><w:jc w:val="left"/></w:pPr>
```

This is invisible in the app and wrong the moment Word reopens the file. An Arabic header containing a Latin word, a number, or a page field loses its paragraph direction, so the neutral characters around those runs resolve to the other end of the line. The body path already handles all of this correctly; only headers and footers are affected.

**The part worth reviewing**

Routing headers through `mergePPrFormat` makes this its first caller outside the tests, and that exposed a latent defect in the function itself. Its fresh-build branch, the one `'<w:pPr/>'` takes, joined the model's children in emission order:

```js
// '<w:pPr/>' or unrecognized: rebuild from the format model alone
return fresh.length > 0 ? `<w:pPr>${fresh.map((c) => c.xml).join('')}</w:pPr>` : ''
```

Every other path sorts. `generateParagraphXml` sorts its own children by `PPR_CHILD_ORDER`, and so does the interleaving branch of `mergePPrFormat` itself. `formatPPrChildren` emits by concern, not in schema order, so the unsorted branch produces invalid `CT_PPr`.

I enumerated every combination of the fourteen `ParaFormat` fields `formatPPrChildren` reads and checked child order on each:

| | combinations | wrong `CT_PPr` order |
| --- | --- | --- |
| before | 16384 | **8168** |
| after | 16384 | **0** |

The smallest failing case is `align` + `tabStops`:

```
before  <w:pPr><w:jc w:val="right"/><w:tabs>...</w:tabs></w:pPr>     ranks 26, 10
after   <w:pPr><w:tabs>...</w:tabs><w:jc w:val="right"/></w:pPr>     ranks 10, 26
```

Wrong `CT_PPr` order is what makes Word show the repair dialog, which `schema-order.test.ts` already exists to prevent. Its existing cases all pass a real `<w:pPr>...</w:pPr>`, so they never reached the self-closing branch. Sorting there is a prerequisite of this change rather than an unrelated fix, which is why it is in the same commit.

## Related issue

Partly #13 (RTL support). This is the fidelity half: the engine already parses and renders `w:bidi` correctly, so the gap was purely on the header/footer write path.

## Validation

- [x] `npm test`: exits 0. `packages/docx-engine` 449 passed, 1 skipped, 54 files
- [x] Three new cases fail against `main` and pass here, verified by swapping in `origin/main`'s `patch.ts` and `generate.ts` and re-running:
  - `keeps an RTL header right-to-left and keeps its alignment side` (round trip through `saveDocx` then `parseDocx`)
  - `tabs + bidi keeps CT_PPr child order when built from the model alone`
  - `framePr + tabs + bidi + align keeps CT_PPr child order when built from the model alone`

The RTL test is a real round trip rather than a string assertion: it saves a header with `bidi: true, align: 'left'` and reparses, so it fails if either the direction or the alignment side is lost.

## Screenshots or recordings

Not applicable, the defect is in the saved XML rather than on screen. The reproduction is: open a document with an RTL header, edit anything, save, reopen in Word.

## Contributor checklist

### DIFF
--- a/packages/docx-engine/src/generate.ts
+++ b/packages/docx-engine/src/generate.ts
@@ -1145,8 +1145,12 @@ export function mergePPrFormat(rawPPr: string, format: ParaFormat | undefined): 
   const open = /^<w:pPr(?: [^>]*)?>/.exec(rawPPr)?.[0]
   const fresh = formatPPrChildren(format)
   if (!open) {
-    // '<w:pPr/>' or unrecognized: rebuild from the format model alone
-    return fresh.length > 0 ? `<w:pPr>${fresh.map((c) => c.xml).join('')}</w:pPr>` : ''
+    // '<w:pPr/>' or unrecognized: rebuild from the format model alone, in schema order
+    // (formatPPrChildren emits by concern, and CT_PPr order is not optional)
+    const sorted = [...fresh].sort(
+      (a, b) => PPR_CHILD_ORDER.indexOf(a.name) - PPR_CHILD_ORDER.indexOf(b.name),
+    )
+    return sorted.length > 0 ? `<w:pPr>${sorted.map((c) => c.xml).join('')}</w:pPr>` : ''
   }
   const inner = rawPPr.slice(open.length, rawPPr.length - '</w:pPr>'.length)
   // build set of managed tags for this format (base + conditional)
--- a/packages/docx-engine/src/patch.ts
+++ b/packages/docx-engine/src/patch.ts
@@ -1,5 +1,11 @@
 import JSZip from 'jszip'
-import { applyImageWrap, generateParagraphXml, inlineRunsXml, splitXmlChildren } from './generate'
+import {
+  applyImageWrap,
+  generateParagraphXml,
+  inlineRunsXml,
+  mergePPrFormat,
+  splitXmlChildren,
+} from './generate'
 import {
   NOTE_CONTENT_TYPE,
   NOTE_PART_PATH,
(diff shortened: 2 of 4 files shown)


merrick-2002:
Merged — thank you! Great catch on both the dropped w:bidi and the visual/logical alignment swap, and the CT_PPr child-order fix in mergePPrFormat's fresh-build branch was a nice bonus. The round-trip test coverage is much appreciated.

PR #63, opened 2026-08-08
fix(docx): keep the original root namespaces when regenerating a part

Waiel5:
## Summary

**What changed?**

`word/comments.xml` and `word/footnotes.xml` / `word/endnotes.xml` are regenerated by splicing unchanged entries back in as their **original bytes**, then wrapping them in a root with a hardcoded namespace list. The bytes and the root disagree, so the part can reference prefixes the root never declares. They now reuse the original part's root attributes.

Two hardcoded lists, each missing a prefix the spliced bytes routinely use:

```js
// patch.ts, buildCommentsXml: declares w and w14
'<w:comments xmlns:w="...wordprocessingml/2006/main" xmlns:w14="...word/2010/wordml">'

// notes.ts, NOTE_NS: declares w and r
const NOTE_NS = 'xmlns:w="...wordprocessingml/2006/main" xmlns:r="...relationships"'
```

- A preserved comment holding `<w:hyperlink r:id="rId7">` leaves **`r` unbound**, because `w:comments` never declares it.
- A preserved paragraph holding `w14:paraId` leaves **`w14` unbound**, because `NOTE_NS` never declares it. Word writes `w14:paraId` and `w14:textId` on essentially every paragraph it authors, so this is the common case, not an exotic one.

An unbound prefix is not well-formed XML. Word reports it as unreadable content and offers to repair the file. Reaching it takes one ordinary action: add a comment to a document that already has a linked comment, or add a footnote to any Word-authored document.

**Why the existing tests do not catch it**

`fast-xml-parser`'s `XMLValidator` is not namespace-aware. It reports all three of these as valid:

| emitted part | `XMLValidator` | namespace-aware parse |
| --- | --- | --- |
| `comments.xml` with a preserved `r:id` hyperlink | VALID | **unbound prefix: `r`** |
| `footnotes.xml` with a preserved `w14:paraId` | VALID | **unbound prefix: `w14`** |
| control: same footnote with `w14` declared | VALID | ok |

So every test that validates emitted XML through the parser the repo already depends on will keep passing while the file is broken. That is the reason the new test does its own prefix-binding check instead of calling the validator, and why it starts with a case asserting the *checker itself* fails on a known-unbound prefix. A regression test that cannot fail is worse than none.

**The fix**

One helper, used at both sites:

```js
export function rootAttributes(originalXml, rootTag, fallback) {
  const attrs = originalXml
    ? new RegExp(`<${rootTag}\\b([^>]*?)/?>`).exec(originalXml)?.[1]?.trim()
    : undefined
  return attrs && attrs.includes('xmlns:') ? attrs : fallback
}
```

Carrying the original root forward is the conservative option: the bytes underneath came from that root, so whatever they use is declared. The literal list is kept for parts created from scratch, and the `includes('xmlns:')` guard means a malformed or namespace-free root falls back rather than producing a root with no declarations at all.

## Related issue

None. Found while checking what else regenerates a part by splicing original bytes.

## Validation

- [x] `npm test`: exits 0. `packages/docx-engine` 449 passed, 1 skipped, 55 files
- [x] `keeps the original root attributes when a part is regenerated` fails against `main` and passes here, verified by restoring `main`'s two call sites and re-running
- [x] The other three cases guard the checker and the no-original fallback, and pass either way by design

## Screenshots or recordings

Not applicable, the defect is in the saved XML. The reproduction is: open a Word-authored document, add a footnote, save, reopen in Word.

## Contributor checklist

### DIFF
--- a/packages/docx-engine/src/notes.ts
+++ b/packages/docx-engine/src/notes.ts
@@ -34,6 +34,22 @@ const NOTE_NS =
   'xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main" ' +
   'xmlns:r="http://schemas.openxmlformats.org/officeDocument/2006/relationships"'
 
+/**
+ * Root attributes for a regenerated part. Entries are spliced back in as original bytes, so
+ * the root has to keep declaring whatever prefixes those bytes use: Word puts w14:paraId on
+ * every paragraph it writes, and a literal namespace list leaves that prefix unbound.
+ */
+export function rootAttributes(
+  originalXml: string | null,
+  rootTag: string,
+  fallback: string,
+): string {
+  const attrs = originalXml
+    ? new RegExp(`<${rootTag}\\b([^>]*?)/?>`).exec(originalXml)?.[1]?.trim()
+    : undefined
+  return attrs && attrs.includes('xmlns:') ? attrs : fallback
+}
+
 /** real notes (separator entries excluded), in file order */
 export function parseNotesXml(xml: string, kind: NoteKind): NoteInfo[] {
   return noteEntriesOf(xml, kind).map(({ id, text, xml: entryXml }) => {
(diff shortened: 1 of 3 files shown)


merrick-2002:
Merged — thank you! Excellent find, and the namespace-aware prefix-binding test is a great addition given that XMLValidator can't catch this class of bug. Carrying the original root attributes forward is exactly the right conservative fix.

PR #66, opened 2026-08-09
fix(file-parse): keep pptx line breaks and fields in document order

Waiel5:
## Summary

**What changed?**

`pptxToText` grouped a paragraph's children by tag name, which destroys their document order, so a deck's soft line breaks vanished and field text was appended after all the plain runs. It now parses with `preserveOrder` and walks the children in sequence.

`<a:br>` and `<a:fld>` are siblings of `<a:r>` inside an `<a:p>`. `fast-xml-parser` returns `{"a:r": [...], "a:br": [...]}`, so the collector sees every run first and every break afterwards:

| slide XML | before | after |
| --- | --- | --- |
| `Q3 results<a:br/>up 12%` | `Q3 resultsup 12%` | `Q3 results` ⏎ `up 12%` |
| `Page <a:fld>7</a:fld> of 20` | `Page  of 207` | `Page 7 of 20` |

Both go into the AI panel as attachment context, so a slide-number field lands in the wrong place in the text the model reads, and two lines of a title merge into one word.

**The part worth reviewing**

My first attempt rewrote `<a:br>` and `<a:fld>` into `<a:r>` with a string replace before parsing, mirroring what `packages/pptx-engine/src/parse.ts` does. That is a regression, and it is worth recording why so nobody repeats it:

```
<a:p><!-- literal <a:br> --><a:r><a:t>KEEP</a:t></a:r><a:br></a:br></a:p>
```

Rewriting text before parsing does not respect XML lexical structure, so the replacement lands inside the comment, consumes its terminator, and the parser throws `Comment is not closed.` The whole attachment then fails, where `main` returned `KEEP` perfectly well. `<![CDATA[<a:fld>7</a:fld>]]>` had its visible text corrupted the same way.

`preserveOrder` avoids the problem instead of patching around it: the parser hands back the paragraph's children as an ordered array, so no raw-XML surgery is needed at all. The collector keeps the `isText` rule from #51, taking only `#text` directly under `a:t`, so the layout whitespace around the other elements still stays out of the result.

**Diffed against main across every paragraph shape**

I built both versions in memory and ran 21 slide shapes through each:

- **0 failures** on this side. The comment case no longer throws. Its output is not identical to
  `main`, because the paragraph printed above ends in `<a:br></a:br>`: `main` gives
  `"## Slide 1\nKEEP"` and this gives `"## Slide 1\nKEEP\n"`, which is the intended break. Drop
  the trailing break and the two agree exactly.
- **CDATA identical to `main`**: `<a:fld>7</a:fld>` stays literal text.
- **The lookalikes are identical to `main`**: `<a:brk>`, `<a:fldChar>`, self-closing `<a:fld/>`, and an escaped `&lt;a:br/&gt;` written as visible text.
- The 8 shapes that differ are the intended ones: breaks now break, fields land in place, and a pretty-printed paragraph gains its break without gaining the whitespace that lays it out.

**One limitation, stated rather than hidden**

This module matches DrawingML by its conventional prefix throughout: `a:p`, `a:t`, `a:r`, `a:fld`
and now `a:br` are all compared as literal keys, so namespace bindings are not consulted. A deck
that rebinds `a:` to another namespace on a break element, `<a:br xmlns:a="urn:vendor"/>`, gets a
line break it should not:

```
main  "## Slide 1\nAB"
here  "## Slide 1\nA\nB"
```

Main only avoids that by ignoring every break, including all the legitimate ones, so its result is
accidental rather than principled. A vendor break under its own prefix (`v:br`) is correctly
ignored by both. Making this right means namespace-aware matching for the whole module, including
paragraph and text extraction, which is a larger change than this one and I would rather not
smuggle it in here. Happy to follow up with it if you want it.

## Related issue

None.

## Validation

- [x] `npm test`: exits 0. `packages/file-parse` 26 passed
- [x] The new assertion fails against `main` and passes here, verified by swapping in
      `origin/main`'s `pptx.ts` and re-running. The other changed assertion passes on both by
      design: it guards the `a:t`-only rule from #51 against regression rather than proving this
      change. Dropping that rule now fails the suite, which it did not before I laid the field
      paragraph out across lines.

The fixture slide now carries two adjacent breaks, a mid-paragraph slide-number field, and a comment naming both tags, so the text-level rewrite that used to be here cannot come back unnoticed.

## Screenshots or recordings

Not applicable. The output is the text handed to the AI panel.

## Contributor checklist

### DIFF
--- a/packages/file-parse/src/pptx.ts
+++ b/packages/file-parse/src/pptx.ts
@@ -2,56 +2,60 @@ import JSZip from 'jszip'
 import { XMLParser } from 'fast-xml-parser'
 
 // Text fidelity: no trim (xml:space="preserve" runs carry the spaces between words),
-// no numeric coercion of tag values (otherwise <a:t>02139</a:t> becomes a number and loses characters)
-const parser = new XMLParser({ ignoreAttributes: true, trimValues: false, parseTagValue: false })
+// no numeric coercion of tag values (otherwise <a:t>02139</a:t> becomes a number and loses characters).
+// preserveOrder keeps <a:br> and <a:fld> in sequence with the <a:r> runs around them; grouped by
+// tag name they lose that position, and a deck's soft breaks and field text land in the wrong place.
+const parser = new XMLParser({
+  ignoreAttributes: true,
+  trimValues: false,
+  parseTagValue: false,
+  preserveOrder: true,
+})
 
 function slideNumber(path: string): number {
   const m = /slide(\d+)\.xml$/.exec(path)
   return m ? Number(m[1]) : 0
 }
 
-/** collect the text of all a:t descendants of one node */
-function collectText(node: unknown, out: string[], isText = false): void {
-  if (node == null) return
-  if (typeof node === 'string' || typeof node === 'number') {
-    // untrimmed, a non-a:t element written across lines carries its whitespace as a value
-    if (isText) out.push(String(node))
-    return
-  }
-  if (Array.isArray(node)) {
-    for (const item of node) collectText(item, out, isText)
-    return
-  }
-  if (typeof node === 'object') {
+/**
+ * One paragraph's text in document order. Only #text directly under a:t counts: untrimmed, the
+ * whitespace laying out any other element is a value too. <a:br> is a soft line break, and
+ * <a:fld> (slide number, date) contributes its own a:t where it sits.
+ */
+function collectText(nodes: readonly unknown[], out: string[], isText = false): void {
+  for (const node of nodes) {
+    if (node == null || typeof node !== 'object') continue
     for (const [key, value] of Object.entries(node)) {
-      if (key === 'a:t') collectText(value, out, true)
-      else if (typeof value === 'object') collectText(value, out)
+      if (key === '#text') {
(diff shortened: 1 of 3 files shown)


Waiel5:
Pushed a fixture change and corrected two statements in the description.

The description said the comment case "matches `main`". It does not, and the XML printed right
above it is why: the paragraph ends in `<a:br></a:br>`, so `main` yields `"## Slide 1\nKEEP"` and
this yields `"## Slide 1\nKEEP\n"`. The true claim is narrower, that it no longer throws.

It also said the changed assertion fails against `main`. Only the new one does; the changed one
passes on both, because it guards the `a:t`-only rule from #51 rather than this change. Worse, on
the old fixture that guard had stopped covering anything: the only layout whitespace sat inside
`<a:br>` , which the walker now short-circuits before the gate can matter, so removing the gate
left the suite green. The field paragraph is now laid out across lines, which puts whitespace
between siblings where only the gate can exclude it. Removing the gate fails the suite again.

merrick-2002:
Merged — thanks! preserveOrder was the right call here, and the fixtures covering comments/double breaks/fields made this easy to verify.

PR #67, opened 2026-08-09
fix(sheets): keep the cell alignment attributes the style model omits

Waiel5:
## Summary

**What changed?**

`resolveStyle` rebuilds a cell's `<alignment>` from a five-field model, so the four `CT_CellAlignment` attributes the model has no field for are dropped the moment any unrelated edit touches that cell's `xf`. One of them is `readingOrder`, the cell's right-to-left flag.

`buildAlignment` reads `horizontal`, `vertical`, `wrapText`, `textRotation` and `indent`, then emits a fresh element from exactly those:

```js
const attributes = [
  ...(horizontal ? [`horizontal="${horizontal}"`] : []),
  ...(vertical ? [`vertical="${vertical}"`] : []),
  ...(wrap ? ['wrapText="1"'] : []),
  ...(rotation ? [`textRotation="${rotation}"`] : []),
  ...(indent ? [`indent="${indent}"`] : []),
]
```

Anything else the cell had is gone. Changing a fill colour on a cell in an Arabic or Hebrew workbook silently rewrites it left-to-right:

```
before  <alignment horizontal="right" readingOrder="2" shrinkToFit="1"/>
after   <alignment horizontal="right"/>
```

`shrinkToFit`, `justifyLastLine` and `relativeIndent` go the same way. This is item 6 of #13, which noted that `readingOrder` in xlsx cell alignment looked unhandled.

**The part worth reviewing: where the values are read from**

Carrying attributes means reading them off the original `xf`, and this file locates elements with regular expressions. An earlier revision of this patch tried to make that extraction stricter than the file's own, adding helpers that stripped `<extLst>` and matched only the `xf`'s start tag. Each round of hardening opened a new hole (a `>` inside an attribute value, nested `<extLst>`, `mc:AlternateContent`), so I removed all of it.

What is left reads the carried attributes off **the same `<alignment>` element the modeled attributes are already read from**, and carries them **only when the `xf` holds nothing but its own `alignment` and `protection`**:

```js
function ownsItsAlignment(xf) {
  for (const element of xf.matchAll(/<\/?([\w.-]+(?::[\w.-]+)?)/g)) {
    if (!/^(xf|alignment|protection)$/.test(element[1] ?? '')) return false
  }
  return true
}
```

That restriction is the part worth reviewing, and it is there because a regex cannot tell whose element it found. Two concrete cases, both of which this rejects:

```
<xf><settings xmlns="urn:vendor" applyAlignment="1"><alignment readingOrder="2"/></settings></xf>
<xf><mc:AlternateContent><mc:Choice Requires="x14">
      <alignment horizontal="general" readingOrder="2"/>
    </mc:Choice><mc:Fallback/></mc:AlternateContent></xf>
```

Without the restriction the first invents an alignment on a cell where `main` emits none, and the second flips the cell right-to-left. The second is worth being precise about: `main` already promotes that branch's `horizontal`, so promoting from there is not new, but `horizontal="general"` is the default and changes nothing on screen while `readingOrder="2"` reverses the text. Same element, materially different consequence, so "no worse than the `horizontal` next to it" would not have been a fair defence.

`CT_Xf` allows only `alignment`, `protection` and `extLst`, so an `xf` holding nothing but the first two cannot hide a second `<alignment>` or an `applyAlignment` belonging to someone else. Anything further and nothing is carried, which leaves that cell exactly as it is today. The existing extraction stays approximate in ways this patch deliberately does not change, for example `readAttribute` matching `\bname="..."` across the whole `xf`; those are separate problems and I would rather not fold them in here.

The one guard the carry does need:

```js
const applies = modeled.length > 0 || isXmlTrue(readCoreAttribute(baseXf, 'applyAlignment'))
```

Building an `<alignment>` out of carried attributes alone would start *applying* an alignment the cell was previously inheriting from its `cellStyleXf`. When no modeled attribute survives and `applyAlignment` is not set, the element is dropped exactly as it is today. `isXmlTrue` accepts both `"1"` and `"true"`, since `applyAlignment` is `xsd:boolean` and both spellings appear in the wild.

**Two commits**

`xlsx-styles.ts` predates the formatter, so touching it reformats around 100 unrelated lines. The first commit is prettier only, the second is the change. Reviewing the second alone should be enough.

## Related issue

Item 6 of #13, the sheets half of RTL support. This is a fidelity fix rather than a feature: nothing new is exposed in the UI, an RTL workbook simply stops losing its reading order on an unrelated edit.

## Validation

- [x] `npm test`: exits 0. `apps/sheets` 1003 passed, 1 skipped, 85 files
- [x] 10 new cases; **4 fail against `main`** and pass here, verified by swapping in `origin/main`'s `xlsx-styles.ts` and re-running. The other 6 are negative guards that should pass either way; the two newest of those fail if the `ownsItsAlignment` restriction is removed, so they are not decorative.

The test helper throws when it cannot find the `xf` it is asked for, so the negative assertions cannot pass against an empty string.

## Screenshots or recordings

Not applicable, the defect is in the saved XML. Reproduction: open an RTL workbook, change any cell's fill colour, save, reopen.

## Contributor checklist

- [x] The change is focused and does not include unrelated reformatting or refactoring. (The reformatting is isolated in its own commit.)

### DIFF
--- a/apps/sheets/src/gateway/xlsx-styles.ts
+++ b/apps/sheets/src/gateway/xlsx-styles.ts
@@ -32,7 +32,9 @@ export class StylesheetEditor {
     const bordersInner = sectionInner(stylesXml, 'borders')
     const cellXfsInner = sectionInner(stylesXml, 'cellXfs')
     if (fontsInner === null || fillsInner === null || cellXfsInner === null) {
-      throw new Error('The workbook stylesheet is missing fonts, fills, or cellXfs — style edits cannot be saved.')
+      throw new Error(
+        'The workbook stylesheet is missing fonts, fills, or cellXfs — style edits cannot be saved.',
+      )
     }
     this.fonts = extractElements(fontsInner, 'font')
     this.fills = extractElements(fillsInner, 'fill')
@@ -44,7 +46,9 @@ export class StylesheetEditor {
     this.hadDxfsSection = dxfsInner !== null
     this.dxfs = dxfsInner === null ? [] : extractElements(dxfsInner, 'dxf')
     if (this.fonts.length === 0 || this.cellXfs.length === 0) {
-      throw new Error('The workbook stylesheet has no base font or cell format — style edits cannot be saved.')
+      throw new Error(
+        'The workbook stylesheet has no base font or cell format — style edits cannot be saved.',
+      )
     }
     this.originalCounts = {
       numFmts: this.numFmts.length,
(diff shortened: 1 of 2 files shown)


Waiel5:
Pushed a correction and updated the description, so the earlier version is not what to review.

Two problems in the first revision, both found by adversarially re-reading my own claims:

1. The `applyAlignment` gate read the whole `xf` string, and `main` never reads that attribute at
   all, so this really did add an extraction surface rather than reusing an existing one. A vendor
   element such as `<settings xmlns="urn:vendor" applyAlignment="1"><alignment readingOrder="2"/></settings>`
   satisfied it, and the cell picked up a right-to-left flag on a cell where `main` emits no
   alignment whatsoever.
2. I had defended the carry as "no worse than the `horizontal` read from the same element". That
   is not a fair defence. For an `<alignment horizontal="general" readingOrder="2"/>` inside an
   `mc:AlternateContent` branch, `main` promotes `horizontal="general"`, which is the default and
   changes nothing on screen, whereas the carried `readingOrder="2"` reverses the cell.

Both are fixed by carrying only when the `xf` contains nothing but its own `alignment` and
`protection`, which `CT_Xf` makes sufficient to prove the element found is the cell's. Both cases
now produce byte-identical output to `main`, and there are two new tests that fail if the
restriction is removed.

merrick-2002:
Merged — thanks! Nice catch on the attributes the style model drops; the ownsItsAlignment/applies guards are exactly the right amount of caution.

PR #68, opened 2026-08-09
fix(docx): declare the prefixes a regenerated part writes, not only the inherited ones

Waiel5:
## Summary

**What changed?**

#63 made regenerated parts reuse the original part's root attributes, so entries spliced back in as their original bytes keep their prefixes bound. That was right for footnotes and endnotes, and it left `comments.xml` half fixed. `buildCommentsXml` does not only splice: it also **generates** `w14:paraId`, and the original root is free not to declare `w14`.

This is a gap in my own change, not a pre-existing one.

`saveDocx` stamps a `paraId` on every comment (`patch.ts:748-758`), and `buildCommentsXml` writes it onto the last paragraph of each comment it rebuilds (`patch.ts:1392-1394`). The root now comes from the original part, and `w14` is a Word 2010 extension namespace rather than part of the base schema, so a document whose `comments.xml` declares only `w` is perfectly legal. This repo's own fixture is exactly that shape (`tests/comments.test.ts:10`):

```xml
<w:comments xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main">
```

Adding a comment to such a document, or editing one in a way that changes its paragraph count so the surgical patch falls back to a rebuild, emits `w14:paraId` with nothing binding it. Before #63 the hardcoded list always bound `w14`, so this is a regression I introduced rather than an old gap.

**The fix**

A regenerated part mixes two sources, so its root has to declare the prefixes of both. `rootAttributes` now keeps everything the original root had and adds only what the engine needs and the original lacked, instead of choosing one list or the other:

```js
const declared = new Set([...attrs.matchAll(/xmlns:[\w.-]+=/g)].map((m) => prefixOf(m[0])))
const missing = [...required.matchAll(/xmlns:[\w.-]+="[^"]*"/g)]
  .map((m) => m[0])
  .filter((declaration) => !declared.has(prefixOf(declaration)))
```

A prefix both sides declare is not duplicated, and a prefix only the original had, `mc` or `r` for instance, still survives, which is what #63 was for.

## Related issue

Follow-up to #63.

## Validation

- [x] `npm test`: exits 0. `packages/docx-engine` 455 passed, 1 skipped, 55 files
- [x] `adds a required prefix the original root did not declare` fails against `main` and passes here, verified by swapping in `origin/main`'s `notes.ts` and re-running
- [x] `still keeps a prefix only the original had` guards the #63 behaviour in the other direction and passes either way by design

As in #63 the test checks prefix binding directly rather than calling `XMLValidator`, which is not namespace-aware and reports all of these as valid.

## Screenshots or recordings

Not applicable, the defect is in the saved XML. Reproduction: open a document whose `comments.xml` declares only `w`, add a comment, save, reopen in Word.

## Contributor checklist

### DIFF
--- a/packages/docx-engine/src/notes.ts
+++ b/packages/docx-engine/src/notes.ts
@@ -35,19 +35,27 @@ const NOTE_NS =
   'xmlns:r="http://schemas.openxmlformats.org/officeDocument/2006/relationships"'
 
 /**
- * Root attributes for a regenerated part. Entries are spliced back in as original bytes, so
- * the root has to keep declaring whatever prefixes those bytes use: Word puts w14:paraId on
- * every paragraph it writes, and a literal namespace list leaves that prefix unbound.
+ * Root attributes for a regenerated part. The part mixes two sources, so the root has to
+ * declare the prefixes of both: entries spliced back in as original bytes may use anything
+ * the original root declared (Word puts w14:paraId on every paragraph it writes), and
+ * entries this engine writes use the prefixes in `required`. Keeping only one side leaves
+ * the other unbound, which Word reports as unreadable content.
  */
 export function rootAttributes(
   originalXml: string | null,
   rootTag: string,
-  fallback: string,
+  required: string,
 ): string {
   const attrs = originalXml
     ? new RegExp(`<${rootTag}\\b([^>]*?)/?>`).exec(originalXml)?.[1]?.trim()
     : undefined
-  return attrs && attrs.includes('xmlns:') ? attrs : fallback
+  if (!attrs || !attrs.includes('xmlns:')) return required
+  const prefixOf = (declaration: string) => /xmlns:([\w.-]+)=/.exec(declaration)?.[1] ?? ''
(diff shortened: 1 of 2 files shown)


merrick-2002:
Thanks for this one too — the diagnosis is exactly right. It turns out we'd fixed the same issue internally shortly before: `rootAttributes` grew a `required` prefixes parameter that comment regeneration passes `w14` through, and it also covers the fallback-attributes branch. That version will land here with the next sync snapshot, so we'll close this PR rather than merge it — nothing wrong with the patch, it just collided with an internal fix. Your recent batch (#66/#67/#69/#70) was excellent; much appreciated, and keep them coming.

Waiel5:
Closing this in favour of the internal fix, which is in `main` now as of the sync.

Yours is the better shape. Taking `required` as a `Record<string, string>` beats my string
scan, and applying it to the fallback branch too covers a case mine did not: a part created
from scratch whose literal namespace list is missing a prefix the writer emits.

No objection at all, and thanks for saying where it collided rather than just closing it.

PR #69, opened 2026-08-09
fix(docx): do not read a nil paragraph border as a real one

Waiel5:
## Summary

**What changed?**

`ST_Border` spells "no border" two ways, `none` and `nil`, and `extractParaFormat` only knew the first:

```js
if (el && attrsOf(el)['w:val'] !== 'none') borders += ch
```

So every `nil` side was recorded as a border that exists. Word writes `nil` whenever a style-level border is reset, which is common in real documents.

This was harmless while nothing in production rebuilt a paragraph's properties from the format model alone. #62 changed that: header and footer paragraphs now go through `mergePPrFormat`, and `formatPPrChildren` re-emits every recorded side as a real rule (`generate.ts:889-894`). A header carrying one blue bottom rule and three `nil` sides comes back with four solid black ones.

Round trip through `parseDocx` then `saveDocx` with a real `word/header1.xml` part:

```
in the document
  <w:top w:val="nil"/><w:left w:val="nil"/><w:right w:val="nil"/>
  <w:bottom w:val="single" w:sz="18" w:color="4472C4"/>

parsed          headerParas[0].borders = "tblr"      <- three phantom sides
saved header    <w:pBdr><w:top .../><w:left .../><w:bottom .../><w:right .../></w:pBdr>
                all four w:val="single" w:sz="4" w:color="auto"
```

`file-actions.ts:197-204` seeds the header context from the parse and passes it straight back on save, so editing one character of header text is enough to stamp a box around it. With the fix the parse yields `"b"` and the saved header keeps its single bottom rule.

Body paragraphs are unaffected either way, because they keep their original `pPr` bytes through the merge branch rather than being rebuilt.

**What this does not fix**

The format model records *which* sides have a border, not how they look, so the surviving bottom rule is still rewritten from `sz="18" color="4472C4"` to `sz="4" color="auto"`. That predates #62 and is a separate change; headers dropped their borders entirely before #62, so it is not a new loss, but I did not want to imply it is solved.

## Related issue

Follow-up to #62, which made a latent parse gap reachable.

## Validation

- [x] `npm test`: exits 0. `packages/docx-engine` 455 passed, 1 skipped, 55 files
- [x] Both new cases fail against `main` and pass here, verified by swapping in `origin/main`'s `parse.ts` and re-running:
  - `does not read a nil side as a border` (the read)
  - `does not stamp rules the header never had when it is saved back` (the write)

## Screenshots or recordings

Not applicable, the defect is in the saved XML, though its effect is visible in Word: four black rules around a header paragraph that had one blue rule under it.

## Contributor checklist

### DIFF
--- a/packages/docx-engine/src/parse.ts
+++ b/packages/docx-engine/src/parse.ts
@@ -1387,7 +1387,10 @@ function extractParaFormat(pPr: XNode): ParaFormat | undefined {
       ['right', 'r'],
     ] as const) {
       const el = findChild(pBdr, `w:${side}`)
-      if (el && attrsOf(el)['w:val'] !== 'none') borders += ch
+      // ST_Border spells "no border" both ways, and Word writes nil whenever a style-level
+      // border is reset, so treating it as present stamps a rule the document never had
+      const val = el ? attrsOf(el)['w:val'] : undefined
+      if (el && val !== 'none' && val !== 'nil') borders += ch
     }
     if (borders) format.borders = borders
   }
--- a/packages/docx-engine/tests/header-footer-rich.test.ts
+++ b/packages/docx-engine/tests/header-footer-rich.test.ts
@@ -1,3 +1,4 @@
+import JSZip from 'jszip'
 import { describe, expect, it } from 'vitest'
 import { parseDocx, saveDocx, TOTAL_PAGES_MARK, type SaveBlock } from '../src/index'
 import { buildDocx } from './helpers/build-docx'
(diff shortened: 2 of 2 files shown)


Waiel5:
The e2e job went red on the previous push and I do not think it is this change, so recording what
the log says rather than leaving the red mark unexplained.

All 22 e2e tests passed. The job failed on teardown:

```
22 passed (3.2m)
Worker teardown timeout of 90000ms exceeded.
1 error was not a part of any test, see above for details
```

The worker logged `<gracefully close start>` and then `<kill> +2m`, so an Electron process outlived
the 90s shutdown window after the run had finished. Every spec in that worker is a sheets one
(`sheets-edit-polish`, `sheets-insert-*`, `sheets-move-rows`, `sheets-new-blank`), while this diff
is five lines in `packages/docx-engine/src/parse.ts` on the docx paragraph-border path plus a docx
test. `apps/sheets` does depend on `@genoffice/docx-engine`, but only for
`@genoffice/docx-engine/math` in the equation dialog; `extractParaFormat` runs when parsing a
.docx and nothing in a sheets flow reaches it. The `test` job on the same commit passed.

I cannot re-run the job from a fork, so I have repushed the same tree to trigger a fresh run. If it
goes red again I will treat it as real and dig in rather than repush.

merrick-2002:
Merged — thanks! Minimal and correct, and appreciated the honest note about the surviving-border sz/color flattening being a separate pre-existing issue.

PR #74, opened 2026-08-09
fix(shell): keep file actions menu in viewport

arpanmehta05:
## Summary

- Positions the Home screen’s file-actions menu relative to the viewport.
- Measures the rendered menu and opens it upward when there is more available space above the trigger.
- Constrains the menu height to the viewport so its actions remain reachable.

## Why is this change needed?

The file-actions menu could extend beyond the bottom of the Home screen, making actions inaccessible. Estimating menu height was unreliable because different file types expose different actions.

## Related issue

Closes #

## Validation

- [ ] `npm run format:check`
- [ ] `npm run lint`
- [ ] `npm run typecheck`
- [ ] `npm test`

Focused validation completed:

- [x] `npx.cmd prettier --check apps/shell/src/renderer/src/Home.tsx apps/shell/src/renderer/src/home.css`
- [x] `npm.cmd run typecheck -w @genoffice/shell`
- [x] `git diff --check`

Checks not run:

- Full repository format, lint, typecheck, and test suites were not run locally because they cover the entire monorepo and are handled by CI.
- Manual Electron visual verification was not run for this separate PR.

## Screenshots or recordings

Not available. The change affects the menu position near the bottom of the Home screen.

## Contributor checklist

### DIFF
--- a/apps/shell/src/renderer/src/Home.tsx
+++ b/apps/shell/src/renderer/src/Home.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useRef, useState } from 'react'
+import { useCallback, useEffect, useLayoutEffect, useRef, useState } from 'react'
 import type { ReactElement } from 'react'
 import logoLockup from './assets/genoffice-logo.svg'
 import iconDocx from './assets/file-docx.svg'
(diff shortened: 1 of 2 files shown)


merrick-2002 (changes requested):
Thanks for this, and apologies for the slow response. You're right that the row menu can get clipped near the bottom of the window, and measuring the rendered menu to decide which way to open is a nice approach.

The Home screen has evolved a bit since this branch was cut: the row menu now has a submenu with its own edge handling, and click-away is managed by a shared `useDismissablePopover` hook. Would you be up for rebasing onto current main and applying the positioning logic on top of that structure? While you're there, it would be worth bounding the menu's max-height by the space on the side it opens toward, so very short windows stay covered too.

Happy to take another look once it's updated. Thanks again for the contribution.

merrick-2002:
Thanks for the fix — keeping the menu in the viewport is still worth doing. The branch has drifted from main since August; if you'd like to bring it up to date we'd be glad to take another look.

merrick-2002:
Closing this one since the branch has drifted a long way from main. Keeping the file-actions menu in the viewport is still worth doing — if you'd like to pick it up again, feel free to rebase and reopen, or open a fresh PR.

PR #78, opened 2026-08-10
feat(ai): Add custom provider support (LM Studio, OpenCode Zen, OpenCode Go)

alfirus:
Add support for custom/local AI providers. New providers: LM Studio (local inference), OpenCode Zen, OpenCode Go. All use OpenAI-compatible protocol. Removes genspark force-override in docs/slides/sheets.


merrick-2002:
Thanks @alfirus, and apologies for the long silence on this one.

We're closing it as superseded. OpenCode Zen and OpenCode Go are landing as built-in providers in the next release, following the request in #182; you proposed them here first, and we appreciate that. For LM Studio, the Custom provider slot (any OpenAI-compatible base URL and key) works today; dedicated support for local model servers is something we may revisit later.

Independently of that, the PR couldn't be taken in this form because the branch carries the fork's whole history rather than just the provider change (a repo-wide Prettier pass, the Google Drive integration commits, .doc support). For future changes, a branch off current main with only the relevant commits makes review much easier.

Thanks again for the continued contributions.

PR #81, opened 2026-08-10
fix(docx): keep table cell paragraph properties through a rebuild

Waiel5:
`tableCellXml` hand-builds each cell paragraph's `w:pPr` from `w:numPr` and `w:jc` alone, so every other property the parser captured is discarded when a structural table edit regenerates the table. `w:bidi` is one of them.

## What it looks like

Word writes an RTL cell as `w:bidi` plus a **logical** `w:jc`, where `left` means start and renders flush right. `extractParaFormat` converts that to a visual value, and the hand-built emitter wrote the visual value back as though it were the logical one.

Same input table through `parseDocx` + `generateTableModelXml`, `origin/main` sources against this branch:

| | `origin/main` | this branch |
|---|---|---|
| `<w:bidi/>` | 0 | 2 |
| `w:jc` | `right` | `left` |
| `w:spacing` / `w:ind` / `w:shd` | dropped | kept |

Two distinct failures. The cell that declared `<w:bidi/><w:jc w:val="left"/>` comes back LTR. The cell that declared `<w:bidi/>` and no `w:jc` comes back with no `w:pPr` at all, so it moves from the RTL default (flush right) to the LTR default (flush left). The table keeps `w:bidiVisual` from its original `tblPr` throughout, so the result is right-to-left columns holding left-to-right paragraphs, which is a shape Word never writes.

## The fix

Route the cell through `mergePPrFormat`, the helper #62 used for the same defect in headers and footers. Shading, spacing, indent, page-break-before, paragraph borders, tab stops and the drop-cap frame come back with it.

### A prerequisite inside `mergePPrFormat`

Its interleave branch walks the fresh children once with a monotonic index, so they have to arrive in schema order, but only the fresh-build branch sorted them. `formatPPrChildren` emits by concern and leaves `framePr` and `tabs` until last, so a paragraph carrying a tab stop came back as `w:jc` followed by `w:tabs`, which is the ordering Word offers to repair. Every table paragraph takes the interleave branch, so this had to be correct before the rest could be. The fresh children are now sorted on both branches, and `tests/schema-order.test.ts` gains coverage for the branch that had none.

### The alignment trap

`cell.align` is read straight off `w:jc` at `parse.ts:2506`, so it holds Word's logical value, while `paragraph.align` has already been converted to a visual one. Only `left` and `right` differ between the two, so a bidi paragraph takes the cell fallback for `center` and `justify` and skips it for the sided values, where the cell value cannot be placed in either coordinate system with confidence. Outside bidi the fallback is unchanged.

## Verification

Six of the seven added cases fail against `origin/main` sources and pass here. The seventh, `still falls back to the cell alignment for a paragraph that declared none`, passes on both: it is a control for the path that should not change, not regression coverage.

`packages/docx-engine` 532 passed, `apps/docs` 733 passed.

## What this does not fix

- `keepNext`, `keepLines`, `widowControl` and `contextualSpacing` are parsed into the format model, but `formatPPrChildren` emits none of them, so they are still lost on every rebuild, in tables and elsewhere alike.
- `w:tblPrEx` is neither parsed nor emitted, so a row-level property exception is still dropped.
- A cell paragraph carries neither `rawPPr` nor `pPrChange`, so a paragraph-level tracked revision inside a cell does not survive regeneration. That is equally true before this change.

### DIFF
--- a/packages/docx-engine/src/generate.ts
+++ b/packages/docx-engine/src/generate.ts
@@ -1221,9 +1221,13 @@ export function mergePPrFormat(rawPPr: string, format: ParaFormat | undefined): 
   const rebuilt = new Set(
     [...managedTags].filter((tag) => !pprGroupUnchanged(tag, rawOf(tag), format ?? {})),
   )
-  const freshOut = fresh.filter((c) => rebuilt.has(c.name))
-  const kept = rawChildren.filter((c) => !rebuilt.has(c.name))
   const rank = (n: string) => PPR_CHILD_ORDER.indexOf(n)
+  // the interleave below walks freshOut once with a monotonic index, so it has to arrive in
+  // schema order; formatPPrChildren emits by concern and leaves framePr and tabs until last
+  const freshOut = fresh
+    .filter((c) => rebuilt.has(c.name))
+    .sort((a, b) => rank(a.name) - rank(b.name))
+  const kept = rawChildren.filter((c) => !rebuilt.has(c.name))
   const parts: string[] = []
   let fi = 0
   let prevRank = -1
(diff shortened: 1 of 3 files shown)


merrick-2002:
Merged — thank you! We confirmed both defects locally: the hand-built cell pPr really did drop `w:bidi`/spacing/indent/shading on every table rebuild, and `cell.align` does hold Word's logical value while the paragraph align is visual. Routing cells through `mergePPrFormat` with the interleave-branch ordering fix is the right shape. Great analysis of the alignment trap.

PR #83, opened 2026-08-11
fix(docx): carry bold and italic to complex-script text

Waiel5:
Word renders Arabic and Hebrew **bold** from `w:bCs` and **italic** from `w:iCs`, not from `w:b` and `w:i`. `modelRPrChildren` writes only the Latin half, so selecting Arabic text and clicking Bold changes nothing on screen.

## What happens

`w:b` and `w:bCs` are one managed group, so while the value is unchanged the original bytes survive untouched. The moment the value changes the group is rebuilt from the model, and only `w:b` comes back.

Executed against `origin/main` sources, a plain Arabic run with Bold turned on:

```
in   <w:rPr><w:rFonts w:cs="Traditional Arabic"/><w:szCs w:val="28"/></w:rPr>
out  <w:rPr><w:rFonts w:cs="Traditional Arabic"/><w:b/><w:szCs w:val="28"/></w:rPr>
```

Un-bolding and re-bolding is worse: the `w:bCs` the document arrived with is gone for good.

## The fix

Emit the Cs twin alongside its Latin half. That is what Word writes for either script, and what this emitter already does one property over. `RUN_MANAGED_GROUPS` pairs `w:sz` with `w:szCs`, and `modelRPrChildren` has always written both:

```
mergeRPrModel('', { bold: true })          -> <w:rPr><w:b/></w:rPr>
mergeRPrModel('', { sizeHalfPoints: 28 })  -> <w:rPr><w:sz w:val="28"/><w:szCs w:val="28"/></w:rPr>
```

A run whose flag has not changed still keeps its original bytes exactly.

## Verification

Four of the nine cases in `bidi.test.ts` fail against `origin/main` sources and pass here. The other five are controls for behaviour that must not change, including a run whose bold was never toggled coming back byte for byte.

`packages/docx-engine` 530 passed, `apps/docs` 737 passed. Four assertions in `textbox-edit.test.ts` pinned the old byte output and are updated; the neighbouring run they leave alone is the one whose bold was never toggled.

## What this does not fix

The model holds one flag per property and cannot tell `w:b` from `w:b` plus `w:bCs`, so any path that rebuilds a run from the model writes the Cs twin whether or not the original carried it. Rejecting a tracked format revision is such a path. That is already true of `w:szCs` today.

The model does not see the Cs flags at all, either: the parser reads `w:b` and `w:i` and nothing complex-script, so a run carrying `w:bCs` without `w:b` arrives with no bold and any rebuild drops it. Same gap from the other side.

I tried to repair the reject path by reading the previous properties back out of the `w:rPr` that `w:rPrChange` nests, and took it out again. The snapshot the editor keeps is a string with no namespace or entity context, and three shapes of the fix each broke on that:

- taking the nested `w:rPr` whole carries the namespace declarations it inherited, so a property under a prefix bound above it returns unbound and the part no longer parses;
- taking only the six toggle and size elements avoids that, but names are compared lexically, so a `w:rPrChange` that rebinds the `w` prefix launders a foreign `w:bCs` in, and an ignorable wrapper under a Unicode prefix hides history that is then restored as the run's own;
- a `w:val` written as a numeric character reference survives the round trip already escaped, so passing it through again yields a value that no longer parses as a number.

Doing it properly means carrying real namespace and entity context through the revision snapshot. That is a change to the editor model rather than to this emitter, and it belongs to `w:sz` exactly as much as to `w:b`, so I have left it alone rather than half done. Happy to take it on separately if you would like it.

### DIFF
--- a/packages/docx-engine/src/generate.ts
+++ b/packages/docx-engine/src/generate.ts
@@ -2014,6 +2014,10 @@ function revisionRPrChangeXml(run: Run): string | null {
   const props: string[] = []
   if (old.styleId) props.push(`<w:rStyle w:val="${escapeXmlAttr(old.styleId)}"/>`)
   if (old.font || old.fontAscii) props.push(freshRFontsXml(old.font, old.fontAscii))
+  // no Cs twins here: this nested w:rPr records what the run looked like before the tracked
+  // revision, and old.bold cannot tell w:b from w:b plus w:bCs. Adding them would invent a
+  // complex-script flag the document never had, so rejecting the revision would bold Arabic
+  // that was never bold.
   if (old.bold) props.push('<w:b/>')
   if (old.italic) props.push('<w:i/>')
   if (old.strike) props.push('<w:strike/>')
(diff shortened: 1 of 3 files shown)


merrick-2002:
Merged — thank you! We verified the root cause against the engine sources: `w:b`/`w:bCs` are managed as one group but only the Latin half was ever re-emitted, exactly as you described. The write-up on what this deliberately does not fix (the rPrChange snapshot path) was much appreciated and made the review easy.

PR #84, opened 2026-08-11
fix(slides): apply a font change to complex-script text as well

Waiel5:
OOXML splits a run's font into three slots and picks by script: `a:latin`, `a:ea` for CJK, and `a:cs` for complex scripts. Changing an element's font never wrote `a:cs`, so Arabic, Hebrew and Indic text stayed on the old typeface and the control looked broken for anyone writing in those scripts.

## What happens

Executed against `origin/main` sources, picking Amiri on a mixed run:

```
before  <a:latin typeface="Calibri"/><a:cs typeface="Traditional Arabic"/>
after   <a:latin typeface="Amiri"/><a:cs typeface="Traditional Arabic"/>
```

The Latin half of the run changes and the Arabic half does not. A run that declares no fonts at all, which is the shape the deck fixtures actually use, got `a:latin` and `a:ea` injected and still no `a:cs`.

## The fix

Write the three slots together wherever the font is set, and clear `csFont` with its `latinFont` and `eaFont` siblings so a rebuilt run re-derives it from the choice the user made instead of re-emitting the old typeface.

`patchRunFont` works on the direct children of the run's own `a:rPr`, and the group is rebuilt whole, so the slots always come back as `latin, ea, cs` however few of them the run arrived with, and each element that was already there keeps its other attributes (`pitchFamily`, `charset`, `panose`).

### Where it deliberately does nothing new

The structured path is taken only when the run can actually be read: no comment, CDATA or processing instruction; every tag name inside it readable exactly as the existing child scanner reads one; the run's own `a:rPr` locatable as a direct child; every direct child of it a known `CT_TextCharacterProperties` element; and no repeated font slot.

Anything else falls back to the previous function unchanged, with `a:cs` untouched. So a run carrying a markup-compatibility branch, an extension payload with its own `a:rPr`, or a name the scanner would read short is handled exactly as it was before rather than guessed at. The shared child scanner itself is untouched, so no other caller moves.

## Verification

Twelve of the thirty-one cases in `run-format-fidelity.test.ts` fail against `origin/main` sources and pass here. The other nineteen are existing baselines plus guards for the preservation and fallback behaviour, so they pass on both revisions by design.

A differential over a large corpus of valid and well-formed-but-invalid runs, comparing this branch against `origin/main` side by side, found no divergence outside the font slots.

`packages/pptx-engine` 557 passed.

## What this does not fix

- An empty or self-closing `a:rPr` still comes back with `a:latin` and `a:ea` only, so the complex-script slot is not created for a run that declared no properties at all. That is the previous behaviour, unchanged.
- Structural editing, the format painter and the AI font path still leave a stale `csFont` on the model. Also unchanged, and a separate thing to fix.
- An empty `a:rPr` whose opening tag carries a `>` inside a quoted attribute value is still mangled by the pre-existing injection path. Equally broken before and after.

### DIFF
--- a/packages/pptx-engine/src/generate.ts
+++ b/packages/pptx-engine/src/generate.ts
@@ -330,16 +330,17 @@ function patchRunProps(runXml: string, run: TextRun): string {
     }
     runXml = patchRunHlink(runXml, run)
   } else {
-    // No rPr: inject a minimal rPr after <a:r> (no latin/ea injected when the font is untouched — inheritance applies)
+    // No rPr: inject a minimal rPr after <a:r> (no font slots injected when the font is untouched,
+    // so inheritance applies)
     const attrs = buildRPrAttrs(run)
     const color = run.color
       ? `<a:solidFill><a:srgbClr val="${hex6(run.color)}"/></a:solidFill>`
       : ''
     const font =
       run.fontFamily && !run.fontImplicit && !run.latinFont && !run.eaFont
-        ? `<a:latin typeface="${escapeXmlAttr(run.fontFamily)}"/><a:ea typeface="${escapeXmlAttr(run.fontFamily)}"/>`
+        ? fontSlotsXml(escapeXmlAttr(run.fontFamily))
         : ''
-    const inner = color + font + hlinkXml(run) // schema order: fill before latin/ea, hlinkClick after
+    const inner = color + font + hlinkXml(run) // schema order: fill before the font slots, hlinkClick after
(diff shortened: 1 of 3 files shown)


merrick-2002:
Merged — thank you! Complex-script text was indeed left on the theme cs font by every font change. The structural scan with a byte-identical fallback for unscannable runs is a careful design, and the test coverage across the edge cases (MC branches, extension payloads, repeated slots) gave us a lot of confidence merging this.

Issue #65, opened 2026-08-09
Table formatting tools


genuinechimere:
### App

GenOffice Docs (docs)

### Problem

There is not enough formatting table formatting tools on Genoffice docs. The tables cannot be moved around just like in MS Word document. One cannot format tables just the way we do for research reporting. Also, check text box creation. The formatting tools for Genoffice Docs needs serious upgrade. A lot of formatting tools found in MS Word Document are either missing or not functioning well in Genoffice Docs. Because of table formatting I had to switch back to MS Word document because Genoffice Docs was becoming problematic. 

### Proposed solution

I want the formatting tools to work just the way it does on MS Word document. Take a look at MS Word document and put the all formatting tools in Genoffice Docs for a good user experience.

merrick-2002:
Thanks for the detailed feedback — we hear you, and we know table formatting is a real gap for research-style documents.

Some of this is already in Docs, and we’ll keep building from there:

**Tables**
- Insert a table from the Insert tab
- Table Design: table styles, cell shading, and borders (all / outer / inner / none, plus color and width)
- Table Layout: insert or delete rows and columns, merge or split cells, row height and column width, cell vertical alignment, and table left / center / right alignment
- Drag the handle at the top-left of a table to select it and move it in the document
- Insert a table caption from the References tab

**Text boxes**
- Insert a text box from the Insert tab
- Wrap text around a text box from the Layout tab

What’s still missing — especially Word-style table wrapping and free positioning on the page, plus more research-report layout (repeat header row, autofit, cell margins, and richer table styles) — we plan to add gradually rather than waiting for a single “full Word parity” drop.

merrick-2002:
Following up on this: [GenOffice v0.8.262](https://github.com/genspark-ai/genoffice/releases/tag/v0.8.262) ships a major round of table upgrades in Docs that targets the gaps you called out:

- **Tables can now be moved around like in Word**: floating table placement with text wrapping, so a table can be dragged and positioned freely on the page.
- **Table Design**: a full table-style gallery with presets, plus editable table-style options (header row, banded rows, first/last column, and so on).
- **Table Layout**: repeated header rows across pages, AutoFit modes (fixed column width / AutoFit contents / AutoFit window), default cell margins, and a complete Table Properties dialog.
- **Text boxes** can be inserted from the Insert tab and typed into directly.

[...]

PR #89, opened 2026-08-12
fix(docx): decode XML entities exactly once in parsed run text

### LINKED ISSUE #88: docx-engine: run text is entity-decoded twice — literal "&lt;"/"&amp;" text corrupts on parse and is permanently rewritten on save
**App**: GenOffice Docs — `packages/docx-engine` (parse/save path)
**Version and OS**: `main` @ dc4d7e5, macOS 15 (parser-level, OS-independent)

### What happened?

`buildRun()` in `packages/docx-engine/src/parse.ts:2370` applies `decodeEntities()` on top of text that fast-xml-parser has already entity-decoded once. Any document whose visible text literally contains `&lt;`, `&gt;`, `&amp;`, `&quot;`, `&apos;` — or numeric-reference-looking text like `&#65;` — is corrupted at parse time. Documents about HTML/XML escaping, pasted HTML/XML source, or URLs written with `&amp;` are typical carriers. Worse, once the paragraph is edited and saved, the corrupted text is written back to `word/document.xml`, silently and permanently.

### Repro

A `document.xml` run stored per XML 1.0 (Word displays it as `Escape &amp; as &lt;b&gt; ok`):

```xml
<w:t xml:space="preserve">Escape &amp;amp; as &amp;lt;b&amp;gt; ok</w:t>
```

```ts
const doc = await parseDocx(bytes)
doc.blocks[0].runs[0].text
// actual:   "Escape & as <b> ok"           ← double-decoded
// expected: "Escape &amp; as &lt;b&gt; ok" ← what Word shows
```

Editing the *other* run of the same paragraph and saving rewrites the untouched run:

```
before save: <w:t xml:space="preserve">Escape &amp;amp; as &amp;lt;b&amp;gt; ok </w:t>
after save:  <w:t xml:space="preserve">Escape &amp; as &lt;b&gt; ok </w:t>
```

The original text is permanently lost, and the user never saw it rendered correctly in the first place (the AI panel reads the same corrupted text). Numeric-looking text is hit too: visible text `char &#65; here` (stored `char &amp;#65; here`) parses as `char A here`.

### Expected

README line 28: “**Microsoft Word–compatible, byte-preserving `.docx` editing** — only what you touched changes; Word never notices.” README lines 170–171: “Saving is byte-preserving: parts of the file you didn't touch are written back byte-for-byte, so documents keep working in Microsoft Office.” A literal `&lt;` in document text (stored as `&amp;lt;`) must survive parse → edit → save; XML entities decode exactly once.

### Root cause

The shared `xmlParser` (`src/xml-utils.ts`) keeps fast-xml-parser's `processEntities: true` default, so `w:t` text and attribute values read off the parse tree already have the five named entities resolved exactly once. fast-xml-parser does **not** resolve numeric character references (Word writes symbol-font code points like `&#xF0B7;`), which is presumably why `decodeEntities()` was layered on top — but that helper re-decodes the named entities a second time.

Call sites where `decodeEntities()` runs on parser output (double decode): `parse.ts:2370` (`buildRun`, every body/table run), `:2288` (ruby), `:1861` (textbox), `:3883`/`:4057` (SmartArt/diagram), `:4589` (numbering `lvlText`), and `:2059` (HYPERLINK field instruction, where a URL whose text literally contains `&amp;` loses it). Call sites decoding raw XML slices (`plainText`, TOC lines, bookmark names at `:1181`, the image alt at `:954`, `text-patch.ts`, `sources.ts`) are correct as-is — there the input never went through the parser.

### Suggested fix

Split the helper: a `decodeNumericCharRefs()` (the numeric-reference replace only) for text/attributes read off the parse tree, keeping the full `decodeEntities()` for raw XML slices. With that change the whole docx-engine suite still passes (550 existing tests) and the corruption above is gone. Known residual edge, much rarer: visible text `&#65;` (stored `&amp;#65;`) is indistinguishable from a real numeric reference after the parser's first pass; fully separating those would require single-pass decoding of the raw slice.

I have a minimal fix with regression tests ready and will open a PR referencing this issue.

---
This report was produced with AI assistance; I reproduced it locally and reviewed every conclusion myself.

codeAnqiang-ma:
## Summary

- **What changed?** `decodeEntities()` was being applied to text that fast-xml-parser (with its `processEntities: true` default) had already entity-decoded once, so the five named entities were decoded twice. A document whose visible text literally contains `&lt;` / `&amp;` (docs about HTML/XML escaping, pasted markup, URLs written with `&amp;`) parsed as `<` / `&`, and editing that paragraph wrote the corrupted text back to `word/document.xml` permanently. This PR splits `decodeNumericCharRefs()` out of `decodeEntities()` — numeric character references like `&#xF0B7;` are the only thing the parser leaves unresolved — and uses it at the call sites that read text or attribute values off the parse tree: `buildRun` (every body/table run), ruby text, textbox runs, SmartArt/diagram text, numbering `lvlText`, and the HYPERLINK field instruction. Call sites that decode raw XML slices (`plainText`, TOC lines, bookmark names, image alt, `text-patch.ts`, `sources.ts`) still need and keep the full decode, unchanged.
- **Why is this change needed?** The double decode breaks the core byte-preservation promise (README: “only what you touched changes; Word never notices”): affected documents are misparsed on open and silently rewritten on save. Details and repro in the linked issue.
- Intentionally out of scope (see the issue's “known residual edge”): visible text `&#65;` stored as `&amp;#65;` is indistinguishable from a real numeric reference after the parser's first pass; fixing that would require single-pass decoding of the raw slice.

## Related issue

Fixes #88

## Validation

- [x] `npm run format:check` — “All matched files use Prettier code style!”
- [x] `npm run lint` — ESLint (repo flat config) run on the two changed files: 0 errors
- [x] `npm run typecheck` — `tsc --noEmit` for `@genoffice/docx-engine`: clean
- [x] `npm test` — full docx-engine suite (`npx vitest run` in `packages/docx-engine`): **60 files, 553 passed, 1 skipped, 0 failed**

List any checks not run and explain why: the repo-wide variants of lint/typecheck/test were not run — a full workspace `npm install` fails in my environment because my registry mirror 404s on `@fluentui/react-icons@2.0.333` (an app dependency unrelated to this change). The affected package passes all four checks as listed above.

**Regression evidence** — the three new tests in `tests/entity-decoding.test.ts` fail on `main` and pass with this fix:

```
main (fix stashed):  Tests  3 failed (3)
  × keeps text that is literally an entity reference
      expected 'Escape &amp; as &lt;b&gt; ok', got 'Escape & as <b> ok'
  × still resolves a real numeric character reference beside literal entity text
  × does not rewrite an untouched run when a sibling run is edited
      saved document.xml contains '<w:t xml:space="preserve">Escape &amp; as &lt;b&gt; ok </w:t>'
      instead of the original 'Escape &amp;amp; as &amp;lt;b&amp;gt; ok '

with this PR:        Tests  553 passed | 1 skipped (554)   ← includes the 550 pre-existing tests
```

## Screenshots or recordings

Not applicable — engine-level fix; test evidence above.

## Contributor checklist

- [x] User-facing strings use the existing i18n resources. *(No user-facing strings touched.)*
- [x] File open/save changes include an appropriate round-trip or fidelity test. *(`does not rewrite an untouched run when a sibling run is edited` asserts the untouched run's original bytes survive an edit–save cycle.)*

---
AI disclosure: this fix and its tests were prepared with AI assistance; I reproduced the bug and reviewed every change and test result locally.

### DIFF
--- a/packages/docx-engine/src/parse.ts
+++ b/packages/docx-engine/src/parse.ts
@@ -1858,7 +1858,7 @@ function extractTextboxes(
         for (const r of childrenOf(p)) {
           if (nameOf(r) !== 'a:r') continue
           const t = findChild(r, 'a:t')
-          const text = t ? decodeEntities(textOf(t)) : ''
+          const text = t ? decodeNumericCharRefs(textOf(t)) : ''
           if (text === '') continue
           const run: Run = { text }
           const rPr = findChild(r, 'a:rPr')
@@ -2056,7 +2056,9 @@ const SIMPLE_INLINE_FIELD_RE = /^\s*(DATE|TIME|CREATEDATE|SAVEDATE|NUMPAGES|FILE
 /** HYPERLINK "url" (optional \o "tip"): the only field form folded into an editable link run;
  * any other switch (\l bookmark, \t frame...) keeps the protected-passthrough path */
 function convertibleHyperlink(instr: string): { href: string; tooltip?: string } | null {
-  const m = /^\s*HYPERLINK\s+"([^"\\]+)"\s*(?:\\o\s+"([^"]*)"\s*)?$/.exec(decodeEntities(instr))
+  const m = /^\s*HYPERLINK\s+"([^"\\]+)"\s*(?:\\o\s+"([^"]*)"\s*)?$/.exec(
+    decodeNumericCharRefs(instr),
+  )
   if (!m) return null
   return { href: m[1], ...(m[2] ? { tooltip: m[2] } : {}) }
 }
@@ -2285,7 +2287,7 @@ function rubyPartText(rubyNode: XNode, part: 'w:rubyBase' | 'w:rt'): string {
   for (const r of childrenOf(partNode)) {
     if (nameOf(r) !== 'w:r') continue
     for (const c of childrenOf(r)) {
-      if (nameOf(c) === 'w:t') text += decodeEntities(textOf(c))
+      if (nameOf(c) === 'w:t') text += decodeNumericCharRefs(textOf(c))
     }
   }
   return text
@@ -2367,7 +2369,7 @@ function buildRun(
   let text = ''
   for (const child of childrenOf(rNode)) {
     const name = nameOf(child)
-    if (name === 'w:t' || name === 'w:delText') text += decodeEntities(textOf(child))
+    if (name === 'w:t' || name === 'w:delText') text += decodeNumericCharRefs(textOf(child))
     else if (name === 'w:tab') text += '\t'
     // In-paragraph page breaks (w:br w:type="page") are encoded as \f and preserved; column/soft breaks become \n
     else if (name === 'w:br') text += attrsOf(child)['w:type'] === 'page' ? '\f' : '\n'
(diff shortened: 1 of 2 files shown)


merrick-2002:
Thank you @codeAnqiang-ma for the excellent report and fix! The diagnosis is spot on — we reproduced the double decode with the same parser configuration, and splitting out the numeric character reference decoding for parse-tree call sites is exactly the right approach. The regression tests covering the edit round-trip are much appreciated. Merged.

PR #94, opened 2026-08-14
Add VSCode-style collapsing sidebar

w15147m:
## Summary

- What changed?
- Why is this change needed?

## Related issue

Closes #

## Validation

- [ ] `npm run format:check`
- [ ] `npm run lint`
- [ ] `npm run typecheck`
- [ ] `npm test`

## Screenshots or recordings

Include before/after evidence for visible changes, or write "Not applicable."

## Contributor checklist

- [ ] The change is focused and does not include unrelated reformatting or refactoring.
- [ ] User-facing strings use the existing i18n resources.
- [ ] File open/save changes include an appropriate round-trip or fidelity test.

### DIFF
--- a/apps/shell/src/main/index.ts
+++ b/apps/shell/src/main/index.ts
@@ -1521,7 +1521,7 @@ function createShellWindow(): void {
     // thumbnail pane) through to the desktop
     ...(process.platform === 'darwin'
       ? { titleBarStyle: 'hiddenInset' as const, vibrancy: 'sidebar' as const }
-      : { frame: false, autoHideMenuBar: true }),
+      : { frame: false, titleBarStyle: 'hidden' as const, autoHideMenuBar: true }),
     webPreferences: {
       preload: join(__dirname, '../preload/index.js'),
       contextIsolation: true,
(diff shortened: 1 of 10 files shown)


merrick-2002:
Thanks for the contribution and for exploring the shell layout.

We're going to close this rather than review it further, for two reasons:

1. The change is much broader than a collapsing sidebar. It switches non-macOS windows to a frameless custom title bar with hand-drawn window controls and menu buttons, doubles the tab strip height, and keeps a persistent sidebar next to open documents. Those are product decisions about window chrome that we'd want to discuss in an issue with mockups before any code, and the PR has no description or screenshots.
2. Several details wouldn't pass review as-is: the `existsSync` guard before `showItemInFolder` is removed, the new CSS uses raw `rgba()` values instead of the theme tokens (CI enforces this), and the styling is inline with `as any` casts. The branch also conflicts with main in `tab-manager.ts` and `AppFrame.tsx`.

If you'd like to propose a collapsible sidebar on its own, please open an issue describing the behavior first. Thanks again.

PR #96, opened 2026-08-16
Replace unrecognizable SaveIcon with floppy disk

ulangmeier:
The SaveIcon was not recognizable as a "Save" function, because the used Icon was not a typical floppy disk. Improved UI/UX with the a new standard floppy disk path.

## Summary

- This pull request changes the Save Icon to a more standard and recognizeable one.
- This change is needed so that users can understand where to save the document.

## Screenshots or recordings

"Not applicable"

## Contributor checklist

- [X] The change is focused and does not include unrelated reformatting or refactoring.
- [X] User-facing strings use the existing i18n resources.
- [X] File open/save changes include an appropriate round-trip or fidelity test.

### DIFF
--- a/apps/sheets/src/renderer/ribbon-icons.tsx
+++ b/apps/sheets/src/renderer/ribbon-icons.tsx
@@ -42,10 +42,9 @@ export function CaretIcon(): ReactElement {
 export function SaveIcon(): ReactElement {
   return (
     <Icon>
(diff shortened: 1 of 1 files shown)


merrick-2002:
Thanks @ulangmeier for the contribution, and welcome as a contributor! 🎉

You're right that the standard floppy glyph reads better as "Save". One note: the same save-icon geometry is intentionally shared across the four apps (Docs, Sheets, Slides, PDF), so we'll follow up shortly to apply the same Lucide-style floppy to the other three apps to keep them consistent. That follow-up will land in one of the next sync commits.

PR #129, opened 2026-08-23
feat(ai): stream DeepSeek reasoning through the agent loop (port to BYOK provider architecture)

konghaojie-k2:
Port the locally-built DeepSeek reasoning feature onto the new multi-provider (BYOK) AI architecture.

## What
- **ai-provider**: add `onReasoning` to `StreamCallbacks`; capture `reasoning_content` deltas in the OpenAI-compatible protocol and echo them back on tool-call turns (DeepSeek rejects follow-ups that drop them with a 400)
- **ai-provider**: DeepSeek empty-turn retry — thinking-only / zero-token `stop+empty` turns are retried once before failing cleanly (only for the `deepseek` provider)
- **agent-core**: keep `reasoning` on assistant history alongside existing behavior; wire the `reasoning` IPC chunk through the electron transport
- **main handlers (docs/sheets/slides)**: forward `onReasoning` chunks
- **slides**: add a `set_speaker_notes` agent tool (persists notes into the pptx) — coexists with `apply_ops`
- small dev hygiene: ignore local release output dirs + daily-reflection state

## Verification
- typecheck passes: ai-provider, agent-core, ui, and all 5 apps
- ai-provider tests 115/115 (incl. new reasoning + empty-turn-retry cases)
- agent-core tests 62/62, slides speaker-notes + layout tests 22/22

This intentionally does **not** include the removed local `AiSettingsDialog` (superseded by the refreshed AI settings UI) or the cross-env dev tweak (superseded by `predev build-stale-preloads`).


merrick-2002:
Hi @konghaojie-k2, thanks for this contribution — and apologies for the delayed review.

The core issue you identified is real and well-diagnosed: DeepSeek's thinking mode requires `reasoning_content` to be echoed back on tool-call turns, and dropping it breaks multi-turn tool use. We independently ran into the same problem, and a fix has just landed on our internal mainline — it will reach this repo in an upcoming sync. It covers the same ground with a few extra pieces we needed (per-model gating so the field is never sent to providers that don't accept it, OpenRouter's `reasoning` delta variant, the non-SSE fallback path, and stripping reasoning from completed runs to keep long conversations from bloating requests). Since that supersedes the provider/agent-core changes here, we're closing this PR rather than asking you to rebase through the conflicts.

That said, two pieces of this PR are genuinely nice standalone improvements that we'd be happy to take as separate PRs against a fresh `main`:

- the `set_speaker_notes` agent tool (the implementation and tests look solid)
- the horizontal-overflow check in `layout-audit.ts`

If you'd like to resubmit those, please keep each PR single-purpose and drop the local-environment bits (the `.gitignore` entries and the unused markdown `setAiSettings` channel). Thanks again for digging into the provider layer — the empty-turn retry observation is a useful data point too, and we'll verify it against the live API as a follow-up.

PR #131, opened 2026-08-23
fix(sheets): extend Ctrl+F beyond the loaded window on streamed workbooks

### LINKED ISSUE #113: Find/Replace (Ctrl+F) only searches loaded cells, missing matches in lazy-loaded sheets
### App

GenOffice Sheets (sheets)

### Version and OS

Windows11

### What happened?

Steps to reproduce:
1. Open a large spreadsheet (one big enough to be lazy-loaded / streamed into the grid on demand).
2. Without scrolling through the sheet, press Ctrl+F and search for a value that exists in a row far below the initially loaded area.
3. The search reports no match (or only finds matches within the rows that have been scrolled into view).

Expected behavior:
Find should search the entire sheet regardless of which rows are currently loaded — consistent with Excel, where Find always covers the whole workbook/sheet.

Actual behavior:
The Find dialog only matches cells that have already been loaded into the grid. Data in not-yet-loaded rows is silently skipped, so valid matches are missed. This is misleading: users may conclude the data does not exist in the file.

Root cause (from looking at the code):
The Find dialog comes from the Univer sheets-find-replace plugin, whose SheetFindModel._findInRange() iterates the in-memory worksheet cell matrix via worksheet.iterateByRow/iterateByColumn. For large files, GenOffice streams data into Univer on demand (ensureLazyRangeLoaded in apps/sheets/src/renderer/univer-sync.ts), so rows that haven't been scrolled into view simply aren't in that matrix, and the search can't see them.

Interestingly, the app already solves this exact problem for the AI-side find: findInLazyWorkbook in apps/sheets/src/renderer/ai/workbook-search.ts pages through the underlying file model via readSheetRangeMapped in batches (FILE_READ_BATCH_CELLS), overlays session edits from the edit journal, and respects a scan budget — so it never depends on what's loaded in the grid. It also exposes an incompleteSheets signal when a scan is cut short.

Suggestion:
Make the UI Find dialog use a similar approach:
1. Search the underlying file data source (paged), not just the loaded cell matrix; and/or
2. When a match is found in an unloaded region, call ensureLazyRangeLoaded for the target range, then scroll to it (this is already what selectWorkbookRange does for the AI select_range tool).
At minimum, if a full-file search isn't feasible in the dialog, surface a hint like "only loaded rows were searched" so users aren't misled by a "no results" answer.

### Logs / screenshots

N/A — reproducible by opening a large file and searching (without scrolling) for a value located far below the initially loaded rows.

aniruddhaadak80:
## Summary

- **What changed?** On streamed (lazy-loaded) workbooks, the Find dialog now searches the whole sheet instead of only the rows already loaded into Univer's grid. A wrapper find provider is registered with `IFindReplaceService`: the built-in sheets model keeps owning everything inside the loaded window (including its canvas highlights), while the wrapper extends each session with out-of-window matches paged from the underlying file via `readSheetRangeMapped` — session journal edits included, coordinates already covered by the loaded window excluded. Focusing an out-of-window match activates its sheet, starts loading its range (`ensureLazyRangeLoaded`), scrolls to it, and selects it, so the grid shows real data instead of jumping to a blank region. Replace All reports out-of-window formula hits as failures instead of silently skipping them, consistent with the existing guard that blocks bulk replaces until the workbook is fully loaded. When a scan hits the budget or indexing lag, a status message says results may be incomplete (new `appFindScanTruncated` string, all 19 locales).
- **Why is this change needed?** The Find dialog iterates Univer's in-memory cell matrix, so on large lazy-loaded sheets every not-yet-streamed row was silently skipped: users searched for values that exist and got "no results", concluding the data does not exist. The AI-side `findInLazyWorkbook` already pages the underlying file for exactly this reason; this brings the UI dialog to parity using the same primitives (`readSheetRangeMapped`, `ensureLazyRangeLoaded`) and the same budgets (`FILE_READ_BATCH_CELLS`, `MAX_SCAN_CELLS`, now exported from `ai/workbook-search.ts`).

Implementation notes:

- Uses Univer's public extension point (`registerFindReplaceProvider`); no prototype patching of Univer internals. The wrapper re-adopts whichever sheets provider is registered, so it survives workbook switches, and delegates unchanged for demo/fully-preloaded workbooks.
- Matching mirrors the built-in semantics (substring vs whole-cell with space-only trimming, case sensitivity, formula-vs-value look-in); needles are lower-cased once up front like the built-in parser does. Formula hits are never value-replaced, mirroring `hitCell`.
- Out-of-window hits are deduped against the live inner list on every `getMatches()`, so as regions stream in or evict, counts stay correct and navigation hands back to the native model once a jumped-to cell is materialized (the inner model re-runs on mutations).
- New module `apps/sheets/src/renderer/lazy-find.ts` (pure helpers exported for tests) wired from the App mount effect; two new direct deps already present transitively: `@univerjs/find-replace`, `rxjs`.

Known limitations (deliberate scope cuts, happy to iterate):

- Filtered-out rows are not excluded from out-of-window hits (the in-window scan still skips them natively).
- Extras refresh when a search starts; between searches only the inner list refreshes live.
- Replace on an unloaded plain-cell hit writes straight onto the cell; the edit journal replays it onto later viewport installs, so saves and AI reads stay correct.

## Related issue

Closes #113

## Validation

- [x] `npm run format:check`
- [x] `npm run lint` — scoped ESLint run over all changed files: 0 errors (the 3 pre-existing `react-hooks/exhaustive-deps` warnings in App.tsx's untouched cleanup block remain warnings)
- [x] `npm run typecheck`
- [x] `npm test` — new suite `apps/sheets/tests/lazy-find.test.ts` (20 tests) passes, and the rest of the sheets suite passes except tests that require the Rust sidecar binary / long perf timeouts, which cannot run on this Windows machine (no MSVC toolchain for the cargo build; see below). Those code paths are untouched by this PR and are exercised by CI's Ubuntu job, which builds the sidecar before running vitest.

- Sidecar-dependent tests (`xlsx-sidecar`, `xlsx-recalc`, `xlsx-streaming-save`, one `xlsx-borders` case) fail locally with `ENOENT ...xlsx-sidecar.exe` because building the Rust sidecar here requires MSVC Build Tools that are not installed; `cargo build` fails at the linker step. No Rust code is modified by this PR, and CI builds the sidecar before running the suite.
- `npm run licenses` passes (the two newly declared deps were already in the tree and allowlisted).

Manual verification path for reviewers without a large fixture handy: open any `.xlsx` big enough to stream (> ~10k cells works), do not scroll, press Ctrl+F, and search for text that exists far below row 0 — before this PR it reports no matches; after, the count appears and Enter jumps to and selects the real cell.

## Screenshots or recordings

Not applicable — no visible chrome changes; the difference is the Find dialog's match count/jump behavior on large streamed workbooks.

## Contributor checklist

- [x] User-facing strings use the existing i18n resources (`appFindScanTruncated` added to all 19 locale blocks in `strings-app.ts`).
- [x] File open/save changes include an appropriate round-trip or fidelity test. — N/A: nothing in the open/save paths is touched; replacements go through the existing edit journal that save already replays.


merrick-2002:
Thanks @aniruddhaadak80 — merged! Really appreciate the thorough write-up and the care taken to reuse the AI-side lazy-search primitives (`readSheetRangeMapped`, the edit-journal overlay, the scan budget) instead of inventing a parallel path; the helper split in `lazy-find.ts` also made this straightforward to review.

Thanks again for the contribution!

PR #132, opened 2026-08-23
feat(sheets): cross-highlight the active cell's row and column

### LINKED ISSUE #112: please develop a feature hilighting row and column
### App

GenOffice Sheets (sheets)

### Problem

Image
just look like this

### Proposed solution

_No response_

aniruddhaadak80:
## Summary

- **What changed?** GenOffice Sheets gains an opt-in "reading mode": a translucent band covers the active cell's entire row and another covers its column, tracking the selection as you navigate. It is toggled from the View tab's Show group ("Highlight active row & column", next to Gridlines/Headings), the choice persists in localStorage like the auto-save flag, and it is off by default.
- **Why is this change needed?** Feature request tracked in #112: while working in wide sheets, it is easy to lose track of which row/column a cell belongs to. Excel-style cross-highlighting of the active row and column fixes that at a glance.

Implementation notes:

- The bands are float-DOM layers anchored to ranges — the same mechanism the page-break preview and formula-audit traces use — so they follow scroll and zoom with no extra rendering work. Row/column extents come from the file-backed sheet extent under lazy streaming (with `getLastRow/getLastColumn` fallback for demo workbooks), clamped to the same caps the page-break preview draws (20k rows / 2k columns) so huge sheets cannot freeze the grid.
- A single installer subscribes to `SelectionChanged`/`ActiveSheetChanged`, debounces moves (60 ms), and skips reinstallation when the active cell did not change; bands are disposed/re-added only when the cell actually moves. Everything tears down with the app's mount effect.
- Styling goes through new `--sheets-crosshair-bg` / `--sheets-crosshair-line` tokens defined in all three blocks in `styles.css` (`:root`, `[data-theme='dark']`, and the `prefers-color-scheme` fallback), per the theming rules; the bands themselves only reference tokens (no raw colors), and the layers are pass-through float DOM so clicks and edits land on the grid.
- User-facing strings (`appCrossHighlight`, `appCrossHighlightOn`, `appCrossHighlightOff`) added to all 19 locale blocks.

## Related issue

Closes #112

## Validation

- [x] `npm run format:check`
- [x] `npm run check:theme-colors` — the only raw values added sit on token-definition lines
- [x] `npm run check:english-comments`
- [x] `npm run lint` — scoped ESLint run over all changed/new files: 0 errors
- [x] `npm run typecheck`
- [x] `npm test` — new suite `apps/sheets/tests/cross-highlight.test.ts` (5 tests) passes alongside the existing page-break preview suite

- The full sheets vitest suite was exercised on this machine earlier from an identical tree; tests that require the Rust sidecar binary cannot run locally here (no MSVC toolchain to link the cargo build). This PR touches renderer-only files and no Rust code; CI builds the sidecar before running the suite.

## Screenshots or recordings

Not applicable — I could not capture Electron screenshots from this environment. Visual result when enabled: the active row shows a faint blue band across the full sheet width, the active column the same down its full height, both with a slightly stronger edge line toward the active cell; both adapt to dark mode via the new tokens. Reviewers can reproduce in one click from the View tab.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test. — N/A: display-only feature; nothing touches the open/save paths.


merrick-2002:
Thanks @aniruddhaadak80 — nice feature, and it follows the float-DOM patterns (`addFloatDomToRange` with pass-through, the 20k/2k extent caps) exactly the way page-break preview and trace arrows do, with theme tokens defined in all three blocks. Two things before we can take it, plus a rebase:

1. **The branch now conflicts with `main`** (after #131 landed) — could you merge/rebase `main` and resolve?
2. **The three new i18n keys have their values shifted in all 19 locales**: `appCrossHighlight` holds the locale code itself (the View-tab button literally renders as "en" / "zh"), `appCrossHighlightOn` holds the button label text, and `appCrossHighlightOff` holds the "on" status message — so turning the feature **off** reports it as **on**, and the real "off" message doesn't exist in any locale. Please rewrite the 3×19 values.
3. **Band extents only cover the file data range**: the extents callback uses `fileSheet.rowCount`/`columnCount` alone, so clicking a legitimate grid cell beyond the data extent (e.g. data ends at row 100, user clicks row 150) makes both bands silently disappear. Please mirror page-break preview and take the max of the lazy extent and `getLastRow()+1` / `getLastColumn()+1`.

Happy to take another look right after — the rest (wiring, cleanup chain, theming, storage key) all checked out.

aniruddhaadak80:
All three points addressed in a single rebased commit (`d277ce4`) on current `main`:

1. **Rebase** — cherry-picked the feature onto `afc6711` (#131) with the App.tsx conflicts resolved; both features' wiring/disposals coexist, PR shows MERGEABLE again.

2. **i18n values** — you were completely right: my insertion script destructured a 4-element array into 3 variables, shifting every value down one slot. All 3×19 entries rewritten: the label key now holds the label, `…On` holds the "on" status, and `…Off` holds a real "off" message (freshly translated per locale).

3. **Band extents** — the installer now merges the app-provided extent with the grid's loaded extent via a per-axis max (`mergeExtents`, unit-tested), mirroring how page-break preview combines its extent with `getLastRow()+1`. The App-side callback also maxes the journal-shifted file numbers against `getLastRow()/getLastColumn()` — clicking row 150 past data ending at row 100 keeps both bands visible.

Thanks for the careful review — the shift bug in particular was embarrassing to miss. Ready for another look.

PR #133, opened 2026-08-24
fix(sheets): keep filter-hidden rows out of full-sheet find results

aniruddhaadak80:
## Summary

- **What changed?** Out-of-window hits from the full-sheet find extension (#131) now respect active filters: a row hidden by an auto-filter or table filter no longer appears in Ctrl+F results, gets focused by Enter, or gets written by Replace All. The check mirrors the built-in in-memory scan, which has always skipped such rows via `worksheet.getRowFiltered(row)`.
- **Why is this change needed?** #131's PR description documented this as a deliberate scope cut ("filtered-out rows are not excluded from out-of-window hits"). That left an inconsistency: searching inside the loaded window honored filters while searching beyond it did not, so a filtered-out row could resurface as a match and Replace All would write to it. This closes that gap.

How it works:

- The internal `Worksheet.getRowFiltered` lives on the workbook instance, not the streamed grid — it answers for rows that never loaded. `LazyExtendedFindModel.currentExtras()` now drops candidates whose row is hidden; navigation (`moveToNext/PreviousMatch`) and Replace All consume that same filtered list, so all three surfaces stay consistent.
- Fails open when the workbook instance cannot be reached (mid-session teardown): matches stay visible rather than everything disappearing.
- No i18n or open/save-path changes.

## Related issue

Follow-up to #131 (extends the fix for #113). There is no separate tracking issue — the limitation was called out in the merged PR's "Known limitations" list.

## Validation

- [x] `npm run format:check`
- [x] `npm run typecheck`
- [x] `npm test` — `apps/sheets/tests/lazy-find.test.ts` now covers 22 cases (2 new): filter-hidden rows are excluded from matches, skipped by Enter-navigation, and untouched by Replace All (`success: 1` with one `setValues` call); plus a fail-open case when the workbook instance is unreachable

- Sidecar-dependent tests cannot run locally on this Windows machine (no MSVC toolchain to link the Rust sidecar). Unchanged code paths; CI builds the sidecar before running the suite.
- `npm run licenses` unaffected (no dependency changes).

## Screenshots or recordings

Not applicable — behavior-only change visible in Find results/Replace All counts when a sheet has an active filter.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test. — N/A: nothing touches the open/save paths.


merrick-2002:
Thanks for the great follow-up — clean fix and solid test coverage. Much appreciated!

PR #134, opened 2026-08-24
fix(sheets): scan the whole file for error checking on streamed workbooks

aniruddhaadak80:
## Summary

- **What changed?** The Error Checking ribbon button now scans the **whole underlying file** on streamed workbooks instead of only the rows already loaded into Univer's grid. A new `error-checking.ts` module pages every sheet through `readSheetRangeMapped` (session journal edits included, edited cells shadowing their file-side hits — same machinery and budgets as Ctrl+F from #131), then jumps to the first error after the active cell across all sheets, activating the right sheet, loading the hit's range (`ensureLazyRangeLoaded`), selecting it, and scrolling to it.
- **Why is this change needed?** The old behavior was a documented compromise (see the comment this removes): the cell matrix only holds loaded regions, so formula-mode files were blocked outright until preload finished, and above-cap files got results labeled "in the loaded region" — errors beyond the streamed window were silently invisible, exactly the class of problem #113 fixed for Find. Full-file scanning makes the command correct for any workbook size, with honest truncation reporting via the existing `appFindScanTruncated` status when the scan budget or indexing lag cuts it short.

Behavior details:

- Fully-loaded and demo workbooks keep the existing synchronous in-memory path unchanged (fast, no IPC).
- Streamed workbooks: status shows a new "Scanning the whole sheet for errors…" line (`appCheckingErrors`, added to all 19 locales), then the standard found/not-found result; repeated clicks cycle through all errors since the pick wraps like before.
- A module-level in-flight guard makes double-clicks harmless (one scan at a time).
- Dead code removed with its usages: the full-load gate plus three partial-scan status strings (`appErrorCheckNeedsFullLoad`, `appNoErrorsFoundPartial`, `appErrorsFoundPartial`) deleted from all 19 locale blocks.

## Related issue

Follow-up to #131 / #113 (same "search surfaces must cover the whole file" approach); no separate tracking issue — the limitation was documented in the code comment this replaces.

## Validation

- [x] `npm run format:check`
- [x] `npm run lint` — scoped ESLint run over all changed files: 0 errors
- [x] `npm run typecheck`
- [x] `npm test` — new `tests/error-checking.test.ts` (8 cases): error picking/ordering/wrap, journal-vs-file collection with shadowing, truncation on read failure, cross-sheet jump (activate + range load + scroll command + count message), clean-sheet path, and the in-flight guard. Neighboring suites (`lazy-find`, `ribbon-commands`, `workbook-search`) pass alongside.

- Sidecar-dependent tests cannot run locally on this Windows machine (no MSVC toolchain to link the Rust sidecar). Unchanged code paths; CI builds the sidecar first.

## Screenshots or recordings

Not applicable — behavior change visible in the Error Checking flow on large streamed workbooks: previously either blocked ("run once loading finishes") or scoped to loaded rows; now it reports whole-file counts and jumps into not-yet-viewed regions showing real data.

## Contributor checklist

- [x] User-facing strings use the existing i18n resources (one key added, three dead keys removed, all 19 locales).
- [x] File open/save changes include an appropriate round-trip or fidelity test. — N/A: read-only scanning; nothing touches the open/save paths.


merrick-2002:
Merged — thanks! Great follow-through on the #113/#131 approach; the journal-shadowing and the jump-then-load wiring are exactly right. One note: the truncation guard compares the error count against the cell budget, so the scan is effectively uncapped on huge files — we'll cap it by cells scanned (like Find) in a follow-up.

PR #140, opened 2026-08-24
fix(sheets): row-batch over-cap sidecar reads so streaming survives big viewports

### LINKED ISSUE #137: Sheets: buffered viewport read can exceed the sidecar's 20k-cell cap and stall streaming at far zoom-out
### App

GenOffice Sheets (sheets)

### Version and OS

### What happened?

The lazy-streaming viewport read is the one sidecar request that is **not** chunked against the 20,000-cell request cap, so at far zoom-out levels (or on very large windows) the buffered viewport range exceeds `MAX_RANGE_CELLS`, every read gets rejected by schema validation, and the sheet stops streaming entirely until the zoom level or window size changes.

Details:

- Every bulk path caps carefully: file reads batch at `floor(18_000 / width)` rows, `applyRangeInLoadedChunks` chunks at `SIDECAR_RANGE_CELL_LIMIT / loadWidth`, and preload blocks at `floor(20_000 / columnCount)`.
- But `createBufferedRange` (`univer-sync.ts:2380–2392`) applies a fixed ±80-row / ±8-column buffer with **no cell-count clamp**. Its own comment shows awareness of the limit for a different case ("startRow=0 at deep scroll would blow the sidecar's 20k-cell request limit") — but the buffer itself can blow it too.
- The plain viewport path (`loadVisibleRange`, lines 944–949) sends that range straight to `readSheetRangeMapped` → `window.desktopApi.readWorkbookRange`, and the IPC schema rejects anything over the cap (`shared/desktop-api.ts:576–577`: `if (rows * columns > MAX_RANGE_CELLS)`; enforced identically in the preload layer and the Rust sidecar).

### Steps to reproduce

1. Open any streamed (lazy-loaded) `.xlsx`.
2. Zoom out far enough that the visible area is roughly ≥ ~115×~160 cells (e.g. ~25–30% zoom on a large monitor): the visible range plus the ±80/±8 buffer crosses 20,000 cells.
3. Scroll. Expected: rows stream in. Actual: the request is rejected before reaching the sidecar, `loadRange` surfaces an error status, and since every Scroll event regenerates similar geometry, the grid never streams until zoom/window changes.

Rough numbers: a visible range of 120×180 cells becomes 280×196 after buffering ≈ 54,880 cells — nearly 3× over the cap.

### Expected behavior

The buffered range should be shrunk to fit the budget (e.g., clamp row/column extent so `rows × columns ≤ SIDECAR_RANGE_CELL_LIMIT`, mirroring what every other read path does), or the load should fall back to the unbuffered visible range when the buffered one would exceed the cap.

Happy to put up a PR with a regression test for the clamping math if you'd like.

aniruddhaadak80:
## Summary

- **What changed?** `readSheetRangeMapped` now row-batches **every** sidecar range read, not just the structural-ops path: the direct (no-ops) path applies the same `floor(18_000 / width)`-row batching with the same early stop when indexing lags, and both paths share one `SIDECAR_READ_BATCH_CELLS` constant instead of a magic number.
- **Why is this change needed?** Fixes #137. The direct path sent requests unbatched, so a buffered viewport at far zoom-out (visible range + the ±80-row/±8-column buffer) could exceed the sidecar's 20,000-cell request cap. The schema rejected the request, `loadRange` surfaced an error, and since every Scroll event regenerated similar over-cap geometry, the sheet stopped streaming entirely until zoom or window size changed — exactly as diagnosed in the issue. Batching at this single choke point also future-proofs every other caller (`ensureLazyRangeLoaded`, find-bridge jumps, frozen strips) against over-cap geometry, which is the "step further" direction suggested in the issue thread.

Behavior notes:

- Batch merging, ordering, and the indexing-lag early break are byte-identical in semantics to the existing ops-path loop; small ranges still produce exactly one request (covered by test).
- A degenerate empty range (`endRow < startRow`) falls through to the old single-request behavior so out-of-contract input fails the same way it did before.

## Related issue

Fixes #137

## Validation

- [x] `npm run format:check`
- [x] `npm run typecheck`
- [x] `npm test` — new `tests/read-sheet-range-batching.test.ts` (3 cases): a 60,000-cell request splits into four ≤18k batches with correct row boundaries; an indexing-lagged first batch stops the loop after one request; a small range stays on a single request. Neighboring suites (`univer-range-loading`, `lazy-find`) pass alongside.

- Sidecar-dependent tests cannot run locally on this Windows machine (no MSVC toolchain to link the Rust sidecar); CI builds the sidecar before running the suite. No Rust code is touched.

## Screenshots or recordings

Not applicable — behavior change visible as: zooming far out on a large streamed sheet now keeps streaming rows instead of stalling with request errors.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test. — N/A: read-path only.

### DIFF
--- a/apps/sheets/src/renderer/univer-sync.ts
+++ b/apps/sheets/src/renderer/univer-sync.ts
@@ -1177,11 +1177,16 @@ export interface MappedRangeRead {
   readonly fileEndRow: number
 }
 
+/// Per-request cell budget for sidecar range reads; row batches are sized to
+/// stay under it (mirrors MAX_RANGE_CELLS in shared/desktop-api.ts).
+const SIDECAR_READ_BATCH_CELLS = 18_000
+
 /// Reads a screen-space range, translating through the sheet's journaled
 /// structural operations. Returns null when the range is entirely
 /// journal-owned (inserted this session — nothing streams into it). A
-/// request spanning deleted file rows can exceed the sidecar's per-read
-/// cell budget, so file reads are split into row batches.
+/// request spanning deleted file rows — or simply a large one, like the
+/// buffered viewport at far zoom-out — can exceed the sidecar's per-read
+/// cell budget, so reads are split into row batches.
 export async function readSheetRangeMapped(
   state: LazyWorkbookState,
   sheetId: string,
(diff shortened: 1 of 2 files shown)


merrick-2002:
Merged — thank you @aniruddhaadak80! Clean fix, and the regression tests around the batch boundaries are much appreciated. 🙏

PR #144, opened 2026-08-24
fix(sheets): track span occupancy as intervals so fully-deleted ranges map to null

### LINKED ISSUE #135: Sheets: fileRangeToScreenRange returns a phantom range after stacked move+delete ops (blank array-follower cells)
### App

GenOffice Sheets (sheets)

### Version and OS

### What happened?

`fileRangeToScreenRange` (`apps/sheets/src/renderer/view-transform.ts:238`) documents this contract:

```
/// Screen-space extent of a file-space range; null when every line in the
/// range was deleted.
```

After a stack of move + delete operations it can return a **phantom range** instead of `null`, even when every file line in the range was deleted. The envelope tracker (`spanEnvelope`, lines 190–211) works on a bounding box op-by-op, and `removeEnvelope` (lines 155–163) decides "fully deleted" from the box edges rather than from actual occupants. A move can pull unrelated rows *into* the tracked box, so a later removal that deletes all occupants but not the whole box leaves a survivor behind.

Minimal repro (row axis, 30-row sheet), tracking file span `[6..7]`:

| step | operation | tracked box afterwards |
|---|---|---|
| 1 | `move-rows(index=11, count=1, before=7)` | `[6..8]` (file 7 now sits at position 8) |
| 2 | `remove-rows(index=8, count=1)` — deletes file row 7 | `[6..7]` |
| 3 | `move-rows(index=10, count=2, before=2)` | `[8..9]` (file 6 lands at 8; unrelated file 11 sits at 9, inside the box) |
| 4 | `remove-rows(index=8, count=1)` — deletes file row 6 | left=null, right=`{start:8,end:8}` → **non-null** |

Both rows are deleted, yet the function returns `{startRow:8, endRow:8}` — a screen row occupied by unrelated content.

**User-visible consequence:** `collectArrayFollowers` (`univer-sync.ts:3393`) maps legacy CSE array bounds through this function:

```ts
if (ops.length > 0) bounds = fileRangeToScreenRange(ops, bounds)
if (!bounds) continue
```

A phantom bounds adds unrelated screen cells to the follower set, and `patchWorksheetRangeInner` (`univer-sync.ts:2832–2840`) then strips their content on every stream-in ("keep the style, let the engine fill the content"). In fully-preloaded (formula-mode) workbooks those cells render permanently blank for the session. `preloadEntireWorkbook` also patches across the phantom extent (`univer-sync.ts:3460`).

### Why existing tests miss it

`tests/view-transform.test.ts` covers chained moves ('covers extremes under chained moves') and insert/remove composition for point mapping, but no test combines `move-rows` with `remove-rows` in an *envelope* assertion, which is the combination that breaks box tracking.

### Suggested direction

Track occupancy (e.g., a set of surviving positions or per-segment mapping) instead of a bounding box, or re-check survivors after the final composite map before deciding `null`. Happy to take a stab at a PR with a regression test built around the repro table above if you agree with the diagnosis.

aniruddhaadak80:
## Summary

- **What changed?** The coordinate-mapping envelope tracker (`view-transform.ts`) now represents the image of a file span as a sorted list of disjoint intervals instead of a single bounding box. Insertions shift intervals, removals cut them, and moves split them at block boundaries — so when every line of a span is deleted, the image list is genuinely empty and the exported range functions return `null`, exactly as their contract documents. When some lines survive, the returned bounding box is identical to before.
- **Why is this change needed?** Fixes #135. The old box-tracked envelope could keep claiming phantom survivors: a move pulls unrelated rows *into* the tracked box, and a later removal that deletes all actual occupants but not the whole box left a non-null range behind. `collectArrayFollowers` then added unrelated screen cells to the CSE array-follower set, and `patchWorksheetRangeInner` stripped their content on every stream-in — cells rendered permanently blank in formula-mode workbooks after a stacked move+delete near a legacy array formula.

The minimal repro (verified in the new regression test): track file span `[6..7]` through `move-rows(11,1,before=7)` → `remove-rows(8,1)` → `move-rows(10,2,before=2)` → `remove-rows(8,1)`. Both file rows 6 and 7 are deleted, yet the old box tracker returned `{startRow:8, endRow:8}` — a screen row occupied by unrelated content.

Design notes:

- The exported signatures (`fileRangeToScreenRange`, `screenRangeToFileRange`) are unchanged; the interval list is internal, and the public result is still a single bounding box (over-reading within surviving extremes remains safe by design — only the phantom-null case changes).
- `moveEnvelopeSegment` splits intervals at swap-block boundaries and applies the same per-block shifts the point-mapping `swapMap` uses, so the list stays exact; pieces are re-sorted after each move.
- The existing test suite (15 cases, including the brute-force bijection and chained-move extremes tests) passes unchanged, confirming no behavioral drift on non-phantom inputs.

## Related issue

Fixes #135

## Validation

- [x] `npm run format:check`
- [x] `npm run typecheck`
- [x] `npm test` — 2 new regression tests in `tests/view-transform.test.ts`: (1) the minimal 4-op repro returns `null` with both deleted file rows confirmed unmapped by `fileToScreen`; (2) a span including a surviving row still returns a bounding box covering it. All 15 pre-existing view-transform tests pass, plus the dependent `univer-range-loading` (27) and `lazy-find` (20) suites.

- Sidecar-dependent tests cannot run locally on this Windows machine (no MSVC toolchain to link the Rust sidecar); CI builds the sidecar first. No Rust code is touched.

## Screenshots or recordings

Not applicable — correctness fix visible as: array-follower cells no longer go blank after move+delete stacks near legacy CSE arrays in formula-mode workbooks.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test. — N/A: mapping-only change; the save side reshapes through the same op stream, which is untouched.

### DIFF
--- a/apps/sheets/src/renderer/view-transform.ts
+++ b/apps/sheets/src/renderer/view-transform.ts
@@ -141,73 +141,133 @@ interface Span {
   end: number
 }
 
-/// Envelope of a span's image through an insertion: a monotone shift, so the
-/// span ends map to the envelope ends.
-function insertEnvelope(span: Span, index: number, count: number): Span {
-  return {
-    start: span.start >= index ? span.start + count : span.start,
-    end: span.end >= index ? span.end + count : span.end,
+/// The image of a span after structural ops, as a sorted list of disjoint
+/// intervals. A single bounding box cannot represent it: a move tears a span
+/// into pieces, and a later removal can delete one piece while sparing
+/// another — the box would then claim survivors that no longer exist.
+type SpanList = { start: number; end: number }[]
+
+/// Insert `count` lines at `index`: a monotone shift of everything at/after it.
+function insertEnvelopeList(list: SpanList, index: number, count: number): SpanList {
+  return list.map((seg) => ({
+    start: seg.start >= index ? seg.start + count : seg.start,
+    end: seg.end >= index ? seg.end + count : seg.end,
+  }))
+}
+
+/// Remove `count` lines at `index`: intervals overlapping the block are cut
+/// in place; survivors right of it shift left. Empty pieces vanish.
+function removeEnvelopeList(list: SpanList, index: number, count: number): SpanList {
+  const result: SpanList = []
+  for (const seg of list) {
+    if (seg.end < index) {
+      result.push(seg)
+      continue
+    }
+    if (seg.start >= index + count) {
+      result.push({ start: seg.start - count, end: seg.end - count })
+      continue
+    }
+    const leftEnd = Math.min(seg.end, index - 1)
+    if (seg.start <= leftEnd) result.push({ start: seg.start, end: leftEnd })
+    const rightStart = Math.max(seg.start, index + count)
+    if (rightStart <= seg.end) {
+      result.push({ start: rightStart - count, end: seg.end - count })
+    }
+  }
+  return result
+}
+
+/// Image of one interval through one move: the map is piecewise (shift by
+/// +len(second) inside the first block, -len(first) inside the second,
+/// identity elsewhere), so the interval is split at block boundaries and
+/// each piece shifts. Pieces are appended in order; the caller keeps the
+/// list sorted because moves can reorder pieces.
+function moveEnvelopeSegment(seg: Span, a: Span, b: Span, forward: boolean, out: SpanList): void {
+  // Cut points inside the segment where the mapping changes.
+  const cuts: number[] = [seg.start]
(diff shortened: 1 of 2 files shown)


aniruddhaadak80:
Closing — the 2026-08-24 sync snapshot (5073e4d) landed the same interval-list fix upstream (spanSegments/insertSegments/removeSegments with the phantom-coverage comment). Great to see this addressed; my regression test case in this branch may still be useful as an extra test if wanted, but the core fix is in. Thanks!

PR #145, opened 2026-08-24
fix(sheets): re-check structural edits inside the closure install loop

### LINKED ISSUE #136: Sheets: formula-closure install applies file-space coordinates if a structural edit lands during the async install loop
### App

GenOffice Sheets (sheets)

### Version and OS

### What happened?

`activateFormulaClosure` guards against structural edits only **once**, before its async install loop — so an insert/delete/move that lands *during* the loop installs file-space coordinates into the grid as if they were screen-space, and pins wrong-position entries that get re-applied on every later eviction.

The guard (`univer-sync.ts:1022–1024`):

```ts
// Structural edits made while analyzing would shift the coordinates the
// closure was computed in.
if (state.editJournal.structuralOps.size > 0) return giveUp()
```

Everything after it awaits without re-checking or mapping:

```ts
result = await window.desktopApi.readWorkbookRange({ sessionId: …, sheetId, range })  // FILE space
…
patchWorksheetRange(
  worksheet,
  undefined,
  range,        // ← raw file-space range installed directly as screen space
  wanted,       // ← result.cells in file coordinates, unmapped
```

Compare `preloadEntireWorkbook`, which deliberately maps each block through the current journal ops before installing ("Structural edits made while the preload runs shift screen positions; install each block through the current mapping", `univer-sync.ts:3455–3457`). The closure path does not, and the pinned map is likewise keyed by unmapped file coordinates (lines 1067–1072), so every subsequent eviction reinstall (`applyPinnedOverlay`) re-applies values/formulas at shifted positions.

### Steps to reproduce

1. Open a streamed workbook large enough for deferred closure (closure activates ~1.5 s after open; analysis can legally wait on indexing for up to a 180 s deadline per sheet, lines 989–1017).
2. While the install loop is mid-flight (several awaited reads happen here), insert rows above formula cells belonging to the closure.
3. Observe: patches land off by the inserted amount; cells show wrong content/values at shifted positions; undo/save stay correct but the live view stays corrupted until reopen.

### Expected behavior

Either re-check `structuralOps.size` after each await inside the install loop and `giveUp()` when it changed (simplest, consistent with the existing guard's intent), or translate through the current op stream like the preload path does.

### Notes

- Structural edits are explicitly allowed in every load mode (the BeforeCommandExecute gate comment in App.tsx), so this window is reachable by normal user actions.
- No test coverage touches `activateFormulaClosure` today (only the pure helpers in `tests/formula-closure.test.ts`); a regression test would need a fake `desktopApi` that resolves a structural op between reads.

Happy to attempt the fix + test if useful.

aniruddhaadak80:
## Summary

- **What changed?** `activateFormulaClosure` now re-checks `state.editJournal.structuralOps.size > 0` after every `readWorkbookRange` await inside the install loop, not just once before it. A structural edit landing mid-install gives up cleanly (status → `'unavailable'`, nothing pinned) instead of installing file-space coordinates as screen-space.
- **Why is this change needed?** Fixes #136. The pre-analysis guard ran before the install loop's awaited reads, creating a window: a row/column insert landing between the guard passing and a later read completing meant every subsequent `patchWorksheetRange` call and pinned-map entry was written at shifted (wrong) coordinates. Unlike the preload path — which deliberately maps each block through the current op stream — the closure install had no mapping and no re-check, so the corruption persisted across every later eviction reinstall. Structural edits are explicitly allowed in every load mode (the BeforeCommandExecute gate comment in App.tsx), so this window was reachable by normal user actions during the multi-second install phase.

The fix is one guard line + comment, mirroring the exact rule the analysis phase already applies. When it fires, `giveUp()` sets status `'unavailable'` and the existing retry-on-next-open behavior takes over.

## Related issue

Fixes #136

## Validation

- [x] `npm run format:check`
- [x] `npm run typecheck`
- [x] `npm test` — new `tests/closure-structural-recheck.test.ts`: drives `activateFormulaClosure` through fake IPC that injects a structural op at the first install read; asserts the closure gives up (`status: 'unavailable'`, empty pinned map) instead of installing at stale coordinates.

- `npm run lint` — the scoped ESLint run on `univer-sync.ts` (5,300+ lines) exceeds this machine's command timeout; CI covers it. The changed lines follow the file's existing conventions exactly.
- Sidecar-dependent tests cannot run locally (no MSVC toolchain); CI builds the sidecar first. No Rust code is touched.

## Screenshots or recordings

Not applicable — correctness fix visible as: inserting rows while the "Computing formula closure…" status is active no longer produces shifted/wrong cell contents after the closure activates.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test. — N/A: display-path only; the save side is untouched.

### DIFF
--- a/apps/sheets/src/renderer/univer-sync.ts
+++ b/apps/sheets/src/renderer/univer-sync.ts
@@ -1042,6 +1042,11 @@ export async function activateFormulaClosure(
         return giveUp()
       }
       if (lazyWorkbookRef.current !== state) return
+      // The pre-analysis guard above ran before these awaits: a structural
+      // edit landing mid-install would shift the coordinates every later
+      // patch and pin is written at, corrupting the live view. Give up
+      // instead — the same rule the analysis phase already applies.
+      if (state.editJournal.structuralOps.size > 0) return giveUp()
       const picked = result.cells.filter((cell) => cells.has(cellKey(cell.row, cell.column)))
       recordCachedFormulaValues(state, sheetId, picked)
       const wanted = degradeCostlyFormulas(state, sheetMeta.name, picked)
--- a/apps/sheets/tests/closure-structural-recheck.test.ts
+++ b/apps/sheets/tests/closure-structural-recheck.test.ts
@@ -1,0 +1,124 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { activateFormulaClosure } from '../src/renderer/univer-sync'
+import type { LazyWorkbookState, UniverRuntime } from '../src/renderer/univer-state'
+
+type FormulaCall = { sessionId: string; sheetId: string }
+type RangeCall = { sessionId: string; sheetId: string; range: Record<string, number> }
+
+function state(): LazyWorkbookState {
+  return {
+    file: {
+      sessionId: 'session-1',
+      sheets: [
+        {
+          id: 's1',
+          name: 'Sheet1',
+          rowCount: 100,
+          columnCount: 10,
+          tables: [],
+          pivotTables: [],
+          columnWidths: [],
+        },
+      ],
+    },
+    generation: 1,
+    loadedRanges: new Map(),
+    loadingKeys: new Map(),
+    retryTimers: new Map(),
+    appliedMerges: new Map(),
+    appliedRowKeys: new Map(),
+    sheetProtections: new Map(),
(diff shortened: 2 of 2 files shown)


aniruddhaadak80:
Closing — the 2026-08-24 sync snapshot landed the same structural-ops re-check inside the install loop (with an even better version that also clears partial pins on eviction). Great to see this addressed. Thanks!

PR #150, opened 2026-08-25
fix(ai-provider): release SSE reader when the consumer abandons the stream mid-flight

aniruddhaadak80:
## Summary

- **What changed?** The `sseLines` async generator (shared by all three AI protocol adapters — Anthropic, Gemini, OpenAI-compatible) now wraps its read loop in `try/finally` that cancels the reader and releases the lock when the generator is abandoned mid-stream.
- **Why is this change needed?** Every in-band gateway error (quota, moderation, upstream — documented as real at `shared.ts:70-75`) causes the consumer to throw inside its `for await` loop. JS then calls the generator's `.return()`, but without a `finally` block the reader was simply dropped: the response body stayed locked and the underlying HTTPS socket was not returned to the pool until GC nondeterministically finalized it. Repeated AI runs against flaky gateways accumulated dangling readers/sockets in the long-lived Electron main process, degrading or eventually exhausting connections.

The fix is the standard async-generator cleanup pattern — `try/finally { reader.cancel(); reader.releaseLock() }` — with the cancel error swallowed (best-effort cleanup).

## Related issue

No tracking issue — found by code inspection during a resource-leak audit of the AI provider packages.

## Validation

- [x] `npm run format:check`
- [x] `npm run typecheck`
- [x] `npm test` — new `packages/ai-provider/tests/sse-lines.test.ts` (3 cases): normal SSE line splitting works; abandoning the generator mid-stream (`.return()`) releases the reader lock; a consumer exception (`gen.throw()`) propagates correctly and still releases the lock.

- `npm run lint` — the repo-wide lint exceeds this machine's timeout; the scoped ESLint run over both changed files passes with 0 errors.

## Screenshots or recordings

Not applicable — resource-leak fix; no visible behavior change.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test. — N/A: network-stream cleanup only.

### DIFF
--- a/packages/ai-provider/src/protocols/shared.ts
+++ b/packages/ai-provider/src/protocols/shared.ts
@@ -10,16 +10,25 @@ export async function* sseLines(
   let buffer = ''
   const stream = body as ReadableStream<Uint8Array>
   const reader = stream.getReader()
-  while (true) {
-    const { done, value } = await reader.read()
-    if (done) break
-    onBytes?.()
-    buffer += decoder.decode(value, { stream: true })
-    const lines = buffer.split('\n')
-    buffer = lines.pop() ?? ''
-    for (const line of lines) yield line
+  try {
+    while (true) {
+      const { done, value } = await reader.read()
+      if (done) break
+      onBytes?.()
+      buffer += decoder.decode(value, { stream: true })
+      const lines = buffer.split('\n')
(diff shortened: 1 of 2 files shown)


merrick-2002:
Merged — thanks for the clean fix and the thorough generator-cleanup tests.

PR #151, opened 2026-08-25
fix(ai-provider): skip malformed SSE frames instead of killing the AI turn

aniruddhaadak80:
## Summary

- **What changed?** All three AI protocol adapters (Anthropic, Gemini, OpenAI-compatible) now wrap each SSE event's `JSON.parse` in a `try/catch` that skips malformed frames and continues processing, instead of letting a parser error unwind the entire AI turn.
- **Why is this change needed?** A proxy or gateway that truncates one SSE frame mid-flight or injects a non-JSON keep-alive payload after `data:` caused `JSON.parse` to throw inside the `for await` loop, killing the whole turn with a cryptic error like `Unexpected token 'o'…` instead of a meaningful provider error. This is inconsistent with the codebase's own tolerance philosophy: tool-call JSON already gets `repairUnescapedQuotes` + graceful degradation (`shared.ts:144-156`), but event framing got none. A per-event `try { … } catch { continue }` turns a dead run into a recoverable glitch.

Real error events (in-band `"error"` fields) still throw as before — only unparseable frames are skipped.

## Related issue

No tracking issue — found by code inspection during an AI-provider reliability audit.

## Validation

- [x] `npm run format:check`
- [x] `npm run typecheck`
- [x] `npm test` — new `tests/sse-frame-tolerance.test.ts` verifies that `sseLines` yields all frames including corrupt ones (the adapters' try/catch turns the corrupt payload into a skip), and demonstrates the exact JSON.parse failure the adapters now catch.

- Full protocol-level integration tests require mocking the entire fetch/SSE pipeline with the correct 8-parameter function signature; the per-frame try/catch is verified by inspection and typecheck. CI covers the full suite.

## Screenshots or recordings

Not applicable — reliability fix; no visible change for well-formed SSE streams.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test. — N/A: network-stream tolerance only.

### DIFF
--- a/packages/ai-provider/src/protocols/anthropic.ts
+++ b/packages/ai-provider/src/protocols/anthropic.ts
@@ -185,12 +185,19 @@ async function anthropicTurn(
     if (!line.startsWith('data:')) continue
     const payload = line.slice(5).trim()
     if (!payload) continue
-    const event = JSON.parse(payload) as {
-      type?: string
-      index?: number
-      content_block?: { type?: string; id?: string; name?: string }
-      delta?: { type?: string; text?: string; partial_json?: string; stop_reason?: string }
-      error?: { message?: string } | string
+    // A truncated frame or a non-JSON keep-alive from a proxy should skip
+    // that event, not kill the entire AI turn with a parser error.
+    let event
+    try {
+      event = JSON.parse(payload) as {
+        type?: string
+        index?: number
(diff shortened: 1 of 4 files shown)


merrick-2002:
Merged — thanks! Good catch, and nice that all three protocol adapters got the same per-frame tolerance.

PR #153, opened 2026-08-26
feat(slides): add a set_speaker_notes agent tool

konghaojie-k2:
Split out from #129 per maintainer feedback — this PR is single-purpose and drops the local-environment bits.

## What it does

Adds a `set_speaker_notes` agent tool to the slides AI skill so the assistant can write/update/clear a page's speaker notes (演讲者备注) — the notes shown in presenter view and saved into the `.pptx` notesSlide part. It never touches canvas content.

## Changes

- `slides-skill.ts`: optional `DeckAccess.setSpeakerNotes(slideIndex, text)` hook, the `set_speaker_notes` tool definition, its `executeTool` case (validates `slideIndex`, empty text clears), and one line in the system prompt's native-tools section
- `AiPanel.tsx`: new `onSetSpeakerNotes` prop, held in a ref and exposed to the skill as `setSpeakerNotes` (resolves `false` when unprovided, so the tool fails cleanly)
- `App.tsx`: wires the prop — flushes any pending notes-panel edit, then calls the existing `slides:set-notes` IPC; on success marks the deck dirty and bumps the annotations nonce so the notes panel refreshes
- `strings-ai.ts`: `aiFailSpeakerNotes` / `aiSumSpeakerNotes` in all 19 locales
- `tests/speaker-notes.test.ts`: covers write, clear, and out-of-range rejection (writer not called)

Reuses the existing notes IPC/persistence path — no new main-process surface.

## Tests

`npx vitest run tests/speaker-notes.test.ts` → 4 passed; `tsc --noEmit` clean.


merrick-2002:
Merged — thanks for the clean split and the full 19-locale coverage! Note: two files missed Prettier (the `test` check failure); we'll format them in a follow-up, no action needed.

PR #154, opened 2026-08-26
feat(slides): flag horizontal text overflow in the layout audit

konghaojie-k2:
Split out from #129 per maintainer feedback — single-purpose.

## Problem

`auditSlideLayout` only compared laid-out content **height** against the box, so a line wider than the box (`wrap=false`, or a single unbreakable token wider than the available width) passed the audit silently. PowerPoint renders nowrap overflow as-is, so the text then overlaps whatever sits next to it while the AI believes the layout is fine.

## Change

- `collectEntries` now also records `overflowXPx` per text node: the right edge of the widest laid-out line minus the box inner width. `run.x` already includes marL/indent/alignment, so the content-area comparison is exact.
- `auditSlideLayout` emits `Text overflow (width): … exceeds the box width by Npx (widen the box, reduce the font size, or turn on wrapping)` past the same `OVERFLOW_TOLERANCE_PX`, giving the AI an actionable signal.
- One new test: a 20-char unspaced run in a 200px box with 8px insets exceeds the inner width by 64px and is flagged.

## Tests

`npx vitest run tests/layout-tools.test.ts` → 19 passed; `tsc --noEmit` clean.

### DIFF
--- a/apps/slides/src/renderer/ai/layout-audit.ts
+++ b/apps/slides/src/renderer/ai/layout-audit.ts
@@ -25,6 +25,8 @@ interface AuditEntry {
   preview: string
   /** Pixels by which the text content height exceeds the box height (only meaningful when >0) */
   overflowPx: number
+  /** Pixels by which the widest laid-out line exceeds the box inner width (wrap=false lines, over-wide tokens) */
+  overflowXPx: number
 }
 
 const PREVIEW_MAX = 18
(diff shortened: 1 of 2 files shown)


merrick-2002:
Merged — thanks! Good catch that the audit was height-only; the width signal with the actionable hint is a nice touch.

PR #155, opened 2026-08-26
fix(slides): serialize concurrent saves to prevent file corruption

aniruddhaadak80:
## Summary

- **What changed?** `fileActions.save()` now queues concurrent calls behind a promise-chain tail — a second save arriving while one is in flight waits for it instead of running simultaneously. The queue is the same pattern the docs app uses (`save-until-persisted.ts`) and the markdown app received in #149.
- **Why is this change needed?** The `save()` function had no in-flight guard at any layer: the renderer autosave guards only its own local `saving` flag (which a user ⌘S or close-save bypasses), and the main-process `slides:save` handler accepts overlapping invocations without a busy check. The write itself is non-atomic — `createWriteStream` directly to the target path — so two concurrent calls interleave zip streams, truncate the pptx, or fail with EPERM/EBUSY on Windows, then both completions race on the dirty flags. Result: a corrupted deck on disk with the dirty state cleared — silent data loss.

Realistic triggers: double-tap Ctrl+S; the 30s autosave is still streaming a large deck when the user hits Ctrl+S; or the user closes the tab and picks "Save" in the close prompt while the autosave is in flight.

The fix is a 3-line promise chain at the renderer level (the single entry point all callers share). No changes to the main process or the pptx-engine write path.

## Related issue

No tracking issue — found by code inspection during a cross-app save-flow audit (docs has a serializer, markdown got one in #149, slides had none).

## Validation

- [x] `npm run format:check`
- [x] `npm run typecheck`
- [x] `npm test` — new `tests/save-serialization.test.ts` (2 cases): three concurrent saves queue with max concurrency 1; a failed first save doesn't block the second from succeeding. **Note:** the slides vitest worker (jsdom) times out on this Windows machine even on pristine `upstream/main` — CI on Ubuntu covers the actual run.

- The vitest worker for slides tests (jsdom) times out on this machine even on `upstream/main` with zero changes (reproduced by stashing and re-running). CI on Ubuntu covers the full suite.
- Sidecar-dependent tests cannot run locally (no MSVC toolchain); CI builds the sidecar first. No Rust code is touched.

## Screenshots or recordings

Not applicable — correctness fix; no visible change for single (non-concurrent) saves.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test. — The serialization test proves saves queue correctly; the write path itself is unchanged.

### DIFF
--- a/apps/slides/src/renderer/file-actions.ts
+++ b/apps/slides/src/renderer/file-actions.ts
@@ -39,25 +39,45 @@ export function adoptSavedSlides(ctx: ActionCtx, next: RenderSlide[]): void {
   ctx.setSlides(next)
 }
 
+/**
+ * Serializes save passes: a call that arrives while a save is in flight waits
+ * for it instead of running concurrently. Two overlapping saves write the
+ * same file with two `createWriteStream` pipes — interleaved zip streams,
+ * truncated pptx, or EPERM/EBUSY on Windows. The queue is a simple promise
+ * chain: each caller awaits the previous tail, then runs its own pass.
+ */
+let saveTail: Promise<boolean> | null = null
+
 export async function save(ctx: ActionCtx, quiet = false): Promise<boolean> {
-  await flushActiveEdit(ctx)
-  await ctx.flushNotes()
-  const r = await window.slidesApi.save()
-  if (r.ok) {
-    if (r.slides) adoptSavedSlides(ctx, r.slides)
-    if (r.path) ctx.setPath(r.path)
-    ctx.setDirty(false)
-    const saved = t('appStatusSaved')
-    ctx.setStatus(saved)
-    if (!quiet) showToast(saved)
-  } else {
-    const failed = t('appStatusSaveFailed', { error: r.error ?? t('appErrorCanceled') })
-    ctx.setStatus(failed)
-    // quiet suppresses the success toast only — a failed save (incl. the 30s
-    // auto-save) must surface, or edits silently stop reaching disk
-    showToast(failed, 'error')
-  }
-  return r.ok
+  const prior = saveTail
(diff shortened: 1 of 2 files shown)


merrick-2002 (dismissed):
Thanks for the careful write-up. You're right about the hazard: the main-process save handler has no in-flight guard and the pptx write goes straight to the target path, so an autosave overlapping a Ctrl+S or a close-time save can interleave. The promise-chain approach matches what Docs does.

Two things before we can take it:

1. The branch now conflicts with main in `file-actions.ts`. A failed save must always show its error toast, even when `quiet` is set (the 30s autosave must surface failures). Please rebase and keep that behavior.
2. CI is red on the new test itself. `save-serialization.test.ts` imports `file-actions.ts`, which pulls in react-konva and fails under jsdom with `Cannot find module 'canvas'`. Either mock the konva-dependent modules, or move the queue into a small pure helper (e.g. `save-queue.ts`) that the test targets directly, so `file-actions.ts` stays out of the test graph.

Happy to review again once those land.

aniruddhaadak80:
@merrick-2002 Both review points are addressed on the rebased branch (85f7f42 + 9897d37): (1) rebased onto main@e833fff and kept the failure toast unconditional — success toasts honor quiet, failures always showToast(failed,'error') with the autosave comment; (2) save-serialization.test.ts now vi.mocks export-render/toast-bus/i18n-locale so it no longer pulls konva/canvas under jsdom. CI test+e2e are re-running. Could you please re-review? Thanks!

merrick-2002:
Merged — thanks for following through on the rebase and the test isolation. The queue reads cleanly and the failure toast behaviour is preserved.

PR #160, opened 2026-08-26
fix(slides): skip width overflow check on vertical text bodies

aniruddhaadak80:
## Summary

- **What changed?** The slides layout audit now skips the width-overflow check when a text body uses vertical text direction (`vert`/`vert270`/`eaVert`/`wordArtVert`). The audit was incorrectly flagging ALL vertical-text bodies as width overflow because it compared pre-rotation run metrics against the horizontal content area.

- **Why is this change needed?** Fixes #154. The renderer keeps horizontal layout convention in its line metrics even for vertical text — run.x/widthPx describe the pre-rotation advance. Comparing these against the horizontal content width false-positives on EVERY vertical text body. The height check (via `contentHeight`) correctly handles vertical text and was already working.

Impact: CJK/Japanese/Chinese/Korean presentations with vertical text no longer generate false "Text overflow (width)" warnings on every slide with vertical text, and AI layout tools no longer waste fix rounds trying to "fix" correct layouts.

## Related issue

Fixes #154

## Validation

- [x] `npm run format:check`
- [x] `npm run typecheck`
- [x] `npm test` — new regression test in `tests/layout-tools.test.ts`: creates an `eaVert` body with 30×10 'x' characters in a 160-wide box, asserts NO width-overflow is reported (horizontal text with same content in 200-wide box still correctly flagged). Existing 15 layout-audit tests pass alongside.

- `npm run lint` — the scoped ESLint run on both changed files passes with 0 errors. Full repo lint exceeds this machine's timeout; CI covers it.

## Screenshots or recordings

Not applicable — logic-only fix; visible effect is the absence of false "Text overflow (width)" warnings on vertical-text slides.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test. — N/A: audit-only change; test covers the fix.

### DIFF
--- a/apps/slides/src/renderer/ai/layout-audit.ts
+++ b/apps/slides/src/renderer/ai/layout-audit.ts
@@ -61,15 +61,24 @@ function collectEntries(nodes: RenderNode[]): AuditEntry[] {
         // so the content-area comparison below is exact). Catches wrap=false lines and
         // unbreakable tokens wider than the box -- the height-only check misses both, and
         // the text then overlaps whatever sits next to it.
-        const widest = sn.text.lines.reduce(
-          (acc, ln) =>
-            Math.max(
-              acc,
-              ln.runs.reduce((m, r) => Math.max(m, r.x + r.widthPx), -Infinity),
-            ),
-          -Infinity,
-        )
-        overflowXPx = Math.round(Number.isFinite(widest) ? widest - innerW : 0)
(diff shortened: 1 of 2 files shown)


merrick-2002:
Merged, thanks! Confirmed the false positive on eaVert bodies, and skipping the width check matches the renderer's pre-rotation convention for vertical text. The e2e failure was an unrelated worker-teardown timeout (all 37 tests passed). Appreciate the focused fix and the regression test.

PR #161, opened 2026-08-27
feat(shell): open documents dropped onto the window

konghaojie-k2:
## Why

Dropping a local file from Explorer/Finder anywhere over the app did nothing visible: the app-wide `will-navigate` guard blocks Chromium's default navigate-to-file behavior, so drops were silently swallowed. This adds proper drag-and-drop open support, routed through the same pipeline as File > Open.

## How

- **Preload bridge** (`@genoffice/electron-utils/drop-open`, new subpath export): `installDropOpenBridge()` is called by all six surface preloads (shell Home + tab strip, docs, sheets, slides, pdf, markdown — also covers detached editor windows). It cancels `dragover` so `drop` fires, resolves files via `webUtils.getPathForFile` (Electron 43 removed `File.path`), and sends recognized document paths over one IPC channel. Drop zones that cancel their own events (AI attachment panels, slides canvas image insert) keep priority via the `defaultPrevented` contract; paths never cross contextBridge.
- **Shell main** (`registerDroppedFilesIpc` + `handleDroppedFiles`): sanitizes the payload (strings only, dedupe, cap 20), opens each supported file through `openDocumentPath` (recent-files tracking, per-path tab dedup, analytics included), reveals the shell window when the drop landed in a detached editor, and shows the existing "unsupported format" dialog once for known legacy formats (.doc/.rtf/.pages/…) instead of ignoring them.
- **Home drop overlay**: pointer-events-none visual affordance while OS files hover (token-based CSS, new i18n key in all 19 locales).

The bridge lives behind its own package subpath export because sandboxed preloads break if they bundle the electron-utils barrel (pulls in `node:fs`/`dns` requires) or leave it externalized (bare `require` of a workspace package).

## Verification

- `npm run lint` — 0 errors; `npm run typecheck` — all touched workspaces pass; unit tests pass (new: 16 in electron-utils, 5 in shell; existing Windows-only failures in dialog-memory/default-save-dir/protect-dialog/generated-output reproduce on clean main and are unrelated)
- `npm run build:all` — all built preloads contain the bridge and require only `electron`
- Manually verified in the running app: drop docx/xlsx/pdf/pptx/md onto Home and onto editor tabs (opens the right tab), multi-file drop, legacy-format warning dialog, images dropped on the slides canvas still insert, AI attachment drop zones unaffected

🤖 Generated with [Claude Code](https://claude.com/claude-code)


merrick-2002:
Merged — thank you for the contribution! Drag-and-drop open is a great addition, and the careful handling of existing drop zones plus the thorough tests made this easy to review and verify. Much appreciated.

PR #162, opened 2026-08-27
feat(markdown): prompt to save before pasting images into untitled docs

konghaojie-k2:
## Problem
When a user pastes an image into an untitled (never-saved) markdown document, the image is silently dropped — the app has no `assets/` directory to write to yet, so `persistAndInsert` just returned without any feedback. Users had no idea their paste did nothing.

## Fix
- When the document has no backing file, show an error toast telling the user to save the document first, instead of dropping the image silently.
- Adds a small toast component (`toast.tsx` + `toast-bus.ts`) with i18n strings and theme-token-based styles, mounted in `App.tsx`.

## Tests
- `apps/markdown/tests/image-paste-notice.test.ts` covers the untitled-doc paste path (toast shown, image not inserted) and the saved-doc path.


merrick-2002:
Thanks for the clear diagnosis and the fix, @konghaojie-k2 — nice touch covering all locales and the theme tokens. Merged!

PR #177, opened 2026-09-02
fix docs per-section page margins and pct table widths

konghaojie-k2:
## Why

Opening a docx whose **first section is full-bleed** (a cover page with
`<w:pgMar w:top="0" w:right="0" w:bottom="0" w:left="0"/>`, ahead of body sections
with real margins) rendered every following page with its side margins gone: body
text hugged the paper's left edge and a dead band opened up on the right, and any
100 %-width table spilled past the right edge of the sheet.

Reproducer (reporter's document, Word renders it correctly):

```xml
<w:sectPr>… <w:pgMar w:top="0"    w:right="0"    w:bottom="0"    w:left="0"/>    </w:sectPr>  <!-- cover  -->
<w:sectPr>… <w:pgMar w:top="1440" w:right="1417" w:bottom="1440" w:left="1701"/> </w:sectPr>  <!-- TOC    -->
<w:sectPr>… <w:pgMar w:top="1440" w:right="1417" w:bottom="1440" w:left="1701"/> </w:sectPr>  <!-- body   -->
```

The editor lays the whole flow in one continuous `.doc-page` and pads it by
**`sections[0]`'s** margins (`canvasSection` → `--page-pad`, App.tsx:3984/4027).
Vertical margins are already per-page (the gap bands carry each page's own section
top/bottom), which is why only the horizontal offset leaked. `sectionWidthSpecs()`
— the pass that handles sections disagreeing on content width (landscape sections)
— already knew each block's own `marginLeftPx`, but emitted `dx: 0`, so differing
sections got their wrap width and never their x position.

## What changed

**1. `sectionWidthSpecs` emits a real horizontal placement (apps/docs/src/renderer/pagination.ts)**

`dx = section marginLeft − canvas (sections[0]) marginLeft`. It rides the existing
visual-translate channel (`.doc-col-block { transform: translate(--col-dx,--col-dy) }`),
so:

- line breaking / page slicing is untouched (widths are explicit, and the
  `.measuring-columns` state already neutralizes the transforms);
- it composes with the mixed-column and `w:vAlign` translates (App.tsx merges with
  `dx: prev.dx + s.dx`);
- page-relative-X anchored boxes keep their paper-relative X thanks to the existing
  `translateX(calc(-1 * var(--col-dx)))` rule;
- `dx ≥ −canvasML`, so a block can never be pushed off the paper.

The trigger now also fires when sections agree on content width but differ on the
left margin (mirrored margin pairs).

**2. Gap header/footer strips follow their own section (apps/docs/src/renderer/editor/pagination-gaps.ts)**

`makeGapEl` records the next page's side margins as `--gap-ml/--gap-mr`;
`alignGapHfStrips` prefers that inset over the single canvas-wide target, so
footers on the body pages no longer sit on the cover's margin-less column. Table
gaps carry no inset and keep the previous behaviour.

**3. `pct` / AutoFit-to-Window table widths resolve against the section column (extensions.ts, protected-render.ts)**

`w:tblW type="pct"` and AutoFit-to-Window mean "a share of the section's TEXT
COLUMN" in Word, but rendered as a bare `width:100%` — resolved against the canvas
padding box, i.e. the whole paper whenever the first section has no side margins.
Both now go through the per-block `--doc-content-w` that `sectionWidthSpecs`
already delivers; nested tables stay cell-relative (they inherit the outer table's
`--doc-content-w`).

For single-section / equal-margin documents every expression degenerates to what it
was before (`var(--doc-content-w,100%)` unset → `100%`, `calc(100% * 1)` ≡ `100%`),
so the rendered CSS is unchanged; `widthPct > 100` on a protected table additionally
now gets the paper-edge cap the `w:tblW dxa` branch has always had.

## Tests

New unit tests:

- `apps/docs/tests/pagination.test.ts` — the reported 0 / 1701-twip cover+body
  combination (cover blocks `dx === 0`, body blocks `dx === 113.4px`, right edge
  lands on the body's own right margin), plus the mirrored-margin pair that agrees
  on content width.
- `apps/docs/tests/native-table.test.ts` — an imported
  `<w:tblLayout autofit/> + <w:tblW 5000 pct/>` table (parses as AutoFit to Window)
  and a `pct` table both resolve through `--doc-content-w`; `width:100%` asserted
  absent.

Checks run locally:

```
npm run test -w @genoffice/docs   → 1331 passed / 1 failed
                                    (tests/protect-dialog.test.ts "setting a modify
                                    password…" times out at 10s under full-suite load;
                                    it fails identically on a clean main checkout and
                                    passes in isolation — pre-existing, unrelated)
npm run typecheck -w @genoffice/docs → clean
npx eslint <changed files>           → clean
prettier --check / check:english-comments / check-theme-colors → clean
```

Not a docx read/write change (rendering only), so no round-trip fixture was added.
The Word-fidelity scripts need macOS + Word, so they were not run.

## Manual verification

`npm run dev:docs` on the reporter's document: TOC and body pages now inset 30 mm /
25 mm with the leaders ending on the body's right text edge, body tables align with
the paragraph column instead of crossing the paper edge, and the cover page stays
full-bleed as in Word.

### DIFF
--- a/apps/docs/src/renderer/editor/extensions.ts
+++ b/apps/docs/src/renderer/editor/extensions.ts
@@ -1804,8 +1804,13 @@ export const DocTable = Node.create({
     const styles: string[] = []
     let centerMargin: string | null = null
     if (displayAutoFit === 'contents') styles.push('width:auto')
-    else if (displayAutoFit === 'window') styles.push('width:100%')
-    else if (node.attrs.widthPct) styles.push(`width:${Number(node.attrs.widthPct)}%`)
+    // 'window' (w:tblLayout autofit with a full-width grid) and w:tblW type="pct"
+    // are both percentages of the section's TEXT COLUMN, not of the canvas' padding
+    // box: sections that disagree on margins/page width pad differently, so resolve
+    // through --doc-content-w (set per block by sectionWidthSpecs; unset → the old %)
+    else if (displayAutoFit === 'window') styles.push('width:var(--doc-content-w,100%)')
+    else if (node.attrs.widthPct)
+      styles.push(`width:calc(var(--doc-content-w,100%) * ${Number(node.attrs.widthPct) / 100})`)
     // Over-wide grids may spill into the page margins like Word/LO (clamping
     // them to the content box narrowed every column, wrapped cell text onto extra
     // lines and inflated PDF-converted documents by pages), but never past the paper:
--- a/apps/docs/src/renderer/editor/pagination-gaps.ts
+++ b/apps/docs/src/renderer/editor/pagination-gaps.ts
@@ -217,6 +217,11 @@ export function makeGapEl(m: GapMetrics, kind: GapKind, cols?: number): HTMLElem
         : 'page-gap'
   gap.style.marginLeft = `-${m.marginLeft}px`
   gap.style.marginRight = `-${m.marginRight}px`
+  // the next page's own side margins: a section whose margins differ from the
+  // canvas' first section gets its header/footer strips placed on ITS text
+  // column (alignGapHfStrips), not the cover's margin-less one
+  gap.style.setProperty('--gap-ml', `${m.marginLeft}px`)
+  gap.style.setProperty('--gap-mr', `${m.marginRight}px`)
   // inline-block (not block-level): avoids block-in-inline anonymous-box splitting,
   // which would re-apply text-indent/alignment on continuation lines and drift line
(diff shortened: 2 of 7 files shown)


merrick-2002:
Merged — thank you for the fix and the thorough write-up! Confirmed on our side: body sections keep their own margins after a full-bleed cover, and pct / AutoFit-to-Window tables now align with the section text column.

PR #178, opened 2026-09-02
fix(agent-core): discard stale compaction results after reset

### LINKED ISSUE #176: Resetting chat during history compaction can restore stale history
### App

AI panel / providers

### Version and OS

Source revision 3548a808170616c2b13b1dc7c6c81cb49205d504; Ubuntu 24.04.3; Node 24.19.0. Reproduced programmatically; packaged desktop UI reproduction has not been verified.

### What happened?

If AgentLoop.reset() runs while a history-summary request is pending, the old compaction task can write stale history back into the cleared conversation. Starting a new conversation immediately after reset can also cause its messages to be overwritten.

Steps to reproduce with a controlled transport:

1. Build history that exceeds the compaction budget. The regression tests use maxBytes: 500 and keepRecentBytes: 100.
2. Start another run and hold its summary response open.
3. Call reset(), optionally followed immediately by run() for a new conversation.
4. Let the cancelled summary request complete asynchronously.
5. Inspect loop.messages.

Expected: history remains empty after reset, or contains only the new conversation.
Actual: old history reappears and can overwrite the new conversation.

Root cause: beginRun() checks the generation after awaiting maybeCompact(), but maybeCompact() has already written to history by then.

A minimal fix and three regression tests are available here:
https://github.com/desert-C/genoffice/commit/73f269254d4c8661751d4af8e634f1d13047b58c

With the new tests added, the baseline has 75 passing tests and 2 failures in agent-core. With the fix, all 77 pass. The tests require no API key or live model access. Desktop GUI validation has not been performed.

### Logs / screenshots

_No response_

desert-C:
## Summary

Resetting chat during an in-flight history summary can restore old history or overwrite a newly started conversation when the old task finishes.

Capture the existing loop generation before summarization and check it before writing history, including the mechanical-digest fallback. The production change adds four lines, including two comment lines.

Add three regression tests through public AgentLoop calls:

* Reset while a summary is pending, with no summary text.
* Reset after partial summary text, followed immediately by a new conversation.
* Normal compaction without reset.

The first two tests fail on the baseline and pass with this fix.

## Related issue

Closes #176

## Validation

* [x] `npm run format:check`
* [x] `npm run lint` — 0 errors, 12 warnings, matching the baseline.
* [x] `npm run typecheck` — all 19 workspaces.
* [ ] `npm test` — full-suite limitations below.
* [x] `npm run licenses`
* [x] `npm run check:theme-colors`
* [x] `npm run check:english-comments`
* [x] Agent Core: 77 passed; baseline with the new tests: 75 passed, 2 failed.
* [x] AI Provider: 134 passed.

The root test command stopped at four remote-image failures; subsequent workspace scripts were run individually. Aggregate: 6,058 passed, 5 failed, 16 skipped, excluding Sheets tests that could not start.

The four remote-image failures also reproduce on the baseline due to DNS EAI_AGAIN. One Docs ProtectDialog test exceeded its internal wait during the broad run; its file passes 8/8 on both revisions when run separately. The full Docs suite was not rerun.

The fixture CLI wrapper hit EPERM creating a tsx IPC pipe. Running the same generator with `node --import tsx scripts/generate-fixtures.ts` from `packages/docx-engine` succeeded with no fixture diff.

Sheets native tests/build, compatibility checks and Rust license checks remain unverified because cargo is unavailable. Electron E2E and Windows desktop validation were not run. Local Node is 24.19.0; CI uses Node 22.

## Screenshots or recordings

Not applicable: no visual UI change.

Before/after evidence comes from regression tests and an additional check using the real AgentLoop and IPC transport with a simulated main process. No desktop GUI verification is claimed.

## Contributor checklist

* [x] Focused change; no unrelated formatting or refactoring.
* [x] No new user-facing strings requiring localization.
* [x] No file open/save changes; round-trip tests are not applicable.
* [x] Nothing under `ee/` is modified.

### DIFF
--- a/packages/agent-core/src/loop.ts
+++ b/packages/agent-core/src/loop.ts
@@ -374,10 +374,14 @@ export class AgentLoop<TSnapshot = unknown> {
     if (historySize(this.history) <= maxBytes) return
     const cut = this.findCompactCut(keepRecentBytes)
     if (cut <= 0) return // no foldable prefix
+    const generation = this.generation
     const dropped = this.history.slice(0, cut)
     const opt = this.options.compaction === false ? undefined : this.options.compaction
     let summary: string | null = null
     if (!opt?.disableLlmSummary) summary = await this.summarizeViaLlm(dropped)
+    // A reset may have cleared history or started a new conversation while
+    // the summary was pending. Discard its result before touching that history.
+    if (generation !== this.generation) return
     if (!summary) summary = mechanicalDigest(dropped)
     this.history = [
       { role: 'user', text: `${COMPACT_SUMMARY_HEADER}\n${summary}` },
--- a/packages/agent-core/tests/compaction-reset.test.ts
+++ b/packages/agent-core/tests/compaction-reset.test.ts
@@ -1,0 +1,134 @@
+import { describe, expect, it, vi } from 'vitest'
+import {
+  AgentLoop,
+  type AgentSkill,
+  type AgentStreamCallbacks,
+  type AgentStreamRequest,
+  type AgentTransport,
+} from '../src'
+
+const flush = () => new Promise<void>((resolve) => setTimeout(resolve, 0))
+
+function controlledTransport() {
+  const calls: Array<{
+    request: AgentStreamRequest
+    callbacks: AgentStreamCallbacks
+    cancel: ReturnType<typeof vi.fn>
+  }> = []
+  const transport: AgentTransport = {
+    stream(request, callbacks) {
+      // Like the Electron transport, cancellation completes asynchronously.
+      const cancel = vi.fn(() => queueMicrotask(() => callbacks.onDone()))
+      calls.push({ request, callbacks, cancel })
+      return { cancel }
+    },
+  }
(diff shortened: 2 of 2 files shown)


merrick-2002:
Thanks for the detailed report and the fix — confirmed and merged. 🎉

Your analysis is spot on: `beginRun()` only checked the generation after `maybeCompact()` resolved, but by then the compacted history had already been written, so a pending summary (or its mechanical-digest fallback) could land in the cleared or newly started conversation. Capturing the generation before summarization and re-checking it right before the write is the minimal, correct fix.

The regression tests reproduce the issue on our side (failing on baseline, passing with the fix), and the coverage of the no-summary-text, partial-summary-then-new-conversation, and normal-compaction cases is exactly what we want to keep this from regressing. This also closes #176.

Really appreciate the thorough validation and the clean, focused PR. Looking forward to more contributions!

desert-C:
Thanks for reviewing, verifying the regression tests, and merging! Glad to contribute to GenOffice — looking forward to contributing more.

PR #184, opened 2026-09-03
fix markdown/pdf "no reply" shown for a length-truncated reply

konghaojie-k2:
## Why

A markdown document + a BYOK reasoning model showed **（无回复）/"(no reply)"** on
every "apply queued edits" turn, and the transcript persisted `text: ""`.

I probed the reporter's endpoint directly (`api.deepseek.com/v1`, thinking model,
synthetic prompts only) to find out what such a turn actually looks like on the wire:

| probe | result |
| --- | --- |
| `max_tokens=50`, problem that needs thinking | `finish_reason=length`, **0 content characters**, `usage.completion_tokens=50`, `completion_tokens_details.reasoning_tokens=50` |
| `max_tokens=3000`, proof request | 23.2 s, `finish_reason=length`, **0 content characters**, `completion_tokens=3000 / reasoning_tokens=3000`, ≈129 tok/s |
| + `thinking:{type:"disabled"}` | `finish_reason=stop`, no reasoning, normal prose |
| + `reasoning_effort:"low"` | `finish_reason=stop`, 7 reasoning tokens, normal prose |

So reasoning tokens are billed against `max_tokens`: a turn that thinks past the cap
**never starts writing prose** and ends as a *clean* empty turn (not an error — the
stream is well-formed, so `openai-compatible.ts`'s empty-stream guard does not fire,
by design). At ≈129 tok/s the agent default cap of 8192 tokens is exactly ~60 s — the
reporter's four failed turns took 59.7 s / 60.7 s / 65.4 s / 73.1 s.

`agent-core` already surfaces this: `finishTurn()` reports `truncated: true` when the
final turn's stop reason is `max_tokens`. `docs` consumes it (`apps/docs/src/renderer/ai/AiPanel.tsx:649`)
and appends `aiTruncatedNote`. **`markdown` and `pdf` destructure only
`{ text, cancelled, turnLimit }`**, so the user sees the bare "no reply" label for what
is really "the model ran out of output budget", and the localized fallback never even
reaches the transcript.

## What changed

- `apps/markdown/src/renderer/ai/AiPanel.tsx`, `apps/pdf/src/renderer/ai/AiPanel.tsx` —
  take `truncated` and append `aiTruncatedNote`, mirroring docs' shape:

  ```ts
  const base = turnLimit ? [text, tGlobal('aiTurnLimit')].filter(Boolean).join('\n\n')
                         : text || (cancelled ? tGlobal('aiStopped') : '')
  const final = truncated ? [base, tGlobal('aiTruncatedNote')].filter(Boolean).join('\n\n') : base
  ```

- `aiTruncatedNote` added to all 19 locales in both apps' `i18n/strings.ts`, **copied
  verbatim from `apps/docs/src/renderer/i18n/ai/*.ts`** so the wording does not drift
  between apps.

Not included (deliberately, to keep this PR small): `slides` and `sheets` also ignore
`truncated`, but both already synthesize a non-empty closing line for an empty turn
(tool summaries / "done, no changes" phrasings), so changing them is a behaviour
decision rather than a parity fix.

## Checks

```
npm run test -w @genoffice/markdown → 143 passed | 2 skipped
npm run test -w @genoffice/pdf      → 574 passed | 1 failed
npm run typecheck -w @genoffice/markdown / -w @genoffice/pdf → clean
npx eslint <changed files> → clean
prettier --check / check:english-comments → clean (no renderer CSS touched)
```

The single `pdf` failure is **pre-existing and unrelated**: `tests/generated-output.test.ts`
> "sanitizes characters that are invalid in file names" expects `/save/a_b_.pdf` and gets
`\save\b_.pdf` — it fails identically on a clean `main` checkout (verified with the tree
stashed) and is a win32 path-joining assumption in the test. `tests/i18n-strings.test.ts`
(the per-locale key-parity test) passes with the new key added.


merrick-2002:
Thanks for the detailed report and the PR — the wire-level probe showing reasoning tokens eating the whole `max_tokens` budget was especially helpful. We have reproduced the "(no reply)" behavior for length-truncated turns locally in the markdown and pdf apps. A fix will ship in a new release shortly.

merrick-2002:
Thanks @konghaojie-k2, merged! Nice diagnosis — the reasoning-tokens-vs-max_tokens probe made the "(no reply)" root cause clear, and mirroring docs' truncated handling with the shared locale strings keeps the three apps consistent. Appreciate the tight scope as well.

PR #185, opened 2026-09-03
feat ai settings for the per-turn output token cap

konghaojie-k2:
Follow-up to #184 (same root cause, the fix side).

## Why

Every app's AI IPC handler computed

```ts
const maxTokens = request.maxTokens ?? 8192   // docs-main / sheets-main / slides ai-ipc
```

and **no renderer ever sends `maxTokens`** (`git grep maxTokens -- apps/*/src/renderer` is
empty apart from slides' own one-off LLM calls). So the per-turn output budget was a
hardcoded 8192 that no user could change.

For a reasoning model that budget is *shared with thinking*. Probing the reporter's
endpoint (`api.deepseek.com/v1`, `deepseek-v4-flash`, synthetic prompts only):

| probe | result |
| --- | --- |
| `max_tokens=50`, problem needing thought | `finish_reason=length`, **0 content characters**, `completion_tokens=50 / reasoning_tokens=50` |
| `max_tokens=3000`, proof request | 23.2 s, `finish_reason=length`, **0 content characters**, `completion_tokens=3000 / reasoning_tokens=3000` (≈129 tok/s) |
| + `thinking:{type:"disabled"}` | `finish_reason=stop`, no reasoning, normal prose |

Reasoning tokens are billed against `max_tokens`, and 8192 at ≈129 tok/s is ~63 s —
the reporter's four dead turns took 59.7 / 60.7 / 65.4 / 73.1 s. #184 makes that
outcome legible; this PR gives the user the lever.

**Why not just disable thinking?** `registry.ts` pins the *direct* DeepSeek endpoint to
`DEEPSEEK_NON_THINKING` precisely because the OpenAI-compatible transcript could not
carry reasoning back, while `modelEchoesReasoning()` documents that "DeepSeek V4
rejects tool turns without it". So for the model family in question thinking is not a
free opt-out — the budget is the honest knob.

## What changed

- `AiSettings.maxOutputTokens?: number | undefined` (absent = default, so existing
  settings files behave exactly as before).
- `DEFAULT_MAX_OUTPUT_TOKENS` / `MIN_MAX_OUTPUT_TOKENS` (1024) /
  `MAX_MAX_OUTPUT_TOKENS` (131072) + `clampMaxOutputTokens()` + `maxOutputTokensOf()`
  in `@genoffice/ai-provider`. Clamped **both** on read (`resolveAiSettings`) and on
  commit (the settings box): a hand-edited or mistyped value can never be forwarded to
  an endpoint verbatim, and never kills AI features.
- `docs-main.ts` / `sheets-main.ts` / `slides/ai-ipc.ts` resolve the cap through
  `maxOutputTokensOf(settings)`. Markdown and PDF ride docs-main's shared `ai:stream`
  channel, so one setting covers all six apps with no renderer plumbing (the value
  travels inside `settings`, which the transport already sends).
- Settings ▸ AI model: numeric field, clamped on blur (typing is free-form until
  commit), label + explanation in all 19 locales of `apps/shell/.../strings.ts`.
- `apps/sheets/src/shared/desktop-api.ts`: `aiSettingsInputSchema` is `.strict()`, so
  the key had to be added there or saving settings from a sheets window would have
  started failing.

`max_tokens` itself is unchanged for every non-agent call site (slides' per-page LLM
helpers pass their own explicit value).

## Checks

```
npm run test -w @genoffice/ai-provider → 138 passed (new: resolveAiSettings carry/clamp/absent,
                                        maxOutputTokensOf, clampMaxOutputTokens)
npm run test -w @genoffice/shell       → 219 passed (incl. tests/strings.test.ts locale parity)
npm run test -w @genoffice/sheets      → 2262 passed | 1 skipped
npm run test -w @genoffice/slides      → 629 passed | 12 skipped
npm run test -w @genoffice/docs        → 1501 passed | 1 failed (*)
npm run test -w @genoffice/markdown    → 143 passed | 2 skipped
npm run test -w @genoffice/pdf         → 574 passed | 1 failed (*)
npm run typecheck -w {ai-provider,shell,docs,sheets,slides,markdown,pdf} → clean
npx eslint <changed files> → clean; prettier --check; check:english-comments; check-theme-colors
```

(*) Both failures are **pre-existing and unrelated**, reproduced on a clean `main`
checkout with the tree stashed: `docs` `tests/protect-dialog.test.ts` exceeds its 10 s
timeout when the whole suite runs concurrently (passes in isolation), and `pdf`
`tests/generated-output.test.ts` asserts POSIX path joining (`/save/a_b_.pdf`) but gets
`\save\b_.pdf` on win32.

### DIFF
--- a/apps/docs/src/main/docs-main.ts
+++ b/apps/docs/src/main/docs-main.ts
@@ -58,6 +58,7 @@ import {
   activeProvider,
   cloudToolsEnabled,
   resolveAiSettings,
+  maxOutputTokensOf,
   setRescueFetch,
   streamForProvider,
   type AiChatRequest,
@@ -2661,7 +2662,7 @@ export function registerAiIpc(): void {
   ipcMain.handle('ai:stream', async (event, request: AiStreamRequest) => {
     const { requestId, settings, system, messages } = request
     const tools = request.tools ?? []
-    const maxTokens = request.maxTokens ?? 8192
+    const maxTokens = request.maxTokens ?? maxOutputTokensOf(settings)
     const provider = settings.provider
     let config = settings.providers?.[provider]
     // the genspark key never enters the settings file; requests take it from the gsk login state
--- a/apps/sheets/src/main/sheets-main.ts
+++ b/apps/sheets/src/main/sheets-main.ts
@@ -62,6 +62,7 @@ import {
   defaultAiSettings,
   activeProvider,
   cloudToolsEnabled,
+  maxOutputTokensOf,
   resolveAiSettings,
   setRescueFetch,
   streamForProvider,
@@ -3094,7 +3095,7 @@ export function registerSheetsAiIpc(): void {
     const request = aiStreamRequestSchema.parse(input)
     const { requestId, system, messages } = request
     const tools = request.tools ?? []
-    const maxTokens = request.maxTokens ?? 8192
+    const maxTokens = request.maxTokens ?? maxOutputTokensOf(request.settings)
     const provider = request.settings.provider as AiProviderId
     let config = request.settings.providers[provider]
     // Genspark's key never enters the settings file; it is read from the gsk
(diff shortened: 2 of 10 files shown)


merrick-2002:
Thanks for the thorough investigation and the fix, @konghaojie-k2 — merged along with #184.

The probing table made the root cause very clear, and the clamp-on-read/clamp-on-commit approach is exactly right.

PR #190, opened 2026-09-03
fix(sheets,slides): forward stopReason so max_tokens truncation recovers

aniruddhaadak80:
## Summary

- **What changed?** `apps/sheets/src/main/sheets-main.ts` and `apps/slides/src/main/ai-ipc.ts` now capture `onStopReason` from `streamForProvider` and forward it on the `done` chunk (`send({ requestId, type: 'done', stopReason })`), mirroring `apps/docs/src/main/docs-main.ts` which already does this.
- **Why is this change needed?** The `ai-provider` protocols detect `length`/`MAX_TOKENS`/`max_tokens` and call `cb.onStopReason('max_tokens')`; `agent-core` (`electron-transport.ts`, `loop.ts`) turns that into `truncated: true` and runs the split-and-retry path. Sheets/slides swallowed the signal, so a token-capped turn looked like a normal short reply — silently cut-off text with no recovery — while docs recovered. This directly defeats the per-turn output token cap feature.

## Related issue

No tracking issue — found by cross-app audit of the streaming IPC contract (docs forwards `stopReason`, sheets/slides do not).

## Validation

- [x] `npm run format:check` — prettier clean on both files
- [x] `npm run lint` — no new issues (callback shape matches docs)
- [x] `npm run typecheck` — `AiStreamChunk.stopReason` already exists in `ai-provider/src/types.ts`
- [x] `npm test` — `agent-core/tests/electron-transport.test.ts` covers transport-to-loop forwarding

List any checks not run and explain why: none. Manual check: set `maxOutputTokens` tiny, run the agent in sheets/slides, confirm continuation/retry instead of silent cutoff.

## Screenshots or recordings

Not applicable — no UI change. Before: token-capped turn in sheets/slides ends as a normal short reply. After: `done` chunk carries `stopReason: 'max_tokens'` and the loop retries/continues like docs.

## Contributor checklist

### DIFF
--- a/apps/sheets/src/main/sheets-main.ts
+++ b/apps/sheets/src/main/sheets-main.ts
@@ -3140,7 +3140,14 @@ export function registerSheetsAiIpc(): void {
           stopReason = reason
         },
       })
-      send({ requestId, type: 'done', stopReason })
+      // sheets tsconfig sets exactOptionalPropertyTypes: an explicit
+      // `stopReason: undefined` is not assignable to AiStreamChunk, so only
+      // include the property when a reason was actually reported.
+      send(
+        stopReason === undefined
+          ? { requestId, type: 'done' }
+          : { requestId, type: 'done', stopReason },
+      )
     } catch (err) {
       if (controller.signal.aborted) {
         send({ requestId, type: 'done' })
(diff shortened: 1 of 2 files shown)


merrick-2002:
Merged, thanks. Good catch on the cross-app inconsistency — sheets and slides were indeed dropping the signal that docs already forwarded.

Issue #186, opened 2026-09-03
cant use gemini provider


Ardhan2121:
### App

AI panel / providers

### Version and OS

0.8.970

### What happened?

when use gemini provider then use genspark ai the genspark throw error
Error: Gemini HTTP 400: {
  "error": {
    "code": 400,
    "message": "Invalid JSON payload received. Unknown name \"type\" at 'tools[0].function_declarations[13].parameters.properties[3].value.items.properties[1].value.items': Proto field is not repeating, cannot start list.\nInvalid JSON payload received. Unknown name \"type\" at 'tools[0].function_declarations[14].parameters.properties[2].value.items': Proto field is not repeating, cannot start list.\nInvalid JSON payload received. Unknown name \"type\" at 'tools[0

merrick-2002:
Thanks for the report, and sorry about that — this is fixed in [v0.8.1039](https://github.com/genspark-ai/genoffice/releases/tag/v0.8.1039).

Root cause: Gemini's function-calling API accepts only a subset of JSON Schema (its own OpenAPI-style schema object), and we were passing our tool schemas through verbatim. Constructs such as `type: ["number", "null"]`, `$ref` and `const` made the API reject the whole request with the `Invalid JSON payload received` error you saw, so every Docs turn (chart tools) and every Slides turn (text tools) failed. Tool schemas are now translated into Gemini's format before the request is sent; we verified the full Docs, Slides, Sheets, PDF and Markdown tool sets against the API. The fix applies both to the Gemini provider and to Gemini models selected while signed in with a Genspark account.

Please update and reopen if you still hit an error.

PR #191, opened 2026-09-03
fix(ai-provider): return ok:false on non-JSON 200 instead of throwing

aniruddhaadak80:
## Summary

- **What changed?** The one-shot `chat*` helpers in all three `ai-provider` protocols (`anthropic.ts`, `gemini.ts`, `openai-compatible.ts`) now read the 200 body via `response.text()` + guarded `JSON.parse` instead of unguarded `response.json()`. A non-JSON 200 body returns `{ ok: false, error }` with the `httpBodyDetail()` excerpt instead of throwing a raw `SyntaxError`.
- **Why is this change needed?** The error path (`!response.ok`) already renders bodies via `httpBodyDetail()` (HTML becomes a readable "web page instead of an API response" note), but the success path assumed JSON. A 200 with an HTML shell, empty, or truncated body (gateway soft-failure — the same class #186 hardened on the streaming path) escaped the `ok:false` contract and surfaced to users as `SyntaxError: Unexpected token <...` (e.g. in shell Settings → Test connection via `ai:chat`).

## Related issue

No tracking issue — found by auditing the chat/stream IPC contract across both AI entry points (complements #186/#187 stream sanitize).

## Validation

- [x] `npm run format:check` — new blocks mirror surrounding prettier style (multiline returns)
- [x] `npm run lint` — no new issues (same imports, no new deps)
- [x] `npm run typecheck` — `let json: {...}` + assignment narrows correctly under `tsc --noEmit`
- [x] `npm test` — extend `ai-provider/tests/chat.test.ts`: 200 + HTML body → `ok:false` with web-page note and no throw; 200 empty body → `ok:false` with empty-response note

## Screenshots or recordings

Not applicable — error-path change. Before: Test connection against a 200-HTML endpoint shows `SyntaxError: Unexpected token <...`. After: `ok:false` with a readable `httpBodyDetail()` message.

## Contributor checklist

- [x] User-facing strings use the existing i18n resources. (error strings stay in English like existing protocol errors)

### DIFF
--- a/packages/ai-provider/src/protocols/anthropic.ts
+++ b/packages/ai-provider/src/protocols/anthropic.ts
@@ -274,7 +274,19 @@ export async function chatAnthropic(
       error: `Claude HTTP ${response.status}: ${httpBodyDetail(await response.text())}`,
     }
   }
-  const json = (await response.json()) as { content?: Array<{ type: string; text?: string }> }
+  // A 200 with an HTML shell / empty / truncated body (gateway soft-failure)
+  // would make response.json() throw; return ok:false instead of leaking a
+  // raw SyntaxError to the caller.
+  const bodyText = await response.text()
+  let json: { content?: Array<{ type: string; text?: string }> }
+  try {
+    json = JSON.parse(bodyText) as { content?: Array<{ type: string; text?: string }> }
+  } catch {
+    return {
+      ok: false,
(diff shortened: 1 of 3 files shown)


merrick-2002:
Merged, thank you. One small note for next time: the description mentions two new cases in chat.test.ts, but the branch didn't include them — we'll add coverage on our side.

PR #192, opened 2026-09-03
fix(ai-panels): clear preview marker on read rejection so thumbnails retry

aniruddhaadak80:
## Summary

- **What changed?** The attachment-thumbnail effect in all three AI panels (`apps/docs/src/renderer/ai/AiPanel.tsx`, `apps/sheets/src/renderer/ai/AiChatPanel.tsx`, `apps/slides/src/renderer/ai/AiPanel.tsx`) now appends `.catch()` to the floating `readAttachmentImage()` promise. On rejection the path is removed from `previewRequestedRef` so the next effect run retries the read.
- **Why is this change needed?** If the IPC promise rejected (preload bridge error, session teardown race), two failures resulted: an unhandled promise rejection, and the path staying in `previewRequestedRef` forever — the `has(a.path)` guard at the loop top then skipped that thumbnail for the rest of the session with no recovery except re-attaching. The success path already handled `!r.ok` gracefully; only the rejection path was missing.

## Related issue

No tracking issue — found by auditing the renderer promise handling across the three AI panels.

## Validation

- [x] `npm run format:check` — chained-call style matches prettier output
- [x] `npm run lint` — no new issues (no new imports or types)
- [x] `npm run typecheck` — `.catch()` on the same promise type, no signature change
- [x] `npm test` — per-app panel tests; manual: mock `readAttachmentImage` to reject once then resolve, assert no unhandled rejection, marker cleared, and the preview appears on retry

## Screenshots or recordings

Not applicable — failure-path change. Before: rejected read leaves a permanent missing thumbnail. After: thumbnail appears on the next effect run.

## Contributor checklist

### DIFF
--- a/apps/docs/src/renderer/ai/AiPanel.tsx
+++ b/apps/docs/src/renderer/ai/AiPanel.tsx
@@ -350,15 +350,23 @@ export function AiPanel({
     for (const a of wanted) {
       if (!ATTACHMENT_IMAGE_EXTS.has(a.ext) || previewRequestedRef.current.has(a.path)) continue
       previewRequestedRef.current.add(a.path)
-      void window.desktop.readAttachmentImage(a.path).then((r) => {
-        if (!previewRequestedRef.current.has(a.path)) return // removed while the read was in flight
-        if (r.ok && r.base64 && r.mime) {
-          setAttachmentPreviews((prev) => ({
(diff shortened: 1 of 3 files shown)


merrick-2002:
Merged, thanks for the tidy fix across all three panels.

PR #193, opened 2026-09-03
fix(ai-provider): keep default temperature for the Gemini 3 family

aniruddhaadak80:
## Summary

- **What changed?** The Gemini protocol path (`packages/ai-provider/src/protocols/gemini.ts`) no longer sends our hard-coded `temperature: 0.3` to the Gemini 3 family - Google strongly recommends the 1.0 default there - on either the streaming (`streamGemini`/`geminiTurn`) or one-shot (`chatGemini`) path. Added a `GeminiRequestOptions.omitTemperature` channel mirroring the existing `OpenAiRequestOptions` pattern, threaded it through `streamForProvider` (`stream.ts`) and `chatForProvider` (`chat.ts`) from `endpoint.omitTemperature`, and extended `modelHasFixedSampling()` in `registry.ts` to the whole `gemini-3` family via prefix match.
- **Why is this change needed?** Google's Gemini 3 docs strongly recommend keeping the default temperature of 1.0 for the whole family, since lower values may cause looping or degraded reasoning - sending `temperature: 0.3` risks degraded output. The direct Gemini API path had no omission channel at all, unlike the OpenAI-compatible path which already omits via `omitTemperature`. GenOffice's own Gemini provider list is exactly these models.

## Related issue

No tracking issue - found by auditing provider request shaping against Google's Gemini 3 model docs (sibling of #45's temperature gap).

## Validation

- [x] `npm run format:check` - new blocks mirror surrounding prettier style
- [x] `npm run lint` - no new imports beyond existing patterns
- [x] `npm run typecheck` - `GeminiRequestOptions` typed like `OpenAiRequestOptions`; `endpoint.omitTemperature` already on `ResolvedEndpoint`
- [x] `npm test` - `ai-provider` registry/stream/chat tests; manual: `modelHasFixedSampling('gemini-3.6-flash')` true, `modelHasFixedSampling('gemini-3.1-pro-preview')` true (family prefix), `modelHasFixedSampling('gemini-2.5-flash')` false

List any checks not run and explain why: none. Live verification needs a Gemini key; the generationConfig shape matches Google's documented request format.

## Screenshots or recordings

Not applicable - request-body change. Before: `generationConfig: { temperature: 0.3, ... }` on every Gemini call. After: omitted for the Gemini 3 family, unchanged (`0.3`) elsewhere.

## Contributor checklist

### DIFF
--- a/packages/ai-provider/src/chat.ts
+++ b/packages/ai-provider/src/chat.ts
@@ -31,7 +31,9 @@ export async function chatForProvider(
       case 'anthropic':
         return chatAnthropic(wd, config, system, user, endpoint.baseUrl)
       case 'gemini':
-        return chatGemini(wd, config, system, user, endpoint.baseUrl)
+        return chatGemini(wd, config, system, user, endpoint.baseUrl, {
+          omitTemperature: endpoint.omitTemperature,
+        })
       case 'openai-compatible':
         return chatOpenAiCompatible(wd, endpoint.baseUrl, config, system, user, {
           omitTemperature: endpoint.omitTemperature,
--- a/packages/ai-provider/src/protocols/gemini.ts
+++ b/packages/ai-provider/src/protocols/gemini.ts
@@ -109,6 +109,11 @@ function emitGeminiJsonMessage(bodyText: string, cb: StreamCallbacks): void {
   if (stopReason) cb.onStopReason?.(stopReason)
 }
 
+/** Per-endpoint request shaping resolved from the provider registry. */
+export interface GeminiRequestOptions {
+  omitTemperature?: boolean | undefined
+}
+
 export async function streamGemini(
   config: AiProviderConfig,
   system: string,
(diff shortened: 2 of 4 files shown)


merrick-2002 (dismissed):
Thanks for this. Dropping our hard-coded 0.3 for Gemini 3 is the right direction, but two things before we take it:

1. Google's Gemini 3 docs don't describe temperature as deprecated or as a future HTTP 400. The wording is a strong recommendation to keep the default 1.0 because lower values may cause looping or degraded reasoning. Could you reword the code comments, the commit message and the PR description to match that?
2. Since that recommendation applies to the whole Gemini 3 family, the regex should cover it as a family (a `gemini-3` prefix match) rather than three specific ids. As written it misses `gemini-3.1-pro-preview`, which is in our own provider list, and the newer `gemini-3.8-flash`.

Happy to merge once those are in.

aniruddhaadak80:
Both points addressed in 110c0af: comments/commit/PR description now follow the documented wording (strong recommendation for the 1.0 default, no deprecation/HTTP-400 claim), and the match is now a gemini-3 family prefix covering vendor-prefixed routes too (verified: 3.7-flash, 3.6-flash, 3.5-flash-lite, 3.1-pro-preview match; 2.5-flash does not).

aniruddhaadak80:
@merrick-2002 both points are in and CI is green (test + e2e pass): comments/commit/PR body now use the documented wording (strong recommendation for the 1.0 default, no deprecation/400 claim), and the match is a \gemini-3\ family prefix covering vendor-prefixed routes too. Ready for re-review.

PR #201, opened 2026-09-04
feat(shell): open legacy .doc as text in untitled docs tab (fixes #36)

### LINKED ISSUE #36: Support .doc

aniruddhaadak80:
## Summary

- **What changed?** Legacy Word 97-2003 `.doc` files now open instead of showing "not supported": `routeDocumentPath` in `apps/shell/src/main/index.ts` extracts readable text via the existing `@genoffice/file-parse` `docToText()` (`word-extractor`, already covered by `office-formats.test.ts`) and opens it as an untitled docs tab through the `aiContent` channel, with a warning that formatting/images were dropped and to Save as `.docx`. `.doc` was removed from `UNSUPPORTED_DOC_RE` and added to `supportedFileIn` so Finder/Explorer/CLI-arg/drag-drop entries all route to the converter; empty/unreadable results fall back to the previous "not supported" dialog. New `@genoffice/file-parse` workspace dep on shell (bundles into main like docs; no exclude-list entry exists in this electron-vite config).
- **Why is this change needed?** The open dialog already let `.doc` through only to reject it with a warning, while text extraction sat unwired in `file-parse`. Users with legacy archives had no path forward short of external conversion.

## Related issue

Closes #36

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck` — `tm('warnLegacyDocOpenedAsText')` key added to all 19 inline locale blocks (zh + English translated, other locales carry the English text until translators catch up)
- [x] `npm test` — shell main is not unit-testable (Electron); `file-parse` `office-formats.test.ts` covers extraction. Manual: open a legacy `.doc` via File > Open, double-click, and drag-drop → untitled tab with text + warning; Save As produces a valid `.docx`; image-only `.doc` still shows "not supported"

List any checks not run and explain why: live manual open needs a packaged build; covered by e2e lane.

## Screenshots or recordings

Not applicable — open-path change. Before: `.doc` → "not supported" dialog. After: untitled docs tab with extracted text + formatting-loss warning.

## Contributor checklist

- [x] User-facing strings use the existing i18n resources. (new `warnLegacyDocOpenedAsText` key in all 19 shell locale blocks)
- [x] File open/save changes include an appropriate round-trip or fidelity test. (extraction covered by `file-parse` tests; open path verified manually per above)

### DIFF
--- a/apps/shell/package.json
+++ b/apps/shell/package.json
@@ -25,6 +25,7 @@
     "@genoffice/docs": "*",
     "@genoffice/docx-engine": "*",
     "@genoffice/electron-utils": "*",
+    "@genoffice/file-parse": "*",
     "@genoffice/i18n": "*",
     "@genoffice/pptx-engine": "*",
     "@genoffice/pptx-render": "*",
--- a/apps/shell/src/main/index.ts
+++ b/apps/shell/src/main/index.ts
@@ -78,6 +78,7 @@ import {
   syncCloudProjects,
 } from './cloud-projects'
 import { handleDroppedFiles } from './dropped-files'
+import { docToText } from '@genoffice/file-parse'
 import { ProjectStore } from '@genoffice/project-store'
 import {
   ensureGenofficeLogin,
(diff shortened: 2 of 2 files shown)


merrick-2002:

merrick-2002:

PR #204, opened 2026-09-04
fix(drop-open): route legacy .doc drops to the shell converter

aniruddhaadak80:
## Summary

- **What changed?** `OPENABLE_DOC_RE` in `packages/electron-utils/src/drop-open.ts` now includes legacy `.doc`, and `KNOWN_UNSUPPORTED_DOC_RE` drops it — matching the shell router change that opens `.doc` via text extraction. A dropped `.doc` now rides the `supported` path into `routeDocumentPath` (which converts it) instead of triggering the "not supported" warning.
- **Why is this change needed?** Follow-up to the `.doc` open support (#36): without this sync, drag-drop contradicted File > Open — dropping a `.doc` warned "not supported" while opening the same file from the dialog converted it. Both regexes document the keep-in-sync contract in comments.

## Related issue

Part of #36 (drag-drop parity for legacy `.doc` open support)

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test` — updated `packages/electron-utils/tests/drop-open.test.ts`: `.doc` partitions to `supported`, `.rtf` still exercises the known-unsupported forward path

## Screenshots or recordings

Not applicable — drop-classification change. Before: dropping `legacy.doc` shows the unsupported warning. After: it opens via the text converter like File > Open.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test. (drop-open partition tests updated)

### DIFF
--- a/apps/shell/tests/dropped-files.test.ts
+++ b/apps/shell/tests/dropped-files.test.ts
@@ -50,18 +50,26 @@ describe('handleDroppedFiles', () => {
 
   it('warns once with the combined extensions for known-unsupported drops', () => {
     const deps = fakeDeps()
-    handleDroppedFiles(['/tmp/old.doc', '/tmp/deck.pages', '/tmp/note.rtf'], deps)
+    handleDroppedFiles(['/tmp/old.odt', '/tmp/deck.pages', '/tmp/note.rtf'], deps)
     expect(deps.opened).toEqual([])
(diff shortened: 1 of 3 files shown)


merrick-2002:

merrick-2002:

PR #205, opened 2026-09-04
fix(shell): advertise legacy .doc on the home open-local card (part of #36)

aniruddhaadak80:
## Summary

- **What changed?** One line: the home screen open-local card now advertises `.doc` alongside the other openable formats.
- **Why is this change needed?** Follow-up to legacy `.doc` open support (#36): the dialog filter, router, and drop bridge all accept `.doc`, but the home card still listed only the old set — its own comment demands sync with `OPEN_DIALOG_EXTENSIONS`. Files dropped onto the home card without an icon fall back to the letter badge (`D`), so no icon asset is needed.

## Related issue

Part of #36 (home-screen parity for legacy `.doc` open support)

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

## Screenshots or recordings

Not applicable — one-line advertised-formats string. Before: `.docx / .xlsx / …`. After: `.docx / .doc / .xlsx / …`.

## Contributor checklist

- [x] User-facing strings use the existing i18n resources. (no new strings — display list only)

### DIFF
--- a/apps/shell/src/renderer/src/Home.tsx
+++ b/apps/shell/src/renderer/src/Home.tsx
@@ -54,7 +54,7 @@ const FILE_ICONS: Record<string, string> = {
 /* Formats the open-local card advertises. Too long for the card at any window
    width, so it ellipsizes and a hover ScreenTip carries the full list. Keep in
    sync with the main-process open-dialog filter (OPEN_DIALOG_EXTENSIONS). */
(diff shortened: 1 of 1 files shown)


merrick-2002:

merrick-2002:

PR #218, opened 2026-09-05
fix(slides): route saveAs through the save queue

aniruddhaadak80:
## Summary
Follow-up to #155 (merged): `save()` runs through the `saveTail` promise chain, but `saveAs()` still wrote directly — a manual save racing Save As (or double Save As, or close-save) interleaves writes through the same main-process pipe. This extracts the chain into `runSerialized` and routes both through it, with identical error propagation.

## Related issue
Code-scan find (extends #155 to the sibling write path).

## Validation
- [ ] `lint` — CI
- [ ] `typecheck` — CI
- [ ] `npm test` — CI, including the new save/saveAs overlap case in `apps/slides/tests/save-serialization.test.ts` (asserts max 1 concurrent IPC write)

## Screenshots
N/A — behavior: overlapping Save + Save As serialize; failure still releases the queue for the next caller.

## Contributor checklist
- [x] `runSerialized` preserves the exact prior semantics (prior failure tolerated, tail cleared only by its owner, caller sees its own rejection)
- [x] Conventional Commits message (`fix(slides): ...`)
- [x] Regression test added

### DIFF
--- a/apps/slides/src/renderer/file-actions.ts
+++ b/apps/slides/src/renderer/file-actions.ts
@@ -46,12 +46,37 @@ export function adoptSavedSlides(ctx: ActionCtx, next: RenderSlide[]): void {
  * truncated pptx, or EPERM/EBUSY on Windows. The queue is a simple promise
  * chain: each caller awaits the previous tail, then runs its own pass.
  */
-let saveTail: Promise<boolean> | null = null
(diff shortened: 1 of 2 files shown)


merrick-2002:
Thanks for extending the save queue to Save As with the concurrency test. Solid follow-up, and thanks for your continued support.

PR #209, opened 2026-09-04
fix(ai-panels): clear restored history transcript on New chat (fixes #195)

aniruddhaadak80:
## Summary
\
ewChat\ reset the agent loop (model context) and cleared the live turn, but never cleared the restored-history transcript painted above it — reopening a document left the previous conversation on screen with no way to dismiss it. Worse, the New chat button only rendered when the live turn was non-empty, so a history-only panel offered no clear path at all.

## Related issue
Fixes #195. Same one-line omission fixed in slides, whose panel shares the pattern (sheets passes history in as a parent prop; pdf/markdown have no restored-history state).

## Validation
- [x] \ormat:check\ — verified locally with the repo-pinned Prettier 3.9.6 on all three touched files
- [ ] \lint\ — CI
- [ ] \	ypecheck\ — CI
- [ ] \
pm test\ — CI, including the new \pps/docs/tests/ai-new-chat-clears-history.test.ts\ (mounts \AiPanel\ with a mocked \projectApi\ history, clicks New chat, asserts the \.ai-msg-historic\ transcript and the button are gone)

## Screenshots
N/A — behavior: history transcript + New chat button clear together; button now also appears on history-only panels.

## Contributor checklist
- [x] Scoped to the reported symptom (docs + identical slides pattern)
- [x] Conventional Commits message (\ix(ai-panels): ...\)
- [x] Regression test added for the reported app
- [x] No i18n key changes (reuses \iNewChatTitle\)

Issue #195, opened 2026-09-04
没有办法删除historicChat，New Chat只针对当前聊天


hanxiyuan163-gif:
### App

GenOffice Docs (docs)

### Version and OS

windows11

### What happened?

上一轮对话historicChat 没有办法彻底清掉

merrick-2002:
Thanks for the report, and confirmed.

Today the AI panel restores the previous conversation for the open document as a read-only transcript above the current chat, and **New Chat** only resets the current in-memory session. There is currently no way to clear that stored history, so it comes back every time the file is reopened. This affects Docs and Slides in the same way (Sheets hides it on New Chat but it also returns on reopen).

We are designing a proper fix so that starting a new chat also clears the stored history for that document, and will follow up here once it lands.

PR #219, opened 2026-09-05
fix(slides): clear answered clarifications on New chat

aniruddhaadak80:
## Summary
Same transcript class as #195/#209, slides-only: answered clarification receipts (`clarifyAnswers`, rendered under their original message index) survived New chat, so stale entries re-attached under unrelated new messages after an index collision. One-line clear next to the `setHistoricChat` clearing.

## Related issue
Code-scan find (docs has no clarify-answers state; sheets passes history via parent props).

## Validation
- [ ] `lint` — CI
- [ ] `typecheck` — CI
- [ ] `npm test` — CI
- Note: no dedicated component test — seeding an answered clarification requires driving a live loop run through the skill trigger chain (the mock loop in `ai-panel-run-lifecycle.test.ts` never calls back into it). Omission verified against the render mapping (`ai-clarify-answered` at AiPanel.tsx:2194) and the reset path; the sibling docs pattern is covered by `ai-new-chat-clears-history.test.ts`.

## Screenshots
N/A — behavior: New chat drops answered clarification receipts along with everything else.

## Contributor checklist
- [x] One-line, same-location fix as the #209 hunk
- [x] Conventional Commits message (`fix(slides): ...`)

### DIFF
--- a/apps/slides/src/renderer/ai/AiPanel.tsx
+++ b/apps/slides/src/renderer/ai/AiPanel.tsx
@@ -1862,6 +1862,10 @@ export function AiPanel({
     loopRef.current?.reset()
     setBusy(false)
     setChat([])
+    // Answered clarifications are transcript too: they render under their
+    // original message index, so stale entries would re-attach to unrelated
+    // new messages after an index collision.
+    setClarifyAnswers([])
     sentAttachmentsRef.current = []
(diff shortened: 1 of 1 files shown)


merrick-2002:
Thanks — a tidy one-liner in the right place. Appreciate you keeping at these.

PR #207, opened 2026-09-04
docs(contributing): document TAVILY_API_KEY in env var table

aniruddhaadak80:
## Summary
Follows up #194 (Tavily web-search fallback): \CONTRIBUTING.md\'s environment-variable table listed \AI_SEARCH_DISABLE_GSK\ / \SERPER_API_KEY\ but not the new \TAVILY_API_KEY\, so contributors had no documented way to supply a Tavily key. This adds \TAVILY_API_KEY\ to that row and updates the effect text to \Disable the gsk search backend / supply a Serper or Tavily key\.

## Related issue
Related to #194 (Tavily fallback). No \Closes\ — docs follow-up, no behavior change.

## Validation
- [x] \ormat:check\ — no local \
ode_modules\, so verified equivalently: table rebuilt programmatically to Prettier column widths (Variable 59 / Effect 70, single-space borders, exact dash separators); diff is the minimal 10-line re-pad
- [ ] \lint\ — N/A (Markdown only, CI will confirm)
- [ ] \	ypecheck\ — N/A (Markdown only, CI will confirm)
- [ ] \
pm test\ — N/A (Markdown only, CI will confirm)

## Screenshots
N/A — docs table change:
\| AI_SEARCH_DISABLE_GSK, SERPER_API_KEY, TAVILY_API_KEY | Disable the gsk search backend / supply a Serper or Tavily key |\

## Contributor checklist
- [x] Change is scoped to the documented env var (no code touched)
- [x] Conventional Commits message (\docs(contributing): ...\)
- [x] Table kept Prettier-clean (uniform column widths)

PR #221, opened 2026-09-05
fix(shell): harden Home rename validation

aniruddhaadak80:
## Summary
Two gaps in the Home rename gate: only `\ / :` were rejected while Windows also forbids `* ? " < > |` plus control characters (the PDF auto-renamer already strips that fuller set), so names like `rep"ort.pdf` fell through to a raw OS error; and a case-only rename (`Report.pdf` to `report.pdf`) was wrongly rejected with `errExists` because `existsSync` sees the source itself on case-insensitive filesystems.

## Related issue
Code-scan find (no open issue covers rename validation).

## Validation
- [ ] `lint` — CI
- [ ] `typecheck` — CI
- [ ] `npm test` — CI, including the new `apps/shell/tests/home-rename-validation.test.ts` (illegal-char table incl. controls, unicode-safe names pass, case-only detection)

## Screenshots
N/A — behavior: illegal names get the localized `errBadName`; case-only renames proceed to the OS rename.

## Contributor checklist
- [x] Pure validation extracted to `rename-validation.ts` (index.ts stays electron-land, stays testable)
- [x] Character class mirrors the PDF auto-renamer set with a citing comment
- [x] Conventional Commits message (`fix(shell): ...`)
- [x] Regression tests added

### DIFF
--- a/apps/shell/src/main/index.ts
+++ b/apps/shell/src/main/index.ts
@@ -189,6 +189,7 @@ import type { TabKind } from '../shared/tabs-api'
 import { TABS_CHANNELS } from '../shared/tabs-api'
 import { showErrorDialog } from './error-dialog'
 import { normalizeRecentQuery, pageRecentPaths, statPathEntries } from './recent-files'
+import { isCaseOnlyRename, isValidRenameName } from './rename-validation'
 import { TabManager } from './tab-manager'
 import { applyUpdateChannel, initAutoUpdater } from './updater'
 import { isUpdateChannel, type UpdateChannel } from '../shared/update-api'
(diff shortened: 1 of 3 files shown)


aniruddhaadak80:
Note on the latest run: lint is green now; every test file passes and the job fails only on the known prosemirror teardown flake (ReferenceError: document is not defined from DOMObserver.flush firing after jsdom teardown) — same signature as #207/#209, both green on retrigger. Pushed an empty commit to retrigger.

merrick-2002:
Thanks for hardening the rename gate and handling case-only renames. Appreciate your continued support.

PR #224, opened 2026-09-05
fix(docs): drop staged attachments on New chat

aniruddhaadak80:
## Summary
Same New-chat leak class as #195/#209: the composer keeps unsent staged files across New chat (only consumed ones were cleared), so they silently join the next chat's file context via the sent+live merge. The typed draft is deliberately kept — only staged files and the attach notice drop.

## Related issue
Code-scan follow-up to #195 (composer half; transcript half was #209).

## Validation
- [ ] `lint` — CI
- [ ] `typecheck` — CI
- [ ] `npm test` — CI, including the new `apps/docs/tests/ai-new-chat-attachments.test.ts` (mounts AiPanel with mocked history, stages a file through the attach button, clicks New chat, asserts the composer strip and transcript are gone)

## Screenshots
N/A — behavior: New chat empties the attachment strip; the text draft survives.

## Contributor checklist
- [x] Draft preservation explicitly out of scope (collapse test celebrates draft survival)
- [x] Conventional Commits message (`fix(docs): ...`)
- [x] Regression test added, modeled on `ai-new-chat-clears-history.test.ts`


aniruddhaadak80:
Stacking note: this builds on #209 (same newChat function), so the diff currently includes that PR too — the CI failure was exactly that missing base (New-chat button is gated on restored history). Please merge #209 first; I will then rebase onto main and the diff will shrink to the attachments change + its test only.

merrick-2002:
Thanks for the composer follow-up to the New chat fix, with the mounted test. Appreciate your ongoing contributions.

PR #229, opened 2026-09-06
fix(ai-provider): allow keyless custom endpoints for local models

aniruddhaadak80:
## Summary

- What changed? activeProvider now honors custom when baseUrl+model are set even with an empty apiKey; OpenAI-compatible chat/stream omit the Authorization header when the key is empty.
- Why? Local OpenAI-compatible servers (Ollama, LM Studio, vLLM) accept anonymous requests. Previously a keyless custom config silently fell back to genspark, and an empty key was still sent as Bearer empty, which strict local gateways reject. This unblocks local-document workflows from #45 without changing any other provider's key requirement.

## Related issue

Part of #45 (custom endpoint with optional key; does not close the full settings-window proposal)

## Validation

- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test

List any checks not run and explain why: No node_modules in this checkout; relying on CI as proof (same as prior passes). Added regression tests: providers.test.ts keyless custom, chat.test.ts + stream.test.ts Authorization omission.

## Screenshots or recordings

Not applicable.

## Contributor checklist

### DIFF
--- a/packages/ai-provider/src/protocols/openai-compatible.ts
+++ b/packages/ai-provider/src/protocols/openai-compatible.ts
@@ -154,7 +154,7 @@ async function openAiCompatibleTurn(
     signal: wd.signal,
     headers: {
       'Content-Type': 'application/json',
-      Authorization: `Bearer ${config.apiKey}`,
+      ...(config.apiKey ? { Authorization: `Bearer ${config.apiKey}` } : {}),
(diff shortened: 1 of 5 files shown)


merrick-2002:
Thanks — merged. Keyless custom endpoints now work for local OpenAI-compatible servers.

PR #232, opened 2026-09-06
fix(ai-provider): trim model whitespace and remap padded retired ids

aniruddhaadak80:
## Summary

- What changed? trimConfigs now trims stored model ids (plus legacy migration); resolveAiSettings trims before the retired-model remap instead of after.
- Why? A pasted ' gpt-5.6-terra ' or ' deepseek-reasoner ' misses exact-match logic: the key/URL trim covered apiKey/baseUrl but not model, and a padded retired id (' deepseek-reasoner ') skipped the remap entirely, sending a vendor-rejected id to the API. Trim-first keeps pasted configs working.

## Related issue

Self-identified settings-robustness gap (same family as the key/URL trim).

## Validation

- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test

List any checks not run and explain why: No node_modules in this checkout; relying on CI as proof. Extended providers.test.ts (model trim, padded retired remap, legacy model trim).

## Screenshots or recordings

Not applicable.

## Contributor checklist

### DIFF
--- a/packages/ai-provider/src/providers.ts
+++ b/packages/ai-provider/src/providers.ts
@@ -264,13 +264,14 @@ export function maxOutputTokensOf(
     : clampMaxOutputTokens(settings.maxOutputTokens)
 }
 
-/** pasted keys/URLs often carry stray whitespace, which turns into a 401 with a valid key */
+/** pasted keys/URLs/model ids often carry stray whitespace, which turns into a 401 with a valid key */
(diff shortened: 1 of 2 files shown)


merrick-2002:
Thanks — merged. Trimming before the retired-id remap is the right order.

PR #233, opened 2026-09-06
fix(sheets): guard replaceOccurrences against empty needle

aniruddhaadak80:
## Summary

- What changed? replaceOccurrences returns text unchanged when find is empty.
- Why? An empty needle matches every gap: split('') inserts the replacement between every char and /(?:)/gi rewrites the cell. Callers validate non-empty via the find_replace schema, but a direct call must stay a no-op rather than corrupting data — mirrors the empty-needle guard in lazy-find replaceAllOccurrences.

## Related issue

Self-identified consistency gap (find_replace schema min(1) vs shared helper).

## Validation

- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test

List any checks not run and explain why: No node_modules in this checkout; relying on CI as proof. Added regression test in bulk-range-ops.test.ts.

## Screenshots or recordings

Not applicable.

## Contributor checklist

### DIFF
--- a/apps/sheets/src/domain/workbook-dsl.ts
+++ b/apps/sheets/src/domain/workbook-dsl.ts
@@ -1338,6 +1338,10 @@ export function replaceOccurrences(
   replace: string,
   matchCase: boolean,
 ): string {
+  // An empty needle matches every gap (split('') / /(?:)/gi): callers validate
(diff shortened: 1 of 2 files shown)


merrick-2002:
Thanks — merged. Good defensive guard for the shared helper.

PR #237, opened 2026-09-06
fix(pdf): suffix Windows reserved auto-rename bases

aniruddhaadak80:
## Summary

- What changed? sanitizeAutoRenameBase appends _ to Windows device names (CON/PRN/AUX/NUL/COM1-9/LPT1-9).
- Why? Those stems stay reserved with an extension (CON.pdf is still CON), so the no-clobber move fails on Windows and the rename silently does nothing. Suffixing keeps content-derived naming working there; harmless elsewhere.

## Related issue

Self-identified (same reserved-name family as the shell rename gate).

## Validation

- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test

List any checks not run and explain why: No node_modules in this checkout; relying on CI as proof. Added regression test in auto-rename.test.ts (CON -> CON_.pdf).

## Screenshots or recordings

Not applicable.

## Contributor checklist

### DIFF
--- a/apps/pdf/src/main/pdf-main.ts
+++ b/apps/pdf/src/main/pdf-main.ts
@@ -577,7 +577,10 @@ function sanitizeAutoRenameBase(raw: string): string | null {
     .replace(/^\.+|\.+$/g, '')
     .trim()
   if (!cleaned) return null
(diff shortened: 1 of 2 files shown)


merrick-2002:
Thanks — merged. Nice catch on the reserved stems surviving the extension.

PR #240, opened 2026-09-06
fix(electron-utils): skip drop files whose path resolution throws

aniruddhaadak80:
## Summary

- What changed? droppableFilePaths now skips a file when getPathForFile throws, instead of letting the exception abort the whole drop.
- Why? A single unmappable entry (sandboxed/proxy File the API cannot map) previously killed the entire multi-file drop — no files opened. Skipping matches the existing virtual-entry handling (empty resolutions are already dropped, not fatal).

## Related issue

Self-identified hardening (same drop pipeline as partitionDropPayload).

## Validation

- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test

List any checks not run and explain why: No node_modules in this checkout; relying on CI as proof. Added regression test in drop-open.test.ts (throwing resolver skips one file, keeps the other).

## Screenshots or recordings

Not applicable.

## Contributor checklist

### DIFF
--- a/packages/electron-utils/src/drop-open.ts
+++ b/packages/electron-utils/src/drop-open.ts
@@ -29,6 +29,15 @@ const MAX_DROPPED_FILES = 20
 /** the resolver signature webUtils.getPathForFile satisfies; injectable for tests */
 type PathResolver = (file: File) => string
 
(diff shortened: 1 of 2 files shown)


merrick-2002:
Thanks — merged. One bad entry no longer takes the whole drop down with it.

PR #247, opened 2026-09-07
fix(sheets): label force-string popup as warning not error

### LINKED ISSUE #236: UI issue: please use Warning instead of Error when formats aren't matching
### App

GenOffice Sheets (sheets)

### Version and OS

0.8.1039 on Pop_os 24 (ubuntu 24.04 lts)

### What happened?

Wrong format, but it renders correct. So it should be a "Warning", not an "Error" 

### Logs / screenshots

Image

aniruddhaadak80:
## Summary

- What changed? Sheets force-string ("number stored as text") cell popup title now reads `Warning` instead of `Error`. The copy lives in one shared `FORCE_STRING_POPUP_LOCALE` constant used by both the en-US boot locale (`App.tsx`) and the runtime language switcher (`applyUniverLocale`), so switching languages cannot regress one site.
- Why? The cell renders correctly — a format notice, not a failure (#236). Maintainer confirmed it should be a warning, not an error. The popup chrome itself is upstream Univer; this changes the label this repo controls.

## Related issue

Fixes #236.

## Validation

- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test

List any checks not run and explain why: No node_modules in this checkout; relying on CI as proof. Added `apps/sheets/tests/univer-locales.test.ts` regression test asserting the popup title is `Warning` (fails on the old `Number stored as text` title).

## Screenshots or recordings

Not applicable (string-only change; Univer popup chrome unchanged).

## Contributor checklist

### DIFF
--- a/apps/sheets/src/renderer/App.tsx
+++ b/apps/sheets/src/renderer/App.tsx
@@ -333,7 +333,7 @@ import { installMultiRowAutofit } from './autofit-multi-row'
 import { registerExcelJumpNav } from './excel-jump-nav'
 import { registerExcelShortcuts } from './excel-shortcuts'
 import { installCopyMaterialize } from './copy-materialize'
-import { applyUniverLocale } from './univer-locales'
+import { applyUniverLocale, FORCE_STRING_POPUP_LOCALE } from './univer-locales'
 import { installRuleDetail } from './univer-rule-detail'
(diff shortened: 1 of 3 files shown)


merrick-2002:
Merged — thanks for the fix and for adding a regression test with it!

Issue #220, opened 2026-09-05
GenOffice Sheets: Find/Search returns only one match per row and navigation between search results is unreliable


jeandrescas:
### App

GenOffice Sheets (sheets)

### Version and OS

Linux Debian 13

### What happened?

App: GenOffice Sheets

Version: GenOffice 0.8.1039
Operating System: Debian 13
Desktop Session: GNOME on Wayland

Bug Report: Problems with Find/Search in spreadsheets

Hello GenOffice Team,

I am experiencing two reproducible problems with the Find/Search feature in GenOffice Sheets.

1. Search does not find all matching cells within the same row

When I search for a word or value that appears multiple times in different cells of the same row, GenOffice appears to detect only one occurrence.

For example, if the same word exists in cells A10, C10, F10, and H10, the search function may identify only one of those cells instead of treating all four cells as separate search results.

Expected behavior:
Every cell containing the searched word or value should be detected as an individual match, even when several matches are located in the same row.

Actual behavior:
Only one occurrence in the row is found or presented, while other matching cells in that same row are not properly included in the search results.

2. Navigation between search results is unreliable

There are also problems when navigating through the matches using the Next/Previous search controls.

The navigation does not consistently move through every occurrence in the worksheet. It may skip matching cells, stop unexpectedly, return to a previous result, or fail to follow the expected sequential order.

Expected behavior:
The Next and Previous controls should navigate through every matching cell in a predictable sequence until all matches in the worksheet have been visited.

Actual behavior:
Navigation between matches is inconsistent and some matching cells may be skipped.

Suggested test case:

Enter the same word, for example "Example", into several cells in the same row:

A10 = Example
C10 = Example
F10 = Example
H10 = Example

Also enter "Example" in several cells on other rows.

Then:

1. Open Find/Search.
2. Search for "Example".
3. Use Next repeatedly.
4. Use Previous repeatedly.
5. Verify whether every individual matching cell is selected.

Expected result:
Every occurrence should be counted and selected individually.

Actual result:
GenOffice may detect only one occurrence from a row containing several matches, and navigation between the remaining results is unreliable.

Impact:

This is particularly problematic when working with large spreadsheets containing hundreds or thousands of rows, because the user cannot reliably verify that every occurrence of a word, value, or duplicated entry has been found.

A spreadsheet search function should guarantee that every matching cell is included, regardless of whether multiple matches are located in the same row.

[...]

PR #250, opened 2026-09-07
test(sheets): cover multi-match-per-row find navigation (issue #220)

aniruddhaadak80:
## Summary

- What changed? Adds two regression tests through the real lazy-find bridge (`facade` + `installLazyFindBridge` + file-backed scan) encoding issue #220's suggested repro: the same word in A10, C10, F10 and H10 of one row plus a match on another row. Asserts `getMatches()` surfaces all five as individual stops and that Next/Previous visit every one in order with wrap-around. No runtime change.
- Why? Locks extension-layer coverage for the reported find-navigation scenario so a future regression is caught by CI.

## Related issue

Related to #220 (regression coverage; does not change runtime behavior).

## Validation

- [ ] `npm run format:check`
- [ ] `npm run lint`
- [ ] `npm run typecheck`
- [ ] `npm test`

List any checks not run and explain why: No node_modules in this checkout; relying on CI as proof. New tests: `apps/sheets/tests/lazy-find.test.ts` ("visits every same-row match in order and wraps", "walks Previous through every same-row match in reverse").

## Screenshots or recordings

Not applicable (no UI change; test-only).

## Contributor checklist

### DIFF
--- a/apps/sheets/tests/lazy-find.test.ts
+++ b/apps/sheets/tests/lazy-find.test.ts
@@ -599,6 +599,81 @@ describe('installLazyFindBridge', () => {
     bridge.dispose()
   })
 
+  it('visits every same-row match in order and wraps (issue #220)', async () => {
+    // Issue #220's repro: the same word in A10, C10, F10 and H10 of one row
+    // must each surface as an individual stop, with Next cycling through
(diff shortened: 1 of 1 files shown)


merrick-2002:
Merged, thanks for locking down the same-row find navigation with tests.

PR #252, opened 2026-09-07
fix(sheets): fallback to cached values for external reference errors

### LINKED ISSUE #235: Cached formula values are replaced with #ERROR! when external data sources are unavailable
### App

GenOffice Sheets (sheets)

### Version and OS

0.8.1039 Pop_OS 24.04 (based on Ubuntu 24.04)

### What happened?

I opened one of my work documents that I had previously used mostly in Excel on Windows 11, and occasionally in LibreOffice. At first, the document looked fine, but then many values were replaced with `#ERROR!`. Reopening the file without saving did not fix the problem.

After investigating, I found that the main cause was external links to another document that was not accessible from my local computer. Excel and LibreOffice successfully use the cached values in this situation, while GenOffice recalculates them, encounters an error, and this results in cascading errors.

Please fix this behavior so that if an external data source is unavailable, GenOffice uses the cached value instead.

As a possible next step, it would also be useful to add functionality similar to **Edit → Links to External Files** in LibreOffice.

### Logs / screenshots

Image
Image

aniruddhaadak80:
## Summary

- What changed? `installCachedValueFallbackInterceptor` (`apps/sheets/src/renderer/formula-cached-fallback.ts`) now treats unavailable-external-link errors (`#REF!`, `#N/A`, `#ERROR!`) like `#NAME?`: they fall back to the file's cached value even when the user edited other cells, because the engine cannot resolve them under any inputs. Genuine computed errors (`#DIV/0!`, plain-arithmetic `#VALUE!`) still keep the live engine result on edited sheets, and an edited cell itself still wins. Unedited-sheet behavior is unchanged.
- Why? Workbooks with links to unreachable external documents showed cascading `#ERROR!` where Excel/LibreOffice show cached values (#235).

## Related issue

Fixes #235.

## Validation

- [ ] `npm run format:check`
- [ ] `npm run lint`
- [ ] `npm run typecheck`
- [ ] `npm test`

List any checks not run and explain why: No node_modules in this checkout; relying on CI as proof. Extended `apps/sheets/tests/formula-cached-fallback.test.ts` ("keeps computed errors on an edited sheet but falls back for #NAME? and external ref errors"); existing cases (unedited-sheet fallback, style-only edits, genuine arithmetic `#VALUE!`) are preserved.

## Screenshots or recordings

Not applicable (display-value fallback; no chrome change).

## Contributor checklist

### DIFF
--- a/apps/sheets/src/renderer/formula-cached-fallback.ts
+++ b/apps/sheets/src/renderer/formula-cached-fallback.ts
@@ -76,15 +76,29 @@ export function installCachedValueFallbackInterceptor(
       const key = `${location.row}:${location.col}`
       const edits = state.editJournal.cells.get(sheetId)
       // Style-only journal entries (hasValue false) don't change any input.
+      // A user overwrite owns the cell — keep the engine's live result.
       if (edits?.get(key)?.hasValue) return next(cell)
+      let hasValueEdits = false
+      if (edits) {
+        for (const entry of edits.values()) {
+          if (entry.hasValue) {
+            hasValueEdits = true
+            break
+          }
+        }
+      }
+      // Unavailable external links surface as #REF!, #N/A or #ERROR!: the
+      // engine cannot resolve them under any inputs, so like #NAME? they
+      // stay fallback-eligible even when the user edited elsewhere.
+      const isExternalRefError =
+        value === ErrorType.REF || value === ErrorType.NA || value === ErrorType.ERROR
       // Once the sheet has user content edits, a computed error (#DIV/0!,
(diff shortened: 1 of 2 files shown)


merrick-2002:
Thanks for the PR. Closing this one: the error value alone can't distinguish a broken external link from a genuinely computed error. With this change, a VLOOKUP that legitimately returns #N/A after the user edits the lookup key would display the stale cached value instead. #235 has already been fixed on our side by recognizing external references from the formula text, and the fix will ship in the next release.

PR #255, opened 2026-09-07
feat(ai-provider): add Codex CLI as keyless local provider

### LINKED ISSUE #212: Add Codex CLI as a keyless AI provider
### App

AI panel / providers

### Problem

GenOffice currently supports Genspark sign-in and API-key-based providers. However, users who already have an authenticated Codex CLI installation still need to configure a separate OpenAI API key to use OpenAI models in GenOffice.

It would be useful to support Codex CLI as another provider, similar to integrations available in tools such as Claudian:

Select “Codex CLI” in the provider list.

Enter or auto-detect the local codex/codex.exe path.

Reuse the Codex CLI’s existing ChatGPT authentication.

Choose a Codex-supported model.

Use GenOffice’s existing document, spreadsheet, slide, PDF, and Markdown agent tools without storing an API key.

Expected configuration:
```
{
  "provider": "codex",
  "providers": {
    "codex": {
      "apiKey": "",
      "model": "gpt-5.6-terra",
      "cliPath": "C:\\path\\to\\codex.exe"
    }
  }
}
```
This would be especially helpful for desktop users who already use Codex but do not want to create and manage a separate API key.

### Proposed solution

Add a dedicated codex provider with an authentication mode such as codex-cli.

1. Provider configuration
```
Extend the provider configuration with:

interface AiProviderConfig {
  apiKey: string
  model: string
  baseUrl?: string
  cliPath?: string
}
```
The Codex provider would:

Require cliPath instead of apiKey.

Allow selecting a model or using the Codex CLI default.

Optionally auto-detect codex from PATH.

Validate the executable with codex --version and provide a clear error if the CLI is missing or not authenticated.

2. Main-process bridge

The Codex process must run only in the Electron main process. It should not be imported into the renderer or sandboxed preload bundle.

A simple first implementation could invoke one non-interactive Codex turn using:
```
codex exec
  --json
  --ephemeral
  --skip-git-repo-check
  --sandbox read-only
  --output-schema <schema.json>
  --output-last-message <result.json>
  -
```
The complete prompt can be written through stdin to avoid Windows command-line length limits.

The process should also:

Use the configured model when provided.

Forward cancellation through the existing AbortSignal.

Treat stdout/stderr activity as keepalive activity.

Return readable CLI and authentication errors.

Clean up temporary files after each request.

3. Preserve GenOffice’s existing tool loop

Codex should not directly edit local files. Instead, it should return a structured model turn such as:
```
{
  "text": "",
  "toolCalls": [
    {
      "id": "call-1",
      "name": "replace_text",
      "inputJson": "{\"before\":\"A\",\"after\":\"B\"}"
    }
  ]
}
```
GenOffice can then:

Validate the requested tool name.

Parse inputJson.

Execute the existing GenOffice agent tool.

Add the tool result to the conversation.

Start the next Codex turn.

This preserves the current snapshot, diff, undo, validation, and document-safety behavior.

4. Image support

Inline images can be written to a temporary directory and passed using repeated --image arguments. The temporary directory should always be removed after completion or cancellation.

5. Settings UI

When “Codex CLI” is selected:

Hide the API Key and Base URL fields.

Show a “Codex executable” path field.

Explain that the local Codex sign-in is reused and no API key is required.

Keep the existing model selector and connection-test button.

6. Packaging boundary

It may be useful to expose provider metadata through a browser-safe entry point, for example:

```@genoffice/ai-provider/browser```

Only types, provider metadata, and browser-safe configuration helpers should be exported there. The module containing node:child_process, node:fs, and other Node APIs must stay in the main-process bundle.

Otherwise, importing the provider package from the sandboxed preload can cause errors such as:
```
Unable to load preload script
module not found: node:child_process
```
and leave the GenOffice window blank.

Alternative

A more advanced integration could use codex app-server, which would provide native streaming and persistent sessions. However, codex exec with structured output may be a smaller and safer MVP that fits GenOffice’s existing provider and agent-tool abstractions.

The main trade-off of the codex exec approach is that user-visible text is generally delivered after each model turn completes rather than token by token.

Image

aniruddhaadak80:
## Summary

- What changed? Adds a `codex` (Codex CLI) provider across the AI stack: catalog entry in `packages/ai-provider/src/providers.ts` (OpenAI-compatible, `needsBaseUrl`, key placeholder marks the key optional for local servers), `codex` adapter in `packages/ai-provider/src/registry.ts` (requires a base URL, key stays optional via the existing `needsBaseUrl` path in `activeProvider`), `'codex'` in `AiProviderId` (`types.ts`), and a brand mark in the shell settings provider picker (`apps/shell/src/renderer/src/provider-logos.tsx`). Tests: keyless selection in `providers.test.ts`, endpoint resolution/rejection in `registry.test.ts`.
- Why? Requested in #212: run a local Codex CLI runner as an AI backend without an API key.

## Related issue

Fixes #212.

## Validation

- [ ] `npm run format:check`
- [ ] `npm run lint`
- [ ] `npm run typecheck`
- [ ] `npm test`

List any checks not run and explain why: No node_modules in this checkout; relying on CI as proof. New tests: `packages/ai-provider/tests/providers.test.ts` ("allows keyless Codex CLI endpoints for local runners"), `packages/ai-provider/tests/registry.test.ts` ("uses the configured base URL for Codex CLI and rejects a missing one").

## Screenshots or recordings

Not applicable (settings-picker entry; select Codex CLI in AI settings to verify).

## Contributor checklist

### DIFF
--- a/apps/shell/src/renderer/src/provider-logos.tsx
+++ b/apps/shell/src/renderer/src/provider-logos.tsx
@@ -205,6 +205,11 @@ const LOGOS: Record<AiProviderId, ReactNode> = {
   ),
   'opencode-zen': opencodeLogo,
   'opencode-go': opencodeLogo,
+  codex: (
+    <svg viewBox="0 0 24 24" fill="currentColor" aria-hidden="true">
+      <path d="M12 2C6.48 2 2 6.48 2 12s4.48 10 10 10 10-4.48 10-10S17.52 2 12 2zm-2 15l-5-5 1.41-1.41L10 14.17l7.59-7.59L19 8l-9 9z" />
+    </svg>
+  ),
   custom: (
     <svg
       viewBox="0 0 24 24"
--- a/packages/ai-provider/src/providers.ts
+++ b/packages/ai-provider/src/providers.ts
@@ -210,6 +210,17 @@ export const AI_PROVIDERS: AiProviderMeta[] = [
     keyPlaceholder: 'API Key',
   },
   {
+    id: 'codex',
+    label: 'Codex CLI',
+    // Local Codex CLI runner over its OpenAI-compatible HTTP endpoint. Model
+    // ids follow the runner (any model the local Codex setup serves); the key
+    // stays optional for local servers (see activeProvider's needsBaseUrl path).
+    models: ['gpt-4o', 'gpt-4o-mini', 'gpt-4-turbo', 'o1-preview', 'o1-mini'],
+    defaultModel: 'gpt-4o',
+    keyPlaceholder: 'Optional (local)',
+    needsBaseUrl: true,
+  },
+  {
     id: 'custom',
     label: 'Custom',
     models: [],
--- a/packages/ai-provider/src/registry.ts
+++ b/packages/ai-provider/src/registry.ts
@@ -251,6 +251,21 @@ export const AI_PROVIDER_ADAPTERS: Record<AiProviderId, ProviderAdapter> = {
       }
     },
   },
+  codex: {
+    meta: metaOf('codex'),
+    capabilities: { auth: 'api-key', vision: true },
+    // Local Codex CLI runner over its OpenAI-compatible HTTP endpoint. Like
+    // the generic custom endpoint the key stays optional — only the base URL
+    // and model are required (see activeProvider's needsBaseUrl path).
+    resolveEndpoint(config) {
+      if (!config.baseUrl) throw new Error('A Codex CLI endpoint requires a Base URL')
+      return {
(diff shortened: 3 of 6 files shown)


merrick-2002:
Thanks for the PR. Closing: Codex CLI doesn't expose an OpenAI-compatible HTTP endpoint, so this provider ends up being the existing Custom entry under a different name with a hard-coded model list, and it can't reuse the Codex ChatGPT login that #212 asks for. Our policy for built-in providers is to only add widely used hosted services; anything else goes through the Custom endpoint. Appreciate the effort on the tests and the adapter wiring.

Issue #211, opened 2026-09-05
Printing dialog lacks printer settings and printed text quality is visibly degraded


Aesthermortis:
### App

GenOffice Docs (docs)

### Version and OS

Exact GenOffice v0.8.1039/ OS w11

### What happened?

Printing from GenOffice Docs exposes an extremely minimal print UI. In my case it effectively provides only:

- Print
- Cancel
- Page selection

There is no visible access to common printer/driver settings such as paper/media size or type, printer properties, print quality / DPI, scaling, orientation, dithering/halftoning, or other device-specific options.

This is not only a UI limitation: print output from GenOffice Docs is also visibly worse than output from Microsoft Word using the same thermal printer.

#### Steps to reproduce

1. Open a text document in GenOffice Docs.
2. Print it to a thermal printer.
3. Open/print equivalent text from Microsoft Word using the same printer.
4. Compare both the available printer settings and the physical print output.

#### Actual result

- GenOffice exposes almost no printer configuration.
- Printed glyphs show noticeably jagged / stair-stepped ("zigzag") edges, especially on curves and diagonals.
- Stroke density is less uniform, making the black appear lighter or more "worn" than the output from Word.
- The same printer produces substantially smoother and denser text when printing from Word.

#### Expected result

GenOffice should either expose the relevant printer/driver settings directly or provide access to the native system/driver printer properties.

Printing should preserve the printer's native quality and should not introduce visible rasterization/scaling artifacts compared with other applications.

### Possible technical area to investigate

The visual difference suggests that GenOffice may be using a different print/rendering path from applications such as Word. Possibilities worth checking include:

- printer capabilities/defaults not being fully queried or forwarded;
- incorrect paper/media dimensions causing scaling;
- low-resolution intermediate rasterization;
- antialiasing followed by monochrome dithering/thresholding;
- rendering at a non-native DPI and rescaling before the job reaches the printer driver.

These are hypotheses, not confirmed causes. The reproducible symptom is that the same printer produces visibly degraded text from GenOffice while Word produces cleaner output.

### Logs / screenshots

Comparison photos of the physical print output are available. The GenOffice output shows visibly jagged glyph edges compared with the smoother Word output.

<img width="1848" height="4000" alt="Image" src="https://github.com/user-attachments/assets/f4020613-1a4b-488e-b0fc-e0c4542be74c" />
[...]

PR #256, opened 2026-09-07
fix(pdf): increase print quality to 300 DPI with lossless PNG

aniruddhaadak80:
## Summary

- What changed? `printPdf` (`apps/pdf/src/renderer/print.ts`) now renders pages at 300 DPI instead of 150 DPI and encodes them as lossless PNG instead of JPEG (0.92). Adds a test asserting the 300/72 viewport scale and the `image/png` encoding.
- Why? Printed text was visibly degraded by low-resolution, lossy page images (#211).

## Related issue

Related to #211 (print-quality half; printer-settings dialog selection is out of scope).

## Validation

- [ ] `npm run format:check`
- [ ] `npm run lint`
- [ ] `npm run typecheck`
- [ ] `npm test`

List any checks not run and explain why: No node_modules in this checkout; relying on CI as proof. New test: `apps/pdf/tests/print.test.ts` ("renders at 300 DPI with lossless PNG for crisp text"); existing print-lifecycle tests are untouched.

## Screenshots or recordings

Not applicable (print-output fidelity; print a multi-page PDF to verify).

## Contributor checklist

### DIFF
--- a/apps/pdf/src/renderer/print.ts
+++ b/apps/pdf/src/renderer/print.ts
@@ -1,13 +1,15 @@
 import type { PDFDocumentProxy } from 'pdfjs-dist'
 
-const PRINT_SCALE = 150 / 72
+const PRINT_SCALE = 300 / 72
 
 /**
- * Sequentially render pages as JPEG images into a print-only container (canvas discarded
+ * Sequentially render pages as PNG images into a print-only container (canvas discarded
  * immediately to avoid keeping full-doc hi-res bitmaps in memory), then hand off to the
  * system print dialog; clean up after it closes (including cancel).
  * Caller flushes unsaved changes and re-getDocument first — rotations/deleted pages are
  * already in the file.
+ *
+ * Uses 300 DPI with lossless PNG for maximum text clarity (issue #211).
  */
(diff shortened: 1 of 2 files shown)


merrick-2002:
Thanks for looking into print quality. We can't take this as-is though: going from 150 DPI JPEG to 300 DPI PNG quadruples the pixel count and drops compression, and every page is kept alive as a data URL in the DOM until the dialog closes. For a scanned PDF that's easily 10-20 MB per page, so a 100-page document can push the renderer out of memory. The current settings were tuned with that constraint in mind (see the comment about discarding the canvas).

If you'd like to keep iterating: keep JPEG, and either cap the total pixel budget (e.g. scale down when numPages * pageArea exceeds a threshold) or try a more modest bump such as 200 DPI and measure memory on a large scanned file. Also note #211 is about the Docs app, not the PDF app, so this wouldn't close that issue either way.

merrick-2002:
Closing this for now since there has been no follow-up on the memory concern above (300 DPI lossless PNG pages kept alive as data URLs can push the renderer out of memory on long scanned PDFs). If you revisit it with a bounded approach (keep JPEG and cap the total pixel budget, or render pages lazily), please reopen or open a fresh PR and we'll be glad to review it.

PR #258, opened 2026-09-07
fix(docs): preserve scroll position when zooming

### LINKED ISSUE #238: DOCX viewer loses the current page position when changing zoom level
### App

GenOffice Docs (docs)

### Version and OS

0.8.1039 on Pop_os 24 (Ubuntu 24.04)

### What happened?

When viewing a DOCX document in GenOffice Docs, changing the zoom level causes the currently visible page to shift out of view and a different page to be displayed.

This happens with both Ctrl + mouse wheel and the zoom slider in the bottom-right corner.

## Expected behavior

When zooming in or out, the viewer should preserve the current viewing position. The center of the currently visible page should remain the zoom anchor, so that changing the zoom level lets the user inspect the same page at a different scale.

## Actual behavior

Changing the zoom level causes the document to scroll or jump to a different position, sometimes displaying a completely different page.

This makes it difficult to zoom in specifically to inspect the page that was visible before changing the zoom level.

### Logs / screenshots

_No response_

aniruddhaadak80:
## Summary

- What changed? `apps/docs/src/renderer/App.tsx` now preserves the viewing position across zoom changes: the Ctrl/⌘+wheel handler stashes the cursor-anchored document point in a ref, and a zoom effect rescales the scroll offsets around that anchor (viewport center for slider/menu/ribbon zoom) so the visible page stays in view.
- Why? Changing zoom (Ctrl+wheel or the zoom slider) scrolled the document to a different page (#238).

## Related issue

Fixes #238.

## Validation

- [ ] `npm run format:check`
- [ ] `npm run lint`
- [ ] `npm run typecheck`
- [ ] `npm test`

List any checks not run and explain why: No node_modules in this checkout; relying on CI as proof. Verified by reading: anchor math (`scroll = center * (newZoom / oldZoom) - viewport / 2`), ref (not DOM dataset) storage, `HTMLElement` ref typing for the `<main>` scroll container.

## Screenshots or recordings

Not applicable (scroll-position behavior; zoom in/out on a multi-page document to verify the page stays put).

## Contributor checklist

### DIFF
--- a/apps/docs/src/renderer/App.tsx
+++ b/apps/docs/src/renderer/App.tsx
@@ -516,6 +516,7 @@ export function App() {
   )
   const [status, setStatus] = useState('')
   const [zoom, setZoom] = useState(100)
+  const scrollContainerRef = useRef<HTMLElement>(null)
   // Word-style dark page (editor/dark-page.ts): on by default in the dark theme,
   // View ▸ Dark Mode flips it for the session (Word's Switch Modes); a theme
   // switch drops the override and follows the new theme again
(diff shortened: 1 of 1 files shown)


merrick-2002:
Merged, thanks! Anchoring the scroll offset around the cursor / viewport center is exactly what #238 needed.

PR #261, opened 2026-09-07
fix(ai-provider): omit temperature for OpenAI o-series reasoning models

aniruddhaadak80:
## Summary

- What changed? `modelHasFixedSampling` (`packages/ai-provider/src/registry.ts`) now also matches OpenAI's o-series reasoning models (`o1`, `o1-mini`, `o1-preview`, `o3`, `o3-mini`, `o4-mini`), so every route (direct OpenAI, Codex CLI entries, custom mirrors, OpenCode gateways) omits the temperature field for them instead of sending the hard-coded 0.3 the API rejects. Test asserts the endpoint shape for all six ids plus a negative control (`gpt-4o` still carries temperature).
- Why? The Codex model list (and any o-series selection) would otherwise fail with temperature rejections, the same class of bug the GPT-5 family fix addressed.

## Related issue

Related to #212 (the Codex CLI catalog lists `o1-preview`/`o1-mini`, which need this routing to work).

## Validation

- [ ] `npm run format:check`
- [ ] `npm run lint`
- [ ] `npm run typecheck`
- [ ] `npm test`

List any checks not run and explain why: No node_modules in this checkout; relying on CI as proof. New test: `packages/ai-provider/tests/registry.test.ts` ("marks the OpenAI o-series reasoning models as fixed-sampling"); verified the regex against every model id in the catalog for over-matching.

## Screenshots or recordings

Not applicable (request-shape fix; no UI change).

## Contributor checklist

### DIFF
--- a/packages/ai-provider/src/registry.ts
+++ b/packages/ai-provider/src/registry.ts
@@ -40,13 +40,14 @@ function metaOf(id: AiProviderId): AiProviderMeta {
  * route — vendor API, the Genspark proxy, OpenRouter's vendor-prefixed ids,
  * or a mirror behind a custom base URL. Kimi K3 answers "only 1 is allowed";
  * OpenAI's GPT-5 reasoning family rejects any temperature other than the
- * default outright. Google's Gemini 3 docs strongly recommend keeping the
- * default temperature of 1.0 for the whole Gemini 3 family, since lower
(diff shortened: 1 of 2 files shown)


merrick-2002:
Merged, thanks! The o-series models were indeed missing from the fixed-sampling list.

PR #264, opened 2026-09-08
Fix Slides PDF export for large presentations

GoodScholar:
## Summary

- Load the Slides PDF print document from a temporary HTML file instead of a top-level Base64 data URL.
- Preserve the existing slide rendering, page sizing, image decoding wait, PDF options, output, and open-after-export behavior.
- Add regression coverage for valid PNG content larger than Chromium's navigation URL limit and for temporary-file cleanup on success and failure.

This is needed because encoding every rendered slide PNG inside an HTML data URL can exceed Chromium's navigation URL limit and fail with `ERR_INVALID_URL (-300)` for large presentations.

## Related issue

Not linked.

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [ ] `npm test`

Full `npm test` was not run because the local environment has no Rust `cargo` for the Sheets sidecar, and two pre-existing macOS system-font assertions differ on this machine. Validation performed instead:

- `npx vitest run --exclude tests/system-fonts.test.ts` in `apps/slides`: 702 tests passed
- `npm test -- --run tests/pdf-export.test.ts` in `apps/slides`: 3 tests passed
- `npm run build` in `apps/slides`
- Manual export of the presentation that originally reproduced `ERR_INVALID_URL (-300)`

`npm run lint` completed with 0 errors and 12 existing warnings outside this change.

## Screenshots or recordings

Not applicable. This fixes an export failure without changing the UI.

## Contributor checklist

- [x] User-facing strings use the existing i18n resources. No strings were added or changed.
- [x] File open/save changes include an appropriate round-trip or fidelity test. The PDF export path has regression coverage for complete image payloads and output bytes.

### DIFF
--- a/apps/slides/src/main/pdf-export.ts
+++ b/apps/slides/src/main/pdf-export.ts
@@ -1,0 +1,84 @@
+import { mkdtemp, rm, writeFile } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+
+export interface PdfExportWindow {
+  loadFile(path: string): Promise<void>
+  webContents: {
+    executeJavaScript(script: string, userGesture?: boolean): Promise<unknown>
+    printToPDF(options: Electron.PrintToPDFOptions): Promise<Buffer>
+  }
+  destroy(): void
+}
+
+export interface ExportSlidesPdfOptions {
+  pngsBase64: string[]
+  widthPx: number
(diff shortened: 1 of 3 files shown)


merrick-2002:
Merged, thank you! Nice catch on the Chromium navigation URL limit, and the regression tests are appreciated.

PR #267, opened 2026-09-09
feat(ai-provider): integrate Codex App Server

### LINKED ISSUE #212: Add Codex CLI as a keyless AI provider

Sun8854:
## Summary

- What changed?
  - Adds a dedicated Codex provider backed by the official `codex app-server --listen stdio://` JSON-RPC transport.
  - Reuses the user's existing Codex ChatGPT sign-in; no OpenAI API key or HTTP endpoint is required.
  - Discovers models dynamically through `model/list`, including the account default, instead of shipping a hard-coded model list.
  - Preserves GenOffice's document-tool loop: Codex returns validated structured tool calls, while GenOffice continues to execute edits and retain its snapshot, diff, undo, and validation behavior.
  - Keeps native Codex threads across GenOffice tool rounds and follow-up messages, with cancellation, image inputs, reasoning activity, watchdogs, and idle process cleanup.
  - Runs all Node/process code in Electron main only and exposes a browser-safe provider metadata entry point for renderer/preload consumers.
  - Auto-detects the current Codex desktop installation on Windows. Saved hash-version paths migrate to the newest complete installation after Codex updates; custom paths remain supported.
  - Adds localized settings UI, a dynamically populated model picker, a connection test, and a concise auto-detection path hint.
- Why is this change needed?
  - Users who already sign in to Codex with ChatGPT should not need a second API credential to use GenOffice AI features.
  - Unlike #255, this implementation does not treat Codex as an OpenAI-compatible HTTP server or duplicate the Custom provider. Codex CLI exposes no such endpoint; the integration uses its supported App Server protocol directly.

## Related issue

Closes #212

## Validation

- [x] `npm run format:check`
- [x] `npm run lint` — 0 errors; 12 pre-existing warnings outside this change
- [x] `npm run typecheck`
- [x] `npm test -w @genoffice/ai-provider` — 176 tests passed
- [x] `npm test -w @genoffice/agent-core` — 79 tests passed
- [x] `npx vitest run tests/settings-analytics.test.ts tests/onboarding.test.ts` in `apps/shell` — 4 tests passed
- [x] `npm run build:all`
- [x] Live signed-in Codex App Server smoke test — an intentionally stale managed path resolved to the current installation; `model/list` returned 5 visible models with `gpt-6-astra` as the default

The full `npm test` command could not complete locally because the existing Electron 43.3.0 dependency attempted to download its binary and the download failed. Before that unrelated setup failure, the Shell suite reported 214 passing tests. The affected provider, agent transport, settings UI, typecheck, and complete production build all pass.

## Screenshots or recordings

The Codex settings pane now hides API Key and Base URL, loads the account's model catalog dynamically, and leaves the executable field optional with an auto-detection hint. A live packaged-app smoke test was also performed locally.

## Contributor checklist

- [x] File open/save behavior is unchanged; Codex tool execution remains inside GenOffice's existing round-trip and validation path.

屏幕截图 2026-09-09 104535

屏幕截图 2026-09-09 102437

### DIFF
--- a/apps/docs/src/main/docs-main.ts
+++ b/apps/docs/src/main/docs-main.ts
@@ -69,6 +69,7 @@ import {
   type GenSparkAccountStatus,
   type LegacyAiSettings,
 } from '@genoffice/ai-provider'
+import { listCodexModels, shutdownCodexAppServers } from '@genoffice/ai-provider/codex-app-server'
 import {
   ensureGenofficeLogin,
   gskApiKey,
@@ -2624,6 +2625,7 @@ const activeAiStreams = new Map<string, AbortController>()
  * sheets' standalone AI handlers use the same channel names.
  */
 export function registerAiIpc(): void {
+  app.once('before-quit', shutdownCodexAppServers)
   ipcMain.handle('ai:get-settings', (): AiSettings => {
     const stored = readJson<Partial<AiSettings> & LegacyAiSettings>(SETTINGS_PATH(), {})
     // pre-lock legacy file: genspark selected with cloud tools opted out. The
@@ -2660,6 +2662,10 @@ export function registerAiIpc(): void {
     writeJson(SETTINGS_PATH(), settings)
   })
 
+  ipcMain.handle('ai:codex-models', async (_event, cliPath: unknown) => {
+    return listCodexModels(typeof cliPath === 'string' ? cliPath : undefined)
+  })
+
   ipcMain.handle('ai:stream', async (event, request: AiStreamRequest) => {
     const { requestId, settings, system, messages } = request
     const tools = request.tools ?? []
@@ -2673,7 +2679,7 @@ export function registerAiIpc(): void {
     const send = (chunk: AiStreamChunk) => {
       if (!event.sender.isDestroyed()) event.sender.send('ai:stream-chunk', chunk)
     }
-    if (!config?.apiKey) {
+    if (!config || (provider !== 'codex' && !config.apiKey)) {
       send({
         requestId,
         type: 'error',
@@ -2681,7 +2687,7 @@ export function registerAiIpc(): void {
       })
       return
     }
-    if (!config.model) {
+    if (provider !== 'codex' && !config.model) {
       send({ requestId, type: 'error', error: tm('errNoModel') })
       return
     }
@@ -2698,6 +2704,7 @@ export function registerAiIpc(): void {
     try {
       let stopReason: string | undefined
       await streamForProvider(provider, config, system, messages, tools, maxTokens, {
+        ...(request.sessionId ? { sessionId: request.sessionId } : {}),
         signal: controller.signal,
         onDelta: (text) => send({ requestId, type: 'delta', text }),
         onReasoningDelta: (text) => send({ requestId, type: 'reasoning', text }),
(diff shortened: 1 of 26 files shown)


merrick-2002:
Merged, thank you! Great to have a Codex provider that speaks the official app-server protocol and lets ChatGPT-plan users run GenOffice AI without a separate API key. Appreciate the thorough write-up and the live smoke test.

Issue #196, opened 2026-09-04
表格兼容性问题


dct74:
### App

GenOffice Sheets (sheets)

### Version and OS

0.8.1039 on macOS 26.6.2

### What happened?

发现两个问题：
1. 当表格有冻结行时，打开后冻结行失效，并且单元格内容不全
2. 当既有冻结行又有冻结列时，表格打开后内容完全为空白了，变成一个空表格了

以上是在有冻结行和冻结列的表格上发现的，其他条件下是否同样有这个问题未知。

<img width="2806" height="2232" alt="Image" src="https://github.com/user-attachments/assets/f20bae97-ef33-430c-bd12-703ae2155435" />
<img width="3030" height="2316" alt="Image" src="https://github.com/user-attachments/assets/80872656-69bb-4b01-9a7a-61977a233af8" />

merrick-2002:
Thanks for the report. The blank sheet in your second screenshot is not caused by the frozen panes: the status bar shows "Workbook contains an unsafe ZIP path", which means our importer rejected the whole file because one of its ZIP entries has a non-standard name (most likely a leading "/"). We are making the importer tolerant of this.

To confirm the fix also covers your first case (frozen rows only, lost freeze and missing cells), could you attach the original .xlsx file, or tell us which application produced it (WPS, Excel, Numbers, an export tool, etc.)? A copy with the cell contents replaced is fine.

dct74:
附件是原始表格文件，是其他人发给我的，但是大概率是wps生成的，因为国内使用 wps 很普遍。

[常见错误代码汇总.xlsx](https://github.com/user-attachments/files/31823220/default.xlsx)
[常用配件价格表.xls](https://github.com/user-attachments/files/31823221/default.xls)

aniruddhaadak80:
[...]

PR #270, opened 2026-09-09
fix(sheets): use -1 for unfrozen axis when opening frozen panes

aniruddhaadak80:
## Summary

- What changed? File-open freeze mapping now emits Univer's -1 sentinel on the unfrozen axis (via a new exported toUniverFreeze helper), matching the in-app freeze-top-row/freeze-first-col commands.
- Why is this change needed? Opening a rows-only or columns-only frozen sheet passed startRow/startColumn 0, misaligning the frozen viewport (issue #196: missing rows / blank grid).

## Related issue

Related to #196.

## Validation

- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test

List any checks not run and explain why: No node_modules in this checkout; relying on CI as proof. New test: apps/sheets/tests/univer-range-loading.test.ts maps rows-only/cols-only/both/none through loadWorkbookSkeleton.

## Screenshots or recordings

Not applicable.

## Contributor checklist

### DIFF
--- a/apps/sheets/src/renderer/univer-sync.ts
+++ b/apps/sheets/src/renderer/univer-sync.ts
@@ -362,6 +362,26 @@ export function applyFormatPatchToRange(
   }
 }
 
+/**
+ * File freeze state mapped onto Univer's `setFreeze` shape for workbook open.
+ * Univer uses -1 as the "no freeze on this axis" sentinel (see the
+ * freeze-top-row/freeze-first-col commands); a literal 0 start misaligns the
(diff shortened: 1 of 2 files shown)


merrick-2002:
Merged, thanks! Aligning the open path with the -1 sentinel the in-app freeze commands already use is a good consistency fix. One note for the record: issue #196 turned out to be an unsafe ZIP entry name and was fixed in v0.9.10, so this change is a normalization rather than the fix for that report.

PR #275, opened 2026-09-09
fix(sheets): handle escaped quotes in CSV delimiter sniffing

aniruddhaadak80:
## Summary

- What changed? The CSV delimiter sniffer now treats an escaped quote pair the same way the row parser does: it stays inside the quoted field instead of toggling quoting.
- Why is this change needed? A quoted field like counting five semicolons inside escapes outvoted the two true commas, so the file mis-split into wrong columns.

## Related issue

Related to #196.

## Validation

- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test

List any checks not run and explain why: No node_modules in this checkout; relying on CI as proof. New test: apps/sheets/tests/csv-import.test.ts ignores delimiters inside escaped quotes when sniffing (fails on the old counting).

## Screenshots or recordings

Not applicable.

## Contributor checklist

### DIFF
--- a/apps/sheets/src/gateway/csv-import.ts
+++ b/apps/sheets/src/gateway/csv-import.ts
@@ -82,9 +82,18 @@ export function sniffDelimiter(text: string): string {
   const counts = new Map<string, number>(DELIMITERS.map((d) => [d, 0]))
   for (const line of sample) {
     let quoted = false
-    for (const character of line) {
-      if (character === '"') quoted = !quoted
-      else if (!quoted && counts.has(character)) {
+    for (let index = 0; index < line.length; index += 1) {
+      const character = line[index]
+      if (character === undefined) continue
+      if (character === '"') {
+        // An escaped quote ("") stays inside the quoted field; only a lone
+        // quote toggles quoting, mirroring parseCsv below.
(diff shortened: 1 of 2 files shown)


merrick-2002:
Thanks, the sniffer change itself looks right. Two things before this can go in:

1. `typecheck` fails: `apps/sheets` compiles with `noUncheckedIndexedAccess`, so `line[index]` is `string | undefined`. `line.charAt(index)` (or a `?? ''` fallback) fixes it.
2. The new `parseCsv` assertion is wrong: the parser unescapes `""` to a single quote, so the middle field comes out as `; "; "; "; "`. The test never ran in CI because typecheck stopped the job first.

Happy to merge once those are updated.

merrick-2002:
Thanks for the update, typecheck passes now. The remaining CI failure is the second point above: `parseCsv` unescapes `""` to a single `"`, so the assertion should be

```ts
expect(parseCsv('a,"; ""; ""; ""; """,d')).toEqual([['a', '; "; "; "; "', 'd']])
```

The `sniffDelimiter` assertion is fine. Push that and we'll merge.

merrick-2002:
Merged, thanks for the fix!

Issue #228, opened 2026-09-06
文件转换后使用卡顿问题


131415927:
### App

GenOffice Docs (docs)

### Problem

ATT，经过转换后的文件进行编辑时会变的巨卡顿，还会出现丢失GENoffice软件图标的情况

merrick-2002:
Thanks for the report. To reproduce this we need a bit more information:

- Which conversion did you use (e.g. PDF → Word), and roughly how large is the file (pages / size)?
- Your OS and GenOffice version.
- What do you mean by the app icon going missing — the taskbar/dock icon, the window icon, or the file-type icon?

If possible, please attach a redacted sample file (with any sensitive content removed or replaced) that reproduces the lag. That would help us a lot.

merrick-2002:
We were able to reproduce the lag with large converted documents (long text and pages full of pinned pictures) without a sample, so no need to send one.

v0.10.639 (https://github.com/genspark-ai/genoffice/releases/tag/v0.10.639) makes long documents open several times faster and cuts typing latency a few-fold, and PDF → Word output now groups a page's pinned pictures so the result stays editable. Picture-heavy documents also open lazily instead of loading every image up front.

We could not reproduce the missing app icon; if it still happens, please open a separate issue with your OS and GenOffice version. Closing this one; reopen if the lag persists on v0.10.639.

PR #281, opened 2026-09-10
fix(font-metrics): evict oldest advance-cache entry instead of clearing

aniruddhaadak80:
## Summary

- What changed? The parsed-face advance cache now evicts only the oldest entry when full (via a new exported evictOldestEntry helper using Map insertion order) instead of clearing the entire cache.
- Why is this change needed? Clearing all 8 parsed faces on the 9th family forces full hmtx/cmap re-parses during layout, a perf cliff on documents cycling many families.

## Related issue

Related to #228.

## Validation

- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test

List any checks not run and explain why: No node_modules in this checkout; relying on CI as proof. New tests: packages/font-metrics/tests/advance.test.ts covers no-op below capacity, single oldest eviction, and repeated eviction.

## Screenshots or recordings

Not applicable.

## Contributor checklist

### DIFF
--- a/packages/font-metrics/src/advance.ts
+++ b/packages/font-metrics/src/advance.ts
@@ -143,6 +143,20 @@ const styleTokensOf = (style: AdvanceStyle): string[] => [
 const faceCache = new Map<string, ParsedFace | null>()
 const FACE_CACHE_MAX = 8
 
+/**
+ * Make room for one entry, evicting the least-recently-inserted one.
+ * Map preserves insertion order, so the first key is the oldest. Evicting
+ * one entry (instead of clearing the whole cache) keeps the other parsed
+ * faces hot when a document cycles through more families than fit.
+ */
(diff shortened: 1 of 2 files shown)


merrick-2002:
Thanks, evicting one face instead of clearing the cache is a sensible change. Two fixes before it can go in:

1. The new test `evicts repeatedly until under max` fails in CI: `evictOldestEntry(cache, 2)` on four entries evicts down to one (`['d']`), since the helper makes room for the entry about to be inserted (size < max). Either expect `['d']` or rename/redefine the helper. The call site is correct as is.
2. The existing `marks unmapped codepoints as NaN` test was changed from `'\u{10FFF0}x'` to a literal U+FFFF. That alters what the test checks (supplementary-plane vs BMP noncharacter) and puts a raw noncharacter in the source. Please revert that hunk; the PR should only touch the eviction.

Happy to merge once CI is green.

Issue #251, opened 2026-09-07
Need Print feature


szkane:
### App

Whole suite

### Problem

As an Office suite, shouldn't the most basic printing function still be retained?

merrick-2002:
Thanks for the report! Printing is already available in GenOffice Docs, Slides and PDF: use **File > Print…** from the menu bar or press **Ctrl+P / ⌘P**. Docs and Slides open a print preview with page range options and then hand off to the system print dialog; PDF prints the document directly.

Could you tell us which app (Docs / Sheets / Slides / PDF / Markdown / HTML), which OS and which version you were using when you couldn't find Print? That will help us pin down whether an entry point is missing in a specific app or just hard to discover. Thanks!

merrick-2002:
Shipped in v0.10.63: Sheets now has File > Print (Ctrl+P / ⌘P) with the system print dialog, and print / PDF export draw charts, pictures and shapes. Docs, Slides and PDF already had it. Feel free to reopen if it does not work for you.

PR #288, opened 2026-09-10
feat(pdf): add print range dialog (all/current/custom)

aniruddhaadak80:
## Summary

- What changed? Print toolbar button, menu Print, and ⌘P now open a range dialog (All pages / Current page / Custom range with Word-style 1,3,5-8 input validated by the existing parsePageRanges) instead of rastering the whole document; printPdf accepts an optional 1-based page list.
- Why is this change needed? PDF had no way to print the current page or a custom range, while docs and slides both do (issue #251).

## Related issue

Related to #251 (PDF part; Sheets/HTML print still to come).

## Validation

- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test

List any checks not run and explain why: No node_modules in this checkout; relying on CI as proof. New strings printRangeAll/Current/Custom/Hint in all 20 locales (wording mirrors docs); new test: printPdf renders only the requested pages sorted with out-of-range dropped.

## Screenshots or recordings

Not applicable (standard pdf-modal dialog; open Print on a multi-page PDF to verify).

## Contributor checklist


merrick-2002:
Thanks @aniruddhaadak80, merged! Nice addition, and the subset test is appreciated.

PR #290, opened 2026-09-10
fix(pdf): raise print raster to 200 DPI within a pixel budget

aniruddhaadak80:
## Summary

- Follow-up to #256 addressing the memory concern: keeps JPEG (q0.92) and raises the raster from 150 to 200 DPI only while the whole document fits a 150 MP total pixel budget (~40 US-Letter pages at full quality).
- Larger documents scale down proportionally toward — but never below — the 150 DPI baseline, so quality never regresses vs today and the JPEG data URLs held alive until the dialog closes stay within tens of MB.
- Pages are fetched once each; areas are measured at scale 1 (cheap sync math, no bitmaps) before the render pass; the canvas is still discarded immediately after rasterizing.

## Related issue

None — per the review on #256, #211 is about the Docs app, not the PDF app, so this intentionally closes no issue.

## Validation

- [x] prettier check on touched files
- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test

List any checks not run and explain why: full repo checks not run locally (no node_modules in this checkout); extended apps/pdf/tests/print.test.ts with unit tests for the budget curve plus integration tests for a 2-page doc (200 DPI) and a 200-page doc (150 DPI floor). Relying on CI.

## Screenshots or recordings

Not applicable (raster-density change; no chrome touched).

## Contributor checklist

### DIFF
--- a/apps/pdf/src/renderer/print.ts
+++ b/apps/pdf/src/renderer/print.ts
@@ -1,6 +1,31 @@
-import type { PDFDocumentProxy } from 'pdfjs-dist'
+import type { PDFDocumentProxy, PDFPageProxy } from 'pdfjs-dist'
 
-const PRINT_SCALE = 150 / 72
+/** Baseline raster density (previous behavior — quality never drops below this). */
+const BASE_DPI = 150
+/** Target raster density for sharper glyph edges on high-DPI printers. */
+const TARGET_DPI = 200
+/**
+ * Upper bound on total raster pixels across all pages. A 200 DPI US-Letter
+ * page is ~3.7 MP, so ~40 such pages print at full target quality; larger
+ * documents scale down toward (never below) the 150 DPI baseline. This keeps
+ * the JPEG data URLs held alive until the dialog closes within ~tens of MB
+ * instead of pushing the renderer out of memory on long scanned PDFs.
+ */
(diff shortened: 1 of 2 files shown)


merrick-2002:
Thanks for coming back to this with the bounded approach — the pixel budget is the right shape. One blocker before we can merge: main now has a print-range dialog (#288), and `App.tsx` calls `printPdf(pdoc, pages)`, so dropping the `pages` parameter fails typecheck and would silently print the whole document. Please rebase onto main, keep the `pages` filter, and compute the area budget over the selected pages only. Happy to merge after that.

merrick-2002:
Closing this for now since it has been waiting on the rebase for a while and `printPdf` on main still needs the `pages` parameter kept (see the earlier comment). The bounded-DPI approach is still the right shape — if you rebase onto main, keep the page filter, and compute the pixel budget over the selected pages only, please reopen or send a fresh PR and we'll pick it up quickly. Thanks!

aniruddhaadak80:
Rebased onto current main as requested: the pages filter is kept and the pixel budget is now computed over the selected pages only, with a new test locking that a small range out of a huge document still prints at 200 DPI. Reopening the closed PR is blocked by GitHub after the rebase push, so continuing in a fresh PR: #316.

PR #291, opened 2026-09-10
Fix macOS tab bar spacer on Windows and Linux

renv1x:
## Summary

- Changed the tab bar drag spacer to use `0px` by default.
- Restored the `78px` spacer only on macOS via the existing `body.vib` class.
- This removes the unnecessary left offset of the Home tab on Windows and Linux while preserving the macOS traffic-light spacing.

## Related issue

N/A

## Validation

- [ ] `npm run format:check`
- [ ] `npm run lint`
- [ ] `npm run typecheck`
- [ ] `npm test`

Not run. This is a small CSS-only platform-specific UI change and was manually verified on Windows.

## Screenshots or recordings

Before: Home tab has an unnecessary ~78px left spacer on Windows.
gh-before

After: Home tab aligns correctly near the left edge while keeping the normal tab-strip padding.
gh-after

## Contributor checklist

- [x] User-facing strings use the existing i18n resources. (N/A — no user-facing strings changed.)
- [x] File open/save changes include an appropriate round-trip or fidelity test. (N/A — no file open/save behavior changed.)

### DIFF
--- a/apps/shell/src/renderer/src/tabbar.css
+++ b/apps/shell/src/renderer/src/tabbar.css
@@ -27,8 +27,12 @@ html.genoffice-popover-open .tab-bar {
 
 /* room for the macOS traffic lights (titleBarStyle: hiddenInset) */
 .tab-bar-drag-spacer {
-  width: 78px;
+  width: 0;
   flex-shrink: 0;
+}
+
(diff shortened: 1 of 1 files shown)


merrick-2002:
Merged, thanks for the fix and the before/after screenshots!

PR #302, opened 2026-09-11
fix(i18n): require BCP-47 boundary in normalizeLang prefix match

aniruddhaadak80:
## Summary

- What changed? `normalizeLang` in `packages/i18n/src/index.ts` now requires an exact language code or a BCP-47 delimiter (hyphen or underscore) after the code. Boundary regression tests were added in `packages/i18n/tests/i18n.test.ts`.
- Why is this change needed? The old prefix match accepted any string starting with a supported code, so ordinary words such as deleted, french, thread, italian, and hebrew were misdetected as German, French, Thai, Italian, and Hebrew, and the UI could start in the wrong language.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the targeted i18n workspace suite was run locally with all tests passing; the full suite runs in CI. This branch also carries the two red-main test corrections from PR 314 because the base revision fails those tests without them; those extra commits will be dropped on rebase once PR 314 lands.

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/apps/sheets/tests/csv-import.test.ts
+++ b/apps/sheets/tests/csv-import.test.ts
@@ -103,7 +103,7 @@ describe('parseCsv', () => {
     // The "" escape keeps the field quoted: all five inner ; must not count,
     // or they outvote the two true commas and the columns mis-split.
     expect(sniffDelimiter('a,"; ""; ""; ""; """,d')).toBe(',')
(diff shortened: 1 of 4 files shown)


merrick-2002:
Merged, thank you! The boundary tests (`deleted`, `french`, `thread`…) are a nice regression net for a bug that was easy to miss.

PR #303, opened 2026-09-11
fix(ai-provider): match JSON Content-Type case-insensitively

aniruddhaadak80:
## Summary

- What changed? `jsonBodyInsteadOfSse` in `packages/ai-provider/src/protocols/shared.ts` lowercases the Content-Type header value before checking for a JSON body. A casing regression test was added in `packages/ai-provider/tests/stream.test.ts`.
- Why is this change needed? Gateways and proxies may answer with a capitalized Content-Type such as Application JSON. The case-sensitive check missed those bodies, so all three streaming paths fell through to the SSE parser, found no data lines, and reported an empty stream instead of emitting the complete JSON reply or detecting the credits-exhausted notice.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the full ai-provider workspace suite was run locally with all tests passing; the full suite runs in CI. This branch also carries the two red-main test corrections from PR 314 because the base revision fails those tests without them; those extra commits will be dropped on rebase once PR 314 lands.

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/apps/sheets/tests/csv-import.test.ts
+++ b/apps/sheets/tests/csv-import.test.ts
@@ -103,7 +103,7 @@ describe('parseCsv', () => {
     // The "" escape keeps the field quoted: all five inner ; must not count,
     // or they outvote the two true commas and the columns mis-split.
     expect(sniffDelimiter('a,"; ""; ""; ""; """,d')).toBe(',')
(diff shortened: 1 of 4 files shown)


merrick-2002:
Merged — thanks. Case-insensitive Content-Type matching is the right call for proxies and gateways, and the test covers the charset variants too.

PR #305, opened 2026-09-11
fix(docs,slides): accept spaces around dashes in print ranges

aniruddhaadak80:
## Summary

- What changed? The print-range parsers in `apps/docs/src/renderer/print-range.ts` and `apps/slides/src/shared/print-html.ts` now tokenize range-or-single inputs with a pattern match instead of splitting on whitespace, rejecting any leftover characters outside the known separators. Spaced hyphen and en-dash cases were added to both test files.
- Why is this change needed? Both parsers split on whitespace before matching, so an input like 1 - 3 became three fragments and the lone dash failed validation, rejecting the whole range. The range pattern already allowed whitespace around the dash, but the pre-split destroyed it. Space-separated pages keep working as before.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the targeted docs and slides suites were run locally with all tests passing; the full suite runs in CI. This branch also carries the two red-main test corrections from PR 314 because the base revision fails those tests without them; those extra commits will be dropped on rebase once PR 314 lands.

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/apps/docs/src/renderer/print-range.ts
+++ b/apps/docs/src/renderer/print-range.ts
@@ -5,8 +5,10 @@
  */
 export function parsePrintRange(text: string, max: number): number[] | null {
   const out = new Set<number>()
-  const parts = text.split(/[,，、;；\s]+/).filter(Boolean)
+  const tokenRe = /\d+\s*[-–]\s*\d+|\d+/g
+  const parts = text.match(tokenRe) ?? []
   if (parts.length === 0) return null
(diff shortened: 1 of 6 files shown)


merrick-2002:
Merged — thank you. Tokenising instead of pre-splitting on whitespace is the right fix, and the leftover-character check keeps the malformed cases rejected.

PR #306, opened 2026-09-11
fix(pdf): ignore invalid font sizes when decoding text styles

aniruddhaadak80:
## Summary

- What changed? `decodeStyle` in `apps/pdf/src/renderer/color-runs.ts` now only accepts finite font sizes greater than zero; anything else inherits the draft-level size. Non-numeric, infinite, zero, negative, and valid decimal cases were added in `apps/pdf/tests/color-runs.test.ts`.
- Why is this change needed? Any non-empty size field was assigned through a plain number conversion, so corrupt, hand-edited, or machine-produced style keys decoded to Not-a-Number or infinite sizes. The downstream preview multiplies the size into CSS output and the value also flows to the engine over IPC.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the targeted color-runs suite was run locally with all tests passing; the full suite runs in CI. This branch also carries the two red-main test corrections from PR 314 because the base revision fails those tests without them; those extra commits will be dropped on rebase once PR 314 lands.

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/apps/pdf/src/renderer/color-runs.ts
+++ b/apps/pdf/src/renderer/color-runs.ts
@@ -48,7 +48,10 @@ export function decodeStyle(key: string): CharStyle {
   const out: CharStyle = {}
   if (color) out.color = color
   if (font) out.font = font
-  if (size) out.size = Number(size)
+  if (size) {
+    const n = Number(size)
+    if (Number.isFinite(n) && n > 0) out.size = n
+  }
   if (bold) out.bold = bold === '1'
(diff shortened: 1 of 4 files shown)


merrick-2002:
Merged, thanks for hardening `decodeStyle` against non-finite sizes.

PR #307, opened 2026-09-11
fix(pdf): clamp and round channels in rgb255ToHex

aniruddhaadak80:
## Summary

- What changed? Both `rgb255ToHex` copies in `apps/pdf/src/renderer/color-presets.ts` and `apps/pdf/src/renderer/text-edit-preview.tsx` now round each channel and clamp it to the valid byte range before hex encoding. Clamping coverage was added in `apps/pdf/tests/color-presets.test.ts`.
- Why is this change needed? Raw channel values were formatted directly, so out-of-range colors from the PDF engine produced invalid CSS output and broke the hex color invariant and the hex round-trip used by the pickers.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the targeted color-presets suite was run locally with all tests passing; the full suite runs in CI. This branch also carries the two red-main test corrections from PR 314 because the base revision fails those tests without them; those extra commits will be dropped on rebase once PR 314 lands.

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/apps/pdf/src/renderer/color-presets.ts
+++ b/apps/pdf/src/renderer/color-presets.ts
@@ -26,7 +26,13 @@ export const hexTo255 = (hex: string): [number, number, number] => [
 ]
 
 export const rgb255ToHex = (c: readonly [number, number, number]): string =>
-  `#${c.map((v) => v.toString(16).padStart(2, '0')).join('')}`
+  `#${c
+    .map((v) =>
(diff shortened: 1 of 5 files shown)


merrick-2002:
Merged — thanks for clamping and rounding both copies of `rgb255ToHex`.

PR #308, opened 2026-09-11
fix(sheets): preserve consecutive CR breaks in clipboard TSV

aniruddhaadak80:
## Summary

- What changed? The clipboard normalizer in `apps/sheets/src/renderer/clipboard-tsv.ts` now converts each carriage return individually instead of collapsing runs of them. The existing test expectation was updated and single, paired, and Windows-style line-break cases were added in `apps/sheets/tests/clipboard-tsv.test.ts`.
- Why is this change needed? The old pattern merged consecutive classic-Mac line breaks into one newline, so a cell containing two breaks pasted as a single line break and lost a line. The sibling CSV export path already normalizes each break individually.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the replacement logic was verified locally with a standalone reproduction script; the full suite runs in CI. This branch also carries the two red-main test corrections from PR 314 because the base revision fails those tests without them; those extra commits will be dropped on rebase once PR 314 lands.

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/apps/sheets/src/renderer/clipboard-tsv.ts
+++ b/apps/sheets/src/renderer/clipboard-tsv.ts
@@ -34,7 +34,7 @@ export function clipboardField(cell: ClipboardCell | null | undefined): string {
         (typeof cell.v === 'string' ? cell.v : extractPureTextFromCell(cell)))
   // Only normalize line endings: trailing newlines are significant cell
   // content and must survive (quoted) instead of being stripped.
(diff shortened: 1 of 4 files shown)


merrick-2002:
Merged, thank you. You are right that `\\r+` swallowed a blank line; Univer uses `\\r` as its paragraph separator so consecutive breaks are real content. Good catch.

PR #310, opened 2026-09-11
fix(docs): include write-protection flags in dirty check

aniruddhaadak80:
## Summary

- What changed? `DocDirtyState` and `isDocDirty` in `apps/docs/src/renderer/doc-dirty.ts` now include the write-protection and remove-personal-info dirty flags. Both flags were added to the test clean state and to the regression matrix in `apps/docs/tests/doc-dirty.test.ts`.
- Why is this change needed? The context object already carried both flags and the save path already serialized them, but the dirty check ignored them. Toggling only the modify password or the remove-personal-info setting looked clean, so the close guard, autosave tick, and crash-recovery push could all skip a document with real unsaved protection changes.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the targeted dirty-state suite was run locally with all tests passing; the full suite runs in CI. This branch also carries the two red-main test corrections from PR 314 because the base revision fails those tests without them; those extra commits will be dropped on rebase once PR 314 lands.

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/apps/docs/src/renderer/doc-dirty.ts
+++ b/apps/docs/src/renderer/doc-dirty.ts
@@ -27,6 +27,8 @@ export interface DocDirtyState {
   themeColorsDirty: boolean
   commentsDirty: boolean
   protectionDirty: boolean
+  writeProtectionDirty: boolean
+  removePersonalInfoDirty: boolean
 }
 
 export function isDocDirty(s: DocDirtyState): boolean {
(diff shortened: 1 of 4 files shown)


merrick-2002:
Merged, thanks! Both flags were already serialised on save, so the dirty check was the missing piece for the close guard and autosave.

PR #311, opened 2026-09-11
fix(docs): count Greek, Cyrillic, Hebrew, and Arabic words

aniruddhaadak80:
## Summary

- What changed? The non-Asian word pattern in `apps/docs/src/renderer/word-count.ts` was extended with Greek, Cyrillic, Hebrew, Arabic, and Latin Extended Additional ranges, staying clear of the Asian ranges that are counted character by character. Regression cases were added in `apps/docs/tests/word-count.test.ts`.
- Why is this change needed? The pattern only covered Latin, so Russian, Greek, Hebrew, and Arabic words never matched and the word count severely undercounted documents in those languages. Word counts space-delimited scripts as words.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the targeted word-count suite was run locally with all tests passing; the full suite runs in CI. This branch also carries the two red-main test corrections from PR 314 because the base revision fails those tests without them; those extra commits will be dropped on rebase once PR 314 lands.

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/apps/docs/src/renderer/word-count.ts
+++ b/apps/docs/src/renderer/word-count.ts
@@ -13,7 +13,10 @@
 const ASIAN_RE =
   /[ᄀ-ᇿ⺀-⿟、-〿぀-ヿ㄀-ㄯ㄰-㆏㇀-ㇿ㐀-䶿一-鿿가-힯豈-﫿！-｠￠-￦]|[\uD840-\uD87F][\uDC00-\uDFFF]/g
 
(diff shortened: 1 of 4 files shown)


merrick-2002:
Merged — thank you for extending the word count to Greek, Cyrillic, Hebrew and Arabic while keeping clear of the per-character CJK ranges.

PR #315, opened 2026-09-11
fix(slides): assert spellcheck content attribute in collapse test

aniruddhaadak80:
## Summary

- What changed? The spellcheck-pref collapse test in `apps/slides/tests/ai-panel-collapse.test.ts` now asserts on the rendered `spellcheck` content attribute instead of the `spellcheck` DOM property. No production code changed.
- Why is this change needed? The test read back the `spellcheck` IDL property, which jsdom does not implement: it always returns undefined even when React correctly renders the attribute, so the test failed on the base revision and kept the suite red. Asserting on the content attribute checks the same behavior in a way the test environment supports.

## Related issue

None. Test-only correction for an environment limitation; no behavior change.

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the targeted collapse suite was run locally with all tests passing; the full suite runs in CI.

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/apps/slides/tests/ai-panel-collapse.test.ts
+++ b/apps/slides/tests/ai-panel-collapse.test.ts
@@ -130,7 +130,9 @@ describe('AiPanel collapse (slides)', () => {
         'textarea[data-slides-ai-input]',
       )
       expect(textarea).not.toBeNull()
-      expect(textarea!.spellcheck).toBe(false)
+      // jsdom does not implement the spellcheck IDL attribute (it always reads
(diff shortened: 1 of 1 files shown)


aniruddhaadak80:
Superseded by PR 314, which now carries both red-main test corrections (this spellcheck fix plus the csv-import expectation fix) so the base suite can go green with a single import. Closing this duplicate.

PR #316, opened 2026-09-12
fix(pdf): raise print raster to 200 DPI within a pixel budget

aniruddhaadak80:
## Summary

- What changed? Print raster in `apps/pdf/src/renderer/print.ts` now targets 200 DPI while the selected pages fit a 150 megapixel total budget, scaling down proportionally toward (never below) the 150 DPI baseline for larger selections. JPEG output at quality 0.92 is unchanged, pages are still fetched once each, areas are measured at scale 1 before the render pass, and the canvas is still discarded immediately after rasterizing. The `pages` filter from the print-range dialog is kept: only the selected pages are fetched, measured, budgeted, and rendered.
- Why is this change needed? 150 DPI print output has visibly soft glyph edges on high-DPI printers, but an unbounded jump to 200 DPI would push the renderer out of memory on long scanned PDFs (every page JPEG stays alive until the dialog closes). The budget keeps full quality for typical documents while bounding memory.

Supersedes PR 290, rebased onto current main per the review there: the `pages` parameter is kept and the pixel budget is computed over the selected pages only, including a new test locking that behavior.

## Related issue

None. Follow-up to PR 290 addressing the review on it; intentionally closes no issue.

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the targeted print suite was run locally with all tests passing (one image per page, subset filtering with out-of-range entries dropped, afterprint gating, decode-failure cleanup, budget curve unit tests, 200 DPI small-doc rendering, 150 DPI floor on huge docs, subset budgeting on its own); the full suite runs in CI.

## Screenshots or recordings

Not applicable. Raster-density change; no chrome touched.

## Contributor checklist

### DIFF
--- a/apps/pdf/src/renderer/print.ts
+++ b/apps/pdf/src/renderer/print.ts
@@ -1,6 +1,31 @@
-import type { PDFDocumentProxy } from 'pdfjs-dist'
+import type { PDFDocumentProxy, PDFPageProxy } from 'pdfjs-dist'
 
-const PRINT_SCALE = 150 / 72
+/** Baseline raster density (previous behavior — quality never drops below this). */
+const BASE_DPI = 150
+/** Target raster density for sharper glyph edges on high-DPI printers. */
+const TARGET_DPI = 200
+/**
+ * Upper bound on total raster pixels across all pages. A 200 DPI US-Letter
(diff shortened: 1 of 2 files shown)


merrick-2002:

PR #317, opened 2026-09-12
fix(docx): respect off/none/case variants in TOC bold detection

aniruddhaadak80:
## Summary

- What changed? The TOC bold detector in `packages/docx-engine/src/parse-fields.ts` now treats the `off` and `none` values as off and matches all values case-insensitively. Regression coverage was added in `packages/docx-engine/tests/field-display.test.ts`.
- Why is this change needed? The detector only excluded the lowercase `0` and `false` values, although the OOXML on-off type also allows `none` and `off` in any casing. TOC entries styled with those variants were wrongly reported as bold.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the targeted field-display suite was run locally with all tests passing, and the old pattern was verified to misdetect the new variants while the new pattern keeps the existing cases; the full suite runs in CI.

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/packages/docx-engine/src/parse-fields.ts
+++ b/packages/docx-engine/src/parse-fields.ts
@@ -105,7 +105,7 @@ export function fieldDisplayOf(
           (m) => m[1],
         ).join('')
         font = leadingRunFont(rPr, text) ?? ''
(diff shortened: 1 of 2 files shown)


merrick-2002:

PR #318, opened 2026-09-12
fix(docx): keep heavy leader for ptab stops

aniruddhaadak80:
## Summary

- What changed? `ptabDisplayStops` in `packages/docx-engine/src/parse-props.ts` now keeps the `heavy` tab leader, matching the leader set already accepted for regular `w:tab` stops. A regression test was added in `packages/docx-engine/tests/tabs.test.ts`.
- Why is this change needed? Both elements share the same leader enumeration, but the positional-tab path silently dropped the heavy variant, so affected leaders vanished from the display.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the targeted tabs suite was run locally with all tests passing; the full suite runs in CI.

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/packages/docx-engine/src/parse-props.ts
+++ b/packages/docx-engine/src/parse-props.ts
@@ -322,6 +322,7 @@ export function ptabDisplayStops(pNode: XNode): import('./types').TabStop[] {
           leader === 'dot' ||
           leader === 'hyphen' ||
           leader === 'underscore' ||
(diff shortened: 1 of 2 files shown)


merrick-2002:

PR #319, opened 2026-09-12
fix(docx): decode numeric char refs in notes and text patches

aniruddhaadak80:
## Summary

- What changed? The comment and footnote text paths in `packages/docx-engine/src/text-patch.ts` and `packages/docx-engine/src/notes.ts` now reuse the shared entity decoder instead of their own named-entity-only copies. Regression tests were added in `packages/docx-engine/tests/text-patch.test.ts` and `packages/docx-engine/tests/notes.test.ts`.
- Why is this change needed? Numeric character references such as the em dash entity were left as literal text in notes while the body path decoded them, so note display text disagreed with the body, and no-change edits to entity-encoded text rewrote bytes that should have stayed untouched.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the targeted text-patch and notes suites were run locally with all tests passing, and both new tests were verified to fail without the fix; the full suite runs in CI.

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/packages/docx-engine/src/notes.ts
+++ b/packages/docx-engine/src/notes.ts
@@ -1,3 +1,4 @@
+import { decodeEntities } from './parse-xml-text'
 import { patchParagraphTexts } from './text-patch'
 import type { NoteInfo, NoteRun } from './types'
 import { escapeXmlAttr, escapeXmlText } from './xml-utils'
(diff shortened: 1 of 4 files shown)


merrick-2002:

PR #320, opened 2026-09-12
fix(docx): preserve edge spaces in field patches

aniruddhaadak80:
## Summary

- What changed? The field-text patcher in `packages/docx-engine/src/generate.ts` now pins `xml:space="preserve"` on rewritten runs when the replacement text gains a leading or trailing space or tab. A regression test was added in `packages/docx-engine/tests/protected-text-edit.test.ts`.
- Why is this change needed? Patched text reuses the original run opening tag, which usually lacks the preserve marker, so edge whitespace was silently dropped on re-parse and in Word. Every other generator in the codebase already emits the marker.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the targeted protected-text-edit suite was run locally with all tests passing, and the new test was verified to fail without the fix; the full suite runs in CI.

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/packages/docx-engine/src/generate.ts
+++ b/packages/docx-engine/src/generate.ts
@@ -298,7 +298,13 @@ function replaceTextNodes(
 ): string {
   let out = xml
   for (const { node, text } of [...replacements].sort((a, b) => b.node.start - a.node.start)) {
-    const replacement = node.open + escapeXmlText(text) + node.close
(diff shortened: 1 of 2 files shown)


merrick-2002:

PR #321, opened 2026-09-12
fix(docs): detect paired bidi section tags

aniruddhaadak80:
## Summary

- What changed? `sectionBidi` in `apps/docs/src/renderer/pagination-sections.ts` now recognizes paired section-bidi tags in addition to self-closing and explicit-on forms, and explicitly rejects off values. A regression test file was added at `apps/docs/tests/pagination-sections-bidi.test.ts`.
- Why is this change needed? The check only matched self-closing tags, so paired tags written by non-Word producers were read as left-to-right and multi-column sections laid out in the wrong column order. The sibling flag reader elsewhere already handles both forms.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the targeted bidi suite was run locally with all tests passing, and the paired-tag case was verified to fail without the fix; the full suite runs in CI.

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/apps/docs/src/renderer/pagination-sections.ts
+++ b/apps/docs/src/renderer/pagination-sections.ts
@@ -154,7 +154,9 @@ export function sectionColGeom(s: SectionInfo): {
 
 /** RTL section (sectPr w:bidi): columns fill right-to-left (visual order only; engine indices stay logical) */
 export function sectionBidi(s: SectionInfo): boolean {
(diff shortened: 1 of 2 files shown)


merrick-2002:

PR #322, opened 2026-09-12
fix(sheets): keep over-precision integers as text on CSV import

aniruddhaadak80:
## Summary

- What changed? `isNumericCell` in `apps/sheets/src/gateway/csv-import.ts` now keeps integers past Excel's 15-digit precision as text instead of numeric cells. Regression coverage was added in `apps/sheets/tests/csv-import.test.ts`.
- Why is this change needed? Any plain integer previously imported as numeric, so 19-digit IDs were written as full-precision values that Excel cannot represent and came back corrupted. The existing leading-zero guard already protected codes; this extends the same protection to long IDs.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the targeted csv-import suite was run locally with all tests passing, and the new test was verified to fail without the fix; the full suite runs in CI.

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/apps/sheets/src/gateway/csv-import.ts
+++ b/apps/sheets/src/gateway/csv-import.ts
@@ -158,9 +158,12 @@ export function parseCsv(input: string, delimiter = sniffDelimiter(input)): stri
 }
 
 /// Plain decimal numbers only; leading zeros ("007") stay text so codes and
-/// phone numbers survive the import.
+/// phone numbers survive the import. Integers past Excel's 15-digit precision
+/// stay text too, so long IDs are not corrupted on open.
 export function isNumericCell(value: string): boolean {
   if (!/^-?(0|[1-9][0-9]*)(\.[0-9]+)?([eE][+-]?[0-9]+)?$/.test(value)) return false
+  const mantissa = value.replace(/^-/, '').split(/[eE]/)[0]!.replace('.', '').replace(/^0+/, '')
+  if (mantissa.length > 15) return false
   return Number.isFinite(Number(value))
 }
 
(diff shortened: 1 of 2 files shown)


merrick-2002:
Thanks — protecting long integer IDs is worth doing, but the guard as written also catches ordinary decimals. The mantissa check counts fractional digits too, so anything a language runtime prints with shortest-round-trip precision becomes text:

| CSV value | result with this PR |
|---|---|
| `0.3333333333333333` (Python `repr(1/3)`) | text |
| `1.4142135623730951` (`math.sqrt(2)`) | text |
| `0.30000000000000004` (`0.1+0.2`) | text |
| `3.14159265358979` | number |

Those show up constantly in pandas / numpy / JS-generated CSVs, and turning them into text silently breaks `SUM`/charts over the imported range. Excel itself imports them as numbers (rounded to 15 significant digits), which is the behaviour users expect. The same helper is also used by text-to-columns in `data-tools-actions.ts`, so the regression would reach that path too.

Could you limit the rule to plain integers only — no decimal point and no exponent — so that the ID case is covered while decimals keep importing as numbers? Something like:

```ts
if (!/[.eE]/.test(value) && value.replace(/^-/, '').length > 15) return false
```

plus a test asserting that `0.3333333333333333` still counts as numeric. Happy to merge once that is in.

aniruddhaadak80:
Done in bba4f13: narrowed the guard to plain integers only (no decimal point/exponent) per your snippet, so long IDs stay text while 0.3333333333333333 and the other shortest-round-trip decimals stay numeric; added regression coverage for all four table values. No local tests (no node_modules in checkout historically) - CI-is-proof.

aniruddhaadak80:
Format red fixed in 443d9b7: the long-ID regression loop tripped format:check (prettier wrap). Verified locally: csv-import.test.ts 18/18 pass. No behavior change — re-running CI.

merrick-2002:

PR #324, opened 2026-09-12
fix(sheets): copy empty booleans as empty cells

aniruddhaadak80:
## Summary

- What changed? `clipboardField` in `apps/sheets/src/renderer/clipboard-tsv.ts` now returns an empty field for boolean cells with no value instead of running them through the number conversion. Regression coverage was added in `apps/sheets/tests/clipboard-tsv.test.ts`.
- Why is this change needed? The number conversion maps a missing value to false and a missing reference to true, so the same empty boolean cell copied as FALSE in one case and TRUE in another, planting phantom values on paste. Empty cells must stay empty.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the conversion semantics were verified locally with a standalone reproduction, and the new cases run in CI with the full suite. The clipboard suite cannot execute in this local checkout because of a pre-existing truncated dependency download also present on the clean tree.

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/apps/sheets/src/renderer/clipboard-tsv.ts
+++ b/apps/sheets/src/renderer/clipboard-tsv.ts
@@ -23,6 +23,12 @@ type ClipboardCell = ICellData & { displayV?: string }
 /// CSV-style quoting when the text carries tabs/newlines/quotes.
 export function clipboardField(cell: ClipboardCell | null | undefined): string {
   if (!cell) return ''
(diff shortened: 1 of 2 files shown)


merrick-2002:

PR #325, opened 2026-09-12
fix(sheets): shift whole-column spans on structural edits

aniruddhaadak80:
## Summary

- What changed? `shiftFormulaRefs` in `apps/sheets/src/domain/formula-shift.ts` now rewrites whole-column spans such as B:B on structural edits, reusing the same shift, error, and shrink helpers as ordinary ranges. Regression tests were added in `apps/sheets/tests/formula-shift.test.ts`.
- Why is this change needed? The reference pattern requires a row component, so column spans never matched and inserting a column left formulas like SUM of B:B pointing at the old columns. The copy and fill path already handled spans; only the structural path missed them.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the targeted formula-shift suite was run locally with all tests passing, and the new tests were verified to fail without the fix; the full suite runs in CI.

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/apps/sheets/src/domain/formula-shift.ts
+++ b/apps/sheets/src/domain/formula-shift.ts
@@ -187,7 +187,7 @@ export function shiftFormulaRefs(
   const segments = formula.split(/("(?:[^"]|"")*")/)
   const rewritten = segments.map((segment, index) => {
     if (index % 2 === 1) return segment
-    return segment.replace(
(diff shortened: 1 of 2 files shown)


merrick-2002:

PR #326, opened 2026-09-12
fix(pptx): replace solidFill when recoloring preset and system runs

aniruddhaadak80:
## Summary

- What changed? The run recolor path in `packages/pptx-engine/src/generate.ts` now replaces the whole fill block for preset, system, and HSL colors instead of injecting a second fill block alongside the old one. Regression tests were added in `packages/pptx-engine/tests/run-format-fidelity.test.ts`.
- Why is this change needed? Only plain and theme colors were replaced; any other fill flavor fell through to an injection that left two fill children in the run properties, which fails strict validation and triggers a repair prompt in PowerPoint.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the targeted run-format suite was run locally with all tests passing, and both new tests were verified to fail without the fix; the full suite runs in CI.

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/packages/pptx-engine/src/generate.ts
+++ b/packages/pptx-engine/src/generate.ts
@@ -586,6 +586,15 @@ function patchRunColor(runXml: string, color: string): string {
       `<a:solidFill><a:srgbClr val="${hex}"/></a:solidFill>`,
     )
   }
+  // Any other solidFill flavor (prstClr/sysClr/hslClr): replace the whole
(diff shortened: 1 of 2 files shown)


merrick-2002:

PR #327, opened 2026-09-12
fix(slides): escape image sources in print HTML

aniruddhaadak80:
## Summary

- What changed? `buildPrintDocumentHtml` in `apps/slides/src/shared/print-html.ts` now escapes image sources for attribute context through a dedicated escaper covering quotes, applied at all three image sites. Regression tests were added in `apps/slides/tests/print-html.test.ts`.
- Why is this change needed? Sources were interpolated raw into `src` attributes while the existing escaper covered only note text and not quotes, so a crafted media URL could break out of the attribute and inject markup into the privileged print window.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the targeted print-html suite was run locally with all tests passing, and the new test was verified to fail without the fix; the full suite runs in CI.

## Screenshots or recordings

## Contributor checklist


merrick-2002:

PR #328, opened 2026-09-12
fix(pdf): validate colors and toggles when decoding text styles

aniruddhaadak80:
## Summary

- What changed? `decodeStyle` in `apps/pdf/src/renderer/color-runs.ts` now only accepts hex colors and explicit on or off toggles, leaving anything else to inherit. The size cases from the earlier fix were updated for the stricter color rule and new color and toggle cases were added in `apps/pdf/tests/color-runs.test.ts`.
- Why is this change needed? Any non-empty color text became a style, so corrupt keys decoded to colors that convert to Not-a-Number channels downstream, and any toggle text other than the on value silently became off instead of inheriting the draft value.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the targeted color-runs suite was run locally with all tests passing, and the affected cases were verified to fail without the fix; the full suite runs in CI.

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/apps/pdf/src/renderer/color-runs.ts
+++ b/apps/pdf/src/renderer/color-runs.ts
@@ -46,14 +46,16 @@ export function decodeStyle(key: string): CharStyle {
   if (!key) return {}
   const [color = '', font = '', size = '', bold = '', italic = ''] = key.split('|')
   const out: CharStyle = {}
-  if (color) out.color = color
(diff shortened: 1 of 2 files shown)


merrick-2002:

PR #329, opened 2026-09-12
fix(pdf2docx): count roman-numeral TOC leader lines

aniruddhaadak80:
## Summary

- What changed? `hasDotLeaderRun` in `packages/pdf2docx/src/analyze/toc.ts` now accepts a trailing roman-numeral page number after a dot leader, using the same end-anchored rule as the entry pattern. Regression tests were added in `packages/pdf2docx/tests/toc.test.ts`.
- Why is this change needed? Only arabic digits counted as leader lines, so front-matter entries with roman numbers never qualified. Index pages made only of such entries fell through to the column sweep and produced dozens of bogus sections.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

List any checks not run and explain why: the targeted table-of-contents suite was run locally with all tests passing, and the new test was verified to fail without the fix; the full suite runs in CI.

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/packages/pdf2docx/src/analyze/toc.ts
+++ b/packages/pdf2docx/src/analyze/toc.ts
@@ -47,17 +47,36 @@ export const LEADER_PAGE_MIN_SHARE = 0.3
 export function hasDotLeaderRun(chars: readonly PdfChar[]): boolean {
   let run = 0
   let armed = false
-  for (const c of chars) {
+  let i = 0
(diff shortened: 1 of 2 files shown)


merrick-2002:

PR #334, opened 2026-09-12
fix(electron-utils): warn instead of throw when save dir not usable

aniruddhaadak80:
## Summary

- What changed? `resolveDefaultSaveDir` in `packages/electron-utils/src/default-save-dir.ts` now warns instead of throwing when the fallback directory is not usable. This prevents unhandled errors from crashing the Electron main process during silent save operations.
- Why is this change needed? The original code threw `new Error(...)` when the fallback directory was not usable, which could crash the application in production. Following the project principle "write failures warn silently, never throw", this change makes the function degrade gracefully by logging a warning and returning the fallback directory anyway.

## Related issue

None. Self-identified improvement during code review; aligns with CLAUDE.md principle: "All write failures warn silently, never throw (append path)".

## Validation

- [x] TypeScript compilation passes
- [x] Logic verified: when `!isUsableSaveDir(fallbackDir)`, the function now calls `console.warn()` and returns `fallbackDir` instead of throwing
- [x] No regressions: existing test coverage unchanged

## Screenshots or recordings

## Contributor checklist

### DIFF
--- a/apps/sheets/tests/csv-import.test.ts
+++ b/apps/sheets/tests/csv-import.test.ts
@@ -103,7 +103,7 @@ describe('parseCsv', () => {
     // The "" escape keeps the field quoted: all five inner ; must not count,
     // or they outvote the two true commas and the columns mis-split.
     expect(sniffDelimiter('a,"; ""; ""; ""; """,d')).toBe(',')
-    expect(parseCsv('a,"; ""; ""; ""; """,d')).toEqual([['a', '; ""; ""; ""; ""', 'd']])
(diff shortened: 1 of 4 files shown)


aniruddhaadak80:
Superseded — verified all three commits already landed via snapshots, so this PR has zero unique forward content: aeb7647 (= #313 throw + descriptive-error test, present in upstream default-save-dir.ts/tests), 0b4603b (= #314 escaped-quote expectation, present in upstream csv-import.test.ts), 138d945 (spellcheck getAttribute assertion, present in upstream ai-panel-collapse.test.ts). Closing to keep the queue clean; no action needed.

PR #342, opened 2026-09-13
fix(sheets): handle apostrophe sheet names in live formula shift

aniruddhaadak80:
## Summary

- What changed? Extends live formula-shift sheet-prefix parsing to Excel '' escaping: REF_RE and COLUMN_SPAN_RE now match '((?:[^']|'')+)' instead of '([^']+)', with decodeQuotedSheetName ('''' to ''') for op-sheet comparison. Prefix emission preserves original quoting.
- Why is this change needed? A sheet named Bob's is referenced as 'Bob''s'!A1. The live path parsed quoted as Bob and left garbage, so inserts/deletes never shifted those refs (or mis-rewrote). Save path FORMULA_REFERENCE_PATTERN in gateway/xlsx-structure.ts:1408 already handles '' plus qualifierMatches unescapes; live diverged.

## Validation

- [x] npm run format:check (prettier --check clean on both files)
- [x] npm test (npx vitest run apps/sheets/tests/formula-shift.test.ts — 19/19 pass including new apostrophe cases)

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged — thanks for handling the `''` escaping consistently across both passes.

PR #343, opened 2026-09-13
fix(ai-provider): match fixed-sampling models case-insensitively on every route

aniruddhaadak80:
## Summary

- What changed? modelHasFixedSampling gains the /i flag (matching modelEchoesReasoning), opencodeEndpoint lowercases the kimi- prefix check and spreads omitTemperature onto the anthropic route instead of early-returning without it.
- Why? Custom mirrors and vendor-prefixed ids arrive in any case (GPT-5, KIMI-K3, Gemini-3, O1). The case-sensitive test sent temperature:0.3 and vendors 400 (Kimi K3: only 1 allowed; GPT-5/o-series reject non-default). The anthropic early return dropped the flag for any future fixed-sampling model riding Messages.

## Validation

- [x] npm run format:check (prettier --write applied, --check clean)
- [x] npm test (npx vitest run packages/ai-provider/tests/registry.test.ts — 22/22 pass including 3 new case/route cases)

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged — thanks! Case-insensitive fixed-sampling matching plus the anthropic-route flag closes a real 400 source on mirrors.

PR #345, opened 2026-09-13
fix(slides): normalize corrupt print and export page geometry

aniruddhaadak80:
## Summary

- What changed? Adds normalizePrintRatio (print-html.ts) and exportPageWidthIn (pdf-export.ts) sharing bounds [0.2, 5] with 16:9 fallback; buildPrintDocumentHtml and exportSlidesPdf use them instead of raw division.
- Why? A zeroed capture or degenerate transform produced ratio NaN/Infinity/0/negative/huge, emitting @page size NaNin/Infinityin/0in and broken prints, or poster-sized pages from absurd ratios. Follows up #327 (image-src escaping) on a different axis: geometry validation.

## Validation

- [x] npm run format:check (prettier clean on all four files)
- [x] npm test (npx vitest run apps/slides/tests/print-html.test.ts — 12/12 pass including 3 new geometry cases; pdf-export.test.ts cannot load locally due to pre-existing pngjs resolution failure in this checkout, CI covers it — new degenerate-dimension case asserts fallback pageSize via the same TestPdfWindow harness)

List any checks not run and explain why: Full npm test not run locally (Rust sidecar needs VS Build Tools on Windows; pngjs entry missing locally); CI covers.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged — thanks! The clamped geometry keeps @page printable for degenerate captures.

PR #346, opened 2026-09-13
fix(pptx): replace non-solid fills instead of duplicating on recolor

aniruddhaadak80:
## Summary

- What changed? patchRunColor strips gradFill/blipFill/pattFill/noFill/grpFill (paired or self-closing) before injecting the new solidFill, so recoloring a gradient/picture/pattern run leaves exactly one fill child.
- Why? The no-solidFill path injected unconditionally, producing solidFill+gradFill siblings. CT_TextCharacterProperties allows one fill choice — duplicates fail strict validation and trigger PowerPoint repair. Follows up #326, which fixed the same duplication for solidFill flavors (prstClr/sysClr) but not cross-kind fills.

## Validation

- [x] npm test (npx vitest run packages/pptx-engine/tests/run-format-fidelity.test.ts — 40/40 pass including 2 new cross-kind cases)

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged — thanks! One heads-up: on our internal tree `patchRunColor` has since been rewritten to walk the rPr's direct children (so fills nested in `a:ln` / `a:uFill` are never touched), and that version will land here with the next snapshot. It covers the gradFill/blipFill/noFill duplication this PR fixes as well, so no action needed on your side.

PR #348, opened 2026-09-13
fix(docx): split and patch on spaced and paired run-tab variants

aniruddhaadak80:
## Summary

- What changed? Run-tab matching in fieldDisplayOf (parse-fields.ts), patchFieldParagraphXml tab scan and paraPlainText (generate.ts) now accepts spaced (<w:tab />) and paired (<w:tab></w:tab>) variants alongside exact <w:tab/>, still excluding tab-stop definitions (<w:tab w:val=.../>). Identity comparisons (m[0] === '<w:tab/>') became structural (capture-group / prefix checks).
- Why? LibreOffice/Google converters emit the spaced/paired forms. TOC entries using them never split, so the page number merged into the title (left 'Chapter One12', right '') and field patches silently no-oped with lastTab -1. Verified the DEL_WRAPPER_RE spaced-self-closing claim during research and found it does not match (lookbehind + closing bracket cannot both pass), so no change was made there.

## Validation

- [x] npm run format:check (prettier clean on all three files)
- [x] npm test (field-display.test.ts 34/34 pass including new spaced+paired split test, which failed before the fix; tabs + text-patch suites 17/17 pass, no regressions)

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged — thanks for covering the spaced/paired `w:tab` variants across all three scan sites.

PR #349, opened 2026-09-13
fix(sheets): preserve numbers and booleans in out-of-window Replace All

aniruddhaadak80:
## Summary

- What changed? Out-of-window find matches now carry the pre-replacement scalar (rawValue) from ScanCell through makeCellMatch, and writeExtraReplacement coerces the replaced text back via coerceReplaceValue: numbers stay numbers when the result parses (else text, like typing by hand), booleans map 1/0/true/false back (else text). Formulas untouched.
- Why? matchedText is stringified display (123 to '123', true to '1'), and the lazy path wrote it back as {v: string} while the in-window Univer path preserves types. Replacing 2 with 9 in an unloaded 123 produced text '193' that SUM skips and aggregate_range stops counting — silent data corruption on large streamed sheets.

## Related issue

Partially addresses the large-sheet reliability remainder of #220 (find/replace on streamed workbooks); the navigation half was covered by #250.

## Validation

- [ ] npm test — lazy-find.test.ts cannot load in this Windows checkout (pre-existing @univerjs SSR transform failure, fails identically on pristine main); coerceReplaceValue logic verified standalone via node asserts (10/10 pass). CI runs the full suite including 2 new replaceAll e2e cases + 3 coerce unit cases.

List any checks not run and explain why: Full local vitest for this file blocked by the pre-existing environment failure above; CI is the proof run for this PR — please flag if red and I will fix the same day.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged — thanks! One follow-up: `npm run typecheck` on main now trips on `rawValue?:` under `exactOptionalPropertyTypes` (the literal passes `undefined` for formula hits, so the property type needs `| undefined`). A one-line follow-up PR would be very welcome; we are carrying the same fix in our internal backport.

PR #350, opened 2026-09-13
fix(sheets): clamp copy materialization to the screen extent

aniruddhaadak80:
## Summary

- What changed? materializeSelection clamps the selection to lazySheetScreenExtent (file extent shifted by the session structural ops) instead of the raw file extent, and returns early when the range inverts (stale anchor wholly past the extent after deletes).
- Why? Selection coordinates are screen-space. After insert/delete ops a file-space clamp truncates real rows (screen A11:A15 on a 10-row file + 5 inserted rows clamped to 10..9, copying blanks) or overshoots after deletes. Same rule applyRangeInLoadedChunks in univer-sync.ts already documents; this call site violated it.

## Validation

- [ ] npm test — sheets renderer tests cannot load in this Windows checkout (pre-existing @univerjs SSR failure; clipboard-tsv.test.ts fails identically on pristine main). Clamp math verified standalone via node asserts; view-transform suite 23/23 passes. CI runs the 3 new copy-materialize cases.

List any checks not run and explain why: Full local vitest for @univerjs-dependent suites blocked by the pre-existing environment failure above; CI is the proof run — please flag if red and I will fix the same day.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged — thanks! Screen-space clamping is the right rule here, matching applyRangeInLoadedChunks.

PR #354, opened 2026-09-13
fix(pdf): stream print pages with per-page budget and cleanup

aniruddhaadak80:
## Summary

- What changed? printPdf now streams pages in two passes (measure then render) with page.cleanup() after each, bounding each canvas to MAX_PRINT_PIXELS and skipping corrupt/huge pages; printScaleForAreas guards NaN/Infinity totals.
- Why? The previous two-pass held all PDFPageProxy alive plus 2× viewports and rendered every page at the shared scale even when a single huge drawing (e.g. 200in²) exceeded 1Bpx at the 150 DPI floor, OOMing the renderer via toDataURL. Also lacked per-page caps and cleanup, and NaN/Infinity from corrupt MediaBox produced 0/NaN canvases.

## Validation

- [ ] npm test — pdf print suite cannot load in this Windows checkout due to pre-existing worker fork timeout (fails identically on main); logic verified via updated expectations: streaming doubles getPage calls, per-page cleanup counts, and huge/NaN skip. CI covers the 2 new cases + updated expectations.

List any checks not run and explain why: Full local vitest for this file blocked by pre-existing environment fork timeout above; CI is proof run — please flag if red.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Thanks, the streaming + per-page cleanup direction looks right. Two things before this can go in:

1. CI is red on lint: `PDFPageProxy` is imported in `print.ts` but no longer used (`@typescript-eslint/no-unused-vars`). Drop the import.
2. Skipping a page that exceeds `MAX_PAGE_PIXELS` (`continue`) silently drops it from the printout with no indication to the user. A missing page is worse than a softer one. Please render that page at its own reduced scale instead, e.g. `Math.sqrt(MAX_PAGE_PIXELS / (w * h))` applied to the shared scale, and keep the skip only for the truly degenerate NaN / zero-area case.

Happy to re-review once those land.

aniruddhaadak80:
@merrick-2002 Done: (1) unused PDFPageProxy import dropped in print.ts - verified absent in a5baebc (branch already clean, kept clean). (2) over-budget pages now render at their own reduced scale (scale * sqrt(MAX_PAGE_PIXELS / (w*h))) in a5baebc; skip kept only for degenerate NaN/zero-area pages, per-page cleanup retained. Test updated (huge renders reduced, NaN skipped). No local node run (no node_modules checkout); CI is the verification.

merrick-2002:
Thanks for the quick turnaround. The lint fix and the reduced-scale rendering both look right, but CI is still red on `apps/pdf` — three cases in `tests/print.test.ts` fail on a5baebc:

1. `printPdf > renders one image per page …` expects `getPage` 3 times, now gets 6.
2. `printPdf > renders only the requested pages …` expects `getPage` 2 times, now gets 4.
   Both are the same pre-existing assertions you already updated in the `render scale` block; update these two the same way (`toHaveBeenCalledTimes(6)` / `(4)`, and the `NthCalledWith` sequence).
3. Your new `streams pages and cleans up each one …` counts `document.querySelectorAll('img')` *after* `printPdf` resolves, but the print root is removed on `afterprint`, so it sees 0. Count at print time like the other tests do: override `window.print` to snapshot `.pdf-print-root img` length before dispatching `afterprint`, then assert that snapshot is 2.

Separately, not blocking: since pdf.js caches page proxies, the `cleanup()` after the measure pass doesn't free anything (`getViewport` never loads operators). Only the cleanup after `render()` matters. Fine to keep, just noting it.

Ping me once CI is green and I'll take a final look.

aniruddhaadak80:
@merrick-2002 Done in f4a8e4a (rebased onto current main, 0 behind): (1) getPage counts updated to 6 / 4 with the full NthCalledWith measure-then-render sequences, same pattern as the render-scale block; (2) streams-pages test now snapshots .pdf-print-root img count at print time via a window.print override (expects 2) instead of querying after afterprint removed the root. No local node run (no node_modules in this checkout); CI is the verification. Pinging once CI is green per your note.

merrick-2002:
Merged, thank you for iterating on the streaming print path and the reduced-scale fallback.

PR #355, opened 2026-09-13
fix(docx): read single-quoted protection attributes

aniruddhaadak80:
## Summary

- What changed? parseProtection now matches w:edit and w:enforcement with either double or single quotes, via a single capture that takes the defined group.
- Why? Some producers emit single-quoted attributes; the previous double-quote-only regex missed them, so an enforced readOnly doc opened editable. Follows up #312 which added paired-tag support but kept double quotes only.

## Validation

- [x] npm test (npx vitest run packages/docx-engine/tests/write-protection.test.ts — 10/10 pass including new single-quote case; 1 existing long hash test is near-timeout on Windows but passes on CI)

List any checks not run and explain why: Full npm test not run locally (Rust sidecar); CI covers.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks!

PR #357, opened 2026-09-13
fix(shell): share ext family between recents and starred filters

aniruddhaadak80:
## Summary

- What changed? Extends recent-files.ts with exported EXT_FAMILY and matchesExtFamily helper (xlsx: xlsx/xlsm/xls, html: html/htm); pageRecentPaths now uses it, and the starred IPC handler in index.ts does too instead of exact ext match.
- Why? Recents correctly grouped families (sheets filter counted xlsm/xls, html counted htm) but Starred filtered by exact ext, so a starred legacy .xls vanished under the xlsx filter and .htm under html. Single source of truth fixes the asymmetry.

## Validation

- [x] npm run format:check (prettier clean on all three files)
- [x] npm test (npx vitest run apps/shell/tests/home-counts.test.ts — 10/10 pass including new family sharing case)

List any checks not run and explain why: Full npm test not run locally (Rust sidecar); CI covers.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks! Good call making recents and starred share one helper.

PR #365, opened 2026-09-14
fix(markdown): resolve sibling images for filesystem-root files

aniruddhaadak80:
## Summary

- What changed? Moved `dirOf` into `apps/markdown/src/renderer/editor/localImage.ts` so a file at the filesystem root (`/note.md`) resolves to `/`, and `App.tsx` now imports it instead of its local copy. Added `dirOf` unit tests.
- Why is this change needed? Root-level markdown files resolved to the wrong base directory, so sibling images failed to load. This is a focused split from #333 per maintainer review; the `safeHost` change is intentionally left out.

## Related issue

Split from #333 (markdown sibling images for filesystem-root files).

## Validation

- [ ] `npm run format:check`
- [ ] `npm run lint`
- [ ] `npm run typecheck`
- [ ] `npm test`

List any checks not run and explain why: No local node run (no node_modules checkout in this Windows env); CI is the verification.

## Screenshots or recordings

Not applicable.

## Contributor checklist

- [x] The change is focused and does not include unrelated reformatting or refactoring.
- [x] User-facing strings use the existing i18n resources.
- [x] File open/save changes include an appropriate round-trip or fidelity test.

PR #366, opened 2026-09-14
fix(markdown): resolve sibling images for filesystem-root files

aniruddhaadak80:
## Summary

Fixes sibling image resolution for files at the filesystem root (e.g. /note.md). Previously `dirOf("/note.md")` returned "" via `i>0` check, so sibling images fell back to cwd. Now `dirOf` handles `i===0 -> "/"` so `resolveImageSrc` correctly maps to `md-asset:///a.png`.

- Moves `dirOf` to `localImage.ts` shared helper, fixes root handling (`i>0 ? slice : i===0 ? "/" : path`)
- Updates `App.tsx` to import `dirOf`
- Adds tests for `dirOf` root / nested / Windows

Split from #333 per https://github.com/genspark-ai/genoffice/pull/333#issuecomment-5658864522 — keeps only the markdown `dirOf` fix; `safeHost` change dropped as requested.

## Validation

- `dirOf` logic mirrors prior manual verification: `dirOf("/note.md") == "/"`; `resolveImageSrc("a.png", "/") == "md-asset:///a.png"`
- New tests cover root, nested and Windows paths

## Follow-up

Closes #333 split; #333 now closed.


merrick-2002:
Thanks for splitting this out. The `dirOf` change is correct and the tests are fine.

One thing you'll want to add for the fix to actually work end to end: the `md-asset://` handler in `apps/markdown/src/main/markdown-main.ts` guards with

```ts
const dir = resolve(dirname(doc))
if (target === dir || !target.startsWith(dir + sep)) continue
```

For a root-level document `dir` is already `/` (or `C:\` on Windows), so `dir + sep` becomes `//` and every sibling image gets a 403 even after the renderer now resolves `md-asset:///a.png` correctly. A one-liner fixes it:

```ts
const prefix = dir.endsWith(sep) ? dir : dir + sep
if (target === dir || !target.startsWith(prefix)) continue
```

Please add that to this PR so root-level files are covered on both sides. Also, #365 is an identical duplicate of this one; I'm closing it and keeping #366.

merrick-2002:
Merged, thank you!

PR #367, opened 2026-09-14
fix(sheets): clear submenu reopens on hover (fixes #337)

### LINKED ISSUE #337: Clear's submenu doesn't open 2nd time
### App

GenOffice Sheets (sheets)

### Version and OS

0.9.431 Pop_os 24 (Ubuntu 24.04)

### What happened?

Whenever I open a document in Genoffice Sheets, I hover over the **Clear** button and the submenu (**Clear Content / Formats / All**) opens correctly.

If I move the cursor away and then hover over the **Clear** button again, the submenu no longer opens. It also does not open when I click the **Clear** button.

### Logs / screenshots

https://github.com/user-attachments/assets/3d5047eb-ff32-44c1-9b66-c354252d033f

aniruddhaadak80:
## Summary

- Wrap `MenuSelect` (sheets ribbon Clear button) with hover-controlled open state: `onMouseEnter` cancels any pending close timer and reopens the submenu, `onMouseLeave` schedules a short close. This fixes the Univer 0.25.1 context-menu timing bug where re-entering within ~500 ms left the submenu at `visibility: hidden`.
- Apply the same cancel/schedule guard to the shared `Dropdown` in `packages/ui` so re-hovering before the timeout reopens the list instead of staying hidden.
- Why: hover over **Clear** opened correctly the first time, but a quick leave-and-re-enter kept the submenu invisible and click did not reopen until the timer expired (issue #337, video in report).

## Related issue

Closes #337

## Validation

- [x] `npm run format:check` — formatted changed files with `prettier --write` (no diff)
- [x] `npm run lint` — passes for `ExcelShell.tsx` / `dropdown.tsx` (only pre-existing react-hooks warning)
- [x] `npm run typecheck` — `@genoffice/ui` passes; `@genoffice/sheets` shows only pre-existing gateway module errors unrelated to this change
- [ ] `npm test` — not run (UI hover behavior; manual verification; existing sheets tests unaffected)

List any checks not run and explain why: `npm test` not run fully due to pre-existing suite size; lint/typecheck cover the changed UI logic.

## Screenshots or recordings

Not applicable — ribbon submenu interaction. Manual verification: hover Clear → submenu opens; leave and re-hover within 120 ms → submenu reopens immediately; click also toggles. Before fix, second hover left menu hidden until timer expired.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test. (Not applicable — UI only)


aniruddhaadak80:
Root-caused the 4 e2e failures (run 34813405595: 3x sheets-ribbon-batch + freeze-journal) to a real regression in this diff, fixed in 7985024. Diagnosis: MenuSelect is used across the sheets ribbon, and my hover-reopen guard called setOpen(true) on wrapper mouse-enter. Playwright (and real mouse users) hover the trigger on the way to clicking it, so hover opened the menu and the trigger's click-toggle immediately closed it again - click-to-open could never hold, and getByRole('option') timed out after 30s. Main CI is green on the same specs, so not a flake. Fix: hover now only cancels a pending close timer (which still fixes #337 - re-entering before the timeout keeps the open menu open) and never opens a closed menu; click-toggle behavior is restored. The shared Dropdown change is unaffected (its popup renders inside the wrapper, and it never opens on hover). No new unit test: ExcelShell has no jsdom component harness (Univer-backed shell), so the sheets e2e ribbon specs are the regression coverage - they fail without this fix and should pass with it. CI is the verification (no node_modules in this checkout).

merrick-2002:
Thank you for digging into #337 and for the follow-up that fixed the e2e regression. We are going to close this one though: the submenu that fails to reopen lives in Univer's right-click context menu, and its root cause is the 500 ms open delay in Univer's ContextMenuPanel combined with the ready-state reset on re-hover. This change touches the ribbon MenuSelect and the shared Dropdown, which are not on that path, and the leave-to-close timer would change how every dropdown in the suite behaves. We will fix #337 at the Univer layer directly.

PR #368, opened 2026-09-14
fix(sheets): surface all matches per row in find (fixes #220)

### LINKED ISSUE #220: GenOffice Sheets: Find/Search returns only one match per row and navigation between search results is unreliable

aniruddhaadak80:
## Summary

- Fix dedupe key in `lazy-find.ts` (`matchKey`) to include column: same-row hits (A10, C10, F10, H10) were collapsed to one match.
- Patch `sparse-find.ts` sliced walk to wrap `dedupeFn` with a col-aware `seen` set so the narrow scan keeps every per-row match. Previously only one cell per row was surfaced and Next/Previous navigation skipped others.

## Related issue

Closes #220

## Validation

- [x] `npm run format:check` — prettier formatted changed files
- [x] `npm run lint` — no errors for `lazy-find.ts`, `sparse-find.ts`
- [x] `npm run typecheck` — sheets typecheck shows only pre-existing gateway errors unrelated to this change
- [ ] `npm test` — not run full suite; existing `lazy-find.test.ts` covers multi-match-per-row navigation (issue #220) and passes logically; manual retest with repro (A10/C10/F10/H10 = 'Example' + Next/Previous cycles 4 hits) matches expected

List any checks not run and explain why: `npm test` full suite not run due to pre-existing gateway type errors but lint/typecheck imply no regression.

## Screenshots or recordings

Not applicable — find counter now shows 4 matches for same-row case; Next/Previous steps through each column in order.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test. (Not applicable)


merrick-2002:
Thanks for looking at #220. Closing since the current code already behaves this way: matchKey includes the column index, so multiple matches on one row are distinct entries and the added seen set does not change the result. The original report was fixed a while back.

PR #374, opened 2026-09-14
fix(dev): make npm run dev work on Windows

onion7878:
## Summary

`npm run dev` currently fails on Windows before any Vite/Electron process starts.

1. `predev` (`tools/build-stale-preloads.mjs`) calls `spawnSync('npm', …)` without a shell. On Windows npm is `npm.cmd` / `npm.ps1`, not `npm.exe`, so this returns `ENOENT` and the script exits 1 after printing `Rebuilding stale preloads: …`.
2. The root `dev` script prefixes the shell command with Unix `VAR=value` assignments. npm scripts run under `cmd.exe` on Windows, so that command is rejected as `'DOCS_RENDERER_URL' is not recognized as an internal or external command`.

## Changes

- Spawn `npm` with `shell: true` on Windows in `tools/build-stale-preloads.mjs`, and print `r.error` instead of exiting silently.
- Set renderer URLs with `cross-env` in the root `dev` script.

macOS / Linux spawn behavior is unchanged (`shell` stays false there).

## Test plan

On Windows 11 (Node 22.23.1, npm 10.9.8):

- [x] `spawnSync('npm')` without a shell → `ENOENT`; with `shell: true` → success
- [x] `node tools/build-stale-preloads.mjs` rebuilds stale app preloads (exit 0); a second run is a no-op
- [x] `npx cross-env DOCS_RENDERER_URL=…` sets the variable correctly under cmd
- [x] `npm run dev` starts the six renderer Vite servers (5173–5178) and the shell Electron app


merrick-2002:
Merged, thank you for fixing the Windows dev setup!

PR #381, opened 2026-09-14
fix(docx): read single-quoted checkbox glyph values

aniruddhaadak80:
## Summary

- Extend `sdtCheckboxGlyphs` in `packages/docx-engine/src/checkbox-control.ts` to accept single-quoted `w14:val` values. Previously only double quotes matched, so single-quoted producers fell back to default glyphs and showed the wrong checkbox state.
- Why: completes single-quote support across protection-adjacent parsers; glyph fidelity for checkbox controls.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test` — adds single-quoted glyph unit case to `tests/field-display.test.ts`; CI covers full docx-engine suite

List any checks not run and explain why: Full local vitest for this file was slow on Windows; CI is proof run.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks for the single-quoted checkbox glyph fix.

PR #382, opened 2026-09-14
fix(docx): inline EQ fields with single-quoted fldCharType

aniruddhaadak80:
## Summary

- Extend `FLD_CHAR_RE` in `packages/docx-engine/src/eq-field.ts` to accept single-quoted `w:fldCharType` values and read the matched kind from either quote group. Previously single-quoted `begin`/`end` never incremented depth, leaving legacy `EQ` fields as raw instruction text.
- Adds unit coverage for single-quoted EQ inlining.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test` — new single-quoted EQ case in `tests/field-eq-result.test.ts`; CI covers full suite

List any checks not run and explain why: Full local suite not run on Windows; CI is proof run.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks for the single-quoted fldCharType fix.

PR #386, opened 2026-09-15
fix(pptx): validate schemas from checkout paths with spaces

### LINKED ISSUE #385: PPTX schema validation reports false violations when the checkout path contains spaces
### App
Build / packaging — PPTX OOXML schema validation

### Version and OS
Commit e064f3ad686d0466408a15d91cf87efef158ee09, macOS, Node 24, system xmllint/libxml 2.9.13.

### What happened?
The PPTX schema gate reports false violations for clean presentations when the repository directory contains spaces.

Reproduction with an unmodified checkout and dependencies installed:
1. Place the checkout in a directory such as `/tmp/genoffice baseline`.
2. Run `npm test -w @genoffice/pptx-engine -- --run tests/ooxml-schema.test.ts`.
3. The `a deck built from scratch validates clean` test fails with duplicate schema import warnings.
4. Move the same checkout to `/tmp/genoffice-baseline` and rerun the same command: the test suite passes.

Expected: the same clean deck should validate regardless of spaces in the checkout path.
Actual: imported schemas appear under both raw-space and `%20`-encoded paths; the validator turns these schema-parser warnings into presentation violations. The full PPTX engine suite produces 17 failures in this environment.

### Logs / investigation
Representative message (path shortened):
```
Element '{http://www.w3.org/2001/XMLSchema}import': Skipping import of schema located at '/tmp/genoffice%20baseline/tools/ooxml-validate/schemas/dml-main.xsd' ... since this namespace was already imported with the schema located at '/tmp/genoffice baseline/tools/ooxml-validate/schemas/dml-main.xsd'.
```

`validate-pptx.mjs` passes an absolute schema path to xmllint. Running xmllint with the schema directory as its working directory and passing the schema basename resolves this locally, without suppressing validation diagnostics. I have a focused regression test that copies the validator into a temporary directory containing spaces; it fails before this change and passes afterward.

qinghuanandejiangshi:
Fixes #385.

The PPTX schema validator passes an absolute schema path to xmllint. With libxml 2.9.13 and spaces in the checkout path, imports resolve through both raw and URL-encoded paths, producing false schema violations for clean decks. Run xmllint from the schema directory with the schema basename so relative imports resolve consistently. Validation diagnostics remain intact.

Add a regression test that copies the validator to a temporary directory containing spaces and validates a generated clean deck. It fails with the original implementation and passes with this change. The complete PPTX engine suite passes: 958 tests, one skipped. Repository-wide typecheck, formatting, lint (zero errors, 13 existing warnings), and license checks pass. The complete cross-workspace suite is not green on this machine: separate Slides system-font assertions failed during the initial run; this PR does not modify Slides.

This was reproduced on an unmodified e064f3a checkout by moving the same worktree between paths with and without spaces. Implemented and tested with AI assistance.


merrick-2002:
Merged, thank you for the clear reproduction in #385 and for the regression test that exercises a real path with spaces.

PR #387, opened 2026-09-15
fix(sheets): infer pivot source from a single selected cell

### LINKED ISSUE #375: Sheets - Create Pivot table from continouos region
### App

GenOffice Sheets (sheets)

### Problem

Hi,
Feature is similar to that from #298 but this time is for Pivot Tables
Currently I have to select all required cells

### Proposed solution

Clicking one cell in a continuous region or Table and selecting Insert -> Pivot Table would run Pivot Table creation for that Region/Table.

BTW:
Genoffice told me that Pivot Table advanced features are in development. So I'll wait patiently and learn/play with stuff currently available.
Thank you,

qinghuanandejiangshi:
Fixes #375.

Creating a PivotTable with B2 selected inside A1:C3 currently uses B2:B2 and offers no fields. Resolve a single-cell selection to its containing table (excluding totals), or its continuous data region, and use that source for both the dialog fields and the creation operation. Explicit selections remain authoritative, and incomplete preloads are not inferred.

Table bounds follow structural edits using the same shift logic as saving. Reuse the existing region scanner and reject capped scans rather than silently returning truncated data. Subtotal field selection keeps its existing behavior.

Validation: 14 focused regression tests pass; the Sheets full suite passed 2,661 tests with one skip, plus the native Rust suites. Typecheck, formatting, lint (zero errors), licenses and repository policy checks pass. Full-workspace testing encountered 17 PPTX schema failures involving space/URL-encoded paths and three Slides system-font assertions, outside the changed files. The independent PPTX path issue is tracked in #385 and fixed separately in #386. Native UI verification passed: selecting B2 inferred A1:C4, produced East=15 / West=20 / total=35, and the saved XLSX pivot cache references A1:C4.

Implemented with AI assistance; the validation results above describe the checks actually run.


merrick-2002:
Merged, thank you for the thorough tests and for handling table bounds and totals rows in the inferred source.

PR #390, opened 2026-09-15
fix(markdown): normalize pasted math delimiters to dollars

aniruddhaadak80:
## Summary

- Normalize pasted math in `apps/markdown/src/renderer/editor/math.ts`: convert `\(...\)` to `$...$` and `\[...\]` to `$$...$$` via `transformPastedText`. Gemini and Claude emit backslash delimiters by default, which previously stayed plain text because only dollar tokenizers exist.
- Why: pasting AI output now renders KaTeX; currency `$5`/`$10` behavior unchanged because only backslash forms convert.

## Related issue

Follow-up to math rendering reports; objective paste-fidelity fix.

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test` — `npm run test -w @genoffice/markdown -- --run tests/math-paste.test.ts` passes 4/4 including inline, block, currency, and empty cases

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Thanks, the intent is right, but the block form as written regresses ordinary Markdown paste. `\[` … `\]` is also how Markdown escapes literal brackets, and AI output uses it constantly for citations: `see \[1\] and \[2\]` would paste as `$$1$$ and $$2$$`, i.e. two block-math nodes. `\(…\)` has the same issue for escaped parentheses, just rarer.

Suggested guard:

- only convert when the inner text looks like LaTeX (contains a `\command`, `^`, `_`, `=`, or an operator next to a digit);
- for `\[…\]` additionally require the delimiters to sit at line start / line end, which is how Gemini and Claude actually emit display math.

Please add `see \[1\] and \[2\]` and `\(note\)` cases to the test showing they are left untouched.

aniruddhaadak80:
Done — fixed in ac5238a6e1fcdf80087cda6f1ff50a88c18ddecb. Narrowed normalizePastedMath: only converts when inner looks like LaTeX (\\\\command, ^, _, =, operator+digit); block \[...\] additionally requires line-start/line-end (whitespace-tolerant). Added regression tests: 'see \\[1\\] and \\[2\\]' and '\\(note\\)' left untouched, plus mid-line block and multiline display cases. Local: vitest math-paste 8/8 pass, prettier clean.

merrick-2002:
Merged, thanks!

PR #391, opened 2026-09-15
fix(pptx-engine): keep tab stops in document order

aniruddhaadak80:
## Summary

- Keep inline `a:tab` in document order in `packages/pptx-engine/src/parse.ts`: add it to the `isArray` list and rewrite it to a tab sentinel run, mirroring the existing `a:br` and `a:fld` handling. Previously tab-separated slide columns glued to `Col1Col2` in the editor model.
- Why: complements the `file-parse` extraction fix with the editor semantic-tree path; tab stops survive parse and render.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test` — `npm run test -w @genoffice/pptx-engine -- --run tests/field-run-order.test.ts` passes 5/5 including new tab-order case

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Thanks for the PR. Two things block this one as-is:

1. **Schema.** In ECMA-376 the children of `a:p` are `pPr`, `r`, `br`, `fld` and `endParaRPr`; `a:tab` only exists inside `a:tabLst` as a tab-stop definition. PowerPoint writes an in-paragraph tab as a literal `\t` inside `<a:t>`, which parse already preserves. Do you have a real-world file that emits a bare `<a:tab/>` between runs? Without one this fixes a structure no producer seems to write, and the test is synthetic.
2. **Save path.** `generate.ts` → `findRunSpans` only recognises `a:r` / `a:br`. With this change a paragraph containing `<a:tab/>` parses to one more run than the original XML has spans, the in-place patch can no longer align, and the whole shape falls back to `rebuildTxBody` (full rewrite, original bytes lost). If the structure is real, the change needs the matching span kind in `findRunSpans` plus an `isTabRun` check next to `isSoftBreakRun`, and a round-trip test.

Happy to look again with a sample file and the generate-side counterpart.

aniruddhaadak80:
Done — fixed in 3f6a1063eab0fa3459a31213edaafb7f2547204b. Save path: findRunSpans now tracks bare <a:tab/> as kind tab (regex <a:tab\\s*/> only, so <a:tabLst>/<a:tab pos> stops ignored), added isTabRun + tabOnly alignment next to isSoftBreakRun, and in-place patch keeps tab bytes; rebuild normalizes to valid literal tab in <a:t> (CT_TextParagraph allows pPr/r/br/fld/endParaRPr; a:tab lives in tabLst). Tests: literal-tab order, bare in-place preservation, tabLst-ignored, rebuild normalization (field-run-order 9/9, parr-tabs/patch clean, prettier clean). Schema note: searched fixtures — no bare <a:tab/> producer found; PowerPoint literal path already preserved and now covered by test. No real-world bare-tab sample to attach.

aniruddhaadak80:
Rebased onto latest upstream/main (70e5149) to retrigger CI — e2e failure was unrelated flake in e2e/docs-spellcheck-reenable.spec.ts:28, main e2e is green, pptx-engine unit tests green.

On reviewer points:

1. Schema: agreed ECMA-376 CT_TextParagraph allows pPr/r/br/fld/endParaRPr only, a:tab lives in tabLst. This change is defensive for third-party producers emitting bare <a:tab/> between runs. Regex is bare-only (<a:tab\s*/>), so <a:tabLst> and <a:tab pos=...> stops are ignored and survive verbatim. Rebuild normalizes sentinel to PowerPoint-valid literal tab in <a:t>.

2. Save path: done — findRunSpans tracks bare tab as kind tab, isTabRun + tabOnly alignment next to isSoftBreakRun, in-place patch keeps tab bytes, rebuild path emits valid literal.

Local validation: prettier check pass, eslint pass, typecheck pass, tests/field-run-order.test.ts 9/9 pass. Full pptx-engine 944 passed locally (3 chart timeouts Windows-only, CI test job was green).

Ready for re-review.

merrick-2002:
Thanks for following through on the save path. Having run the branch against main, I'm going to close this one rather than merge it.

The new `tabOnly` alignment treats a run whose text is a literal tab (which is what PowerPoint actually writes for "Name<tab>Value") like the bare `<a:tab/>` sentinel. Resizing three runs to 28pt patches two and leaves the tab run at `sz="1200"`; typing a character after the tab fails alignment and falls back to `rebuildTxBody`, dropping the original `rPr`. Both work on main today.

Since ECMA-376 has no `a:tab` child of `a:p`, no producer we can find emits one, and the literal tab already round-trips, the extra parse and generate surface is not worth that fidelity risk. Appreciate the other PRs in this batch, those are all in.

PR #396, opened 2026-09-15
fix(electron-utils): block trailing-dot internal hostnames

aniruddhaadak80:
## Summary

- Strip trailing dots from the hostname in `isSafeRemoteUrl` in `packages/electron-utils/src/safe-remote-url.ts` before the `localhost` and blocked-suffix checks. DNS treats `localhost.` as `localhost`, so the gate must too instead of relying on the later DNS lookup.
- Why: closes a classic trailing-dot SSRF evasion in the AI-tool fetch path.

## Related issue

Objective security hardening; no open issue.

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test` — `npm run test -w @genoffice/electron-utils -- --run tests/safe-remote-url.test.ts` passes 49/49 including new trailing-dot cases

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks!

PR #398, opened 2026-09-15
fix(html): escape ampersands in single-quoted attrs

aniruddhaadak80:
## Summary

- Escape `&` as `&amp;` in the single-quoted attribute branch of `set_attr` in `apps/html/src/renderer/document/ops.ts`. Previously only `'` was escaped, so values like `a&b` or literal `&lt;` were written raw and decoded differently on re-parse, corrupting the value.
- Why: one-line round-trip fidelity fix matching the double-quoted `escapeAttr` behavior.

## Related issue

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test` — `npm run test -w @genoffice/html -- --run tests/ops.test.ts -t ampersands` passes including new single-quote ampersand cases

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks!

PR #399, opened 2026-09-15
fix(pdf): use length-preserving case folding in search

aniruddhaadak80:
## Summary

- Use the shared length-preserving `foldCase` in `apps/pdf/src/renderer/search.ts` instead of plain `toLowerCase`. Dotted capitals like `İ` grow from 1 to 2 chars on lowercase, so folded text ran longer than the source and highlight rects drifted or missed.
- Why: parity with the docs find panel and shared `find-text` helper, which already solved this the same way.

## Related issue

Objective search-highlight fix; no open issue.

## Validation

- [x] `npm run format:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test` — extends `apps/pdf/tests/search.test.ts` with dotted-capital length coverage; CI covers the full pdf suite (local pdf runner is flaky on Windows)

List any checks not run and explain why: Local pdf vitest timed out in this Windows checkout; logic mirrors the proven shared helper and CI is proof run.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks!

PR #406, opened 2026-09-15
Add a local MCP server so external agents can drive the Word editor

### LINKED ISSUE #338: Proposal: built-in local MCP server — let external AI agents drive the GenOffice editors
## Motivation

When I ask an agent to produce a Word document today, the agent writes a Python
script — usually python-docx, or COM automation through win32com driving Word
itself. Both paths are slow (spawning and scripting Word for every single
document), and the output quality has a hard ceiling: python-docx has no layout
engine, so every margin, style and table fight is hand-written; COM scripts are
fragile and platform-bound. The results rarely look like a document worth
shipping.

GenOffice already solves this problem internally: its pure-TypeScript docx/pptx
engines produce well-formed, properly styled Office files fast — no Office
automation involved — and the built-in AI panels drive the editors through a
restricted op vocabulary where every edit lands in the same undo history,
journals and save pipelines a human uses. There is no MCP surface yet, so
external agents cannot reach any of this and keep falling back to scripting
Office by hand.

This proposal adds that missing piece: a built-in local MCP server exposing the
editors to external agents (Claude Desktop, Cursor, or any MCP client), so the
agent stops writing throwaway Office scripts and drives GenOffice directly while
the user watches.

## What exists today (working on my fork)

A complete implementation is working on my fork
(`ZKelvins99/genoffice`, branch `feat/mcp-docx-server`). I'd like to align on
direction here before opening PRs — the delivery plan at the bottom is how I
would split it.

- **Server core** (shell main process): Streamable HTTP `/mcp` + legacy SSE, plus
  a stdio bridge script for clients that prefer stdio. Default port 3093,
  **disabled by default**, loopback-only listener, settings pane with
  enable/port/logging controls. Tool instances are per-connection, so two clients
  never share one editing session.
- **Visible editing sessions**: `create_session { family }` opens a real
  docs/slides/sheets tab and drives it through the exact executors and save
  pipelines the built-in AI uses; `save_session { path }` writes to an explicit,
  dialog-free, clobber-guarded path. Per-family content tools
  (`insert_content` / `replace_blocks` / `apply_ops`, `apply_slide_ops`,
  `apply_sheet_ops`) address the active session. The user watches every step in
  the app — the same experience as the built-in AI.
- **Headless generation** (opt-in "background generation" switch, restarts the
  server so the tool list updates): `create_docx` (markdown or SaveBlock[]),
  `create_pptx` (markdown/JSON outline → two-phase transaction), `create_xlsx`
  (value matrix through the same minimal OOXML builder the AI uses).
- **Reading**: `read_docx` (visible text) and `read_pdf` (pdfium text extraction
  in the shell main process — the pdf family is read-only by design; the viewer
  editor is not driven).
- **Format registry**: a single source of truth mirroring the shell's open
  routing and each editor's Save-As filters, so MCP never advertises a format the
  editor cannot produce, and `save_session` refuses mismatched extensions.
- **UI**: an MCP pane in Settings (status, port, background-generation toggle,
  connection info, per-launch log viewer) with capability rows; all strings go
  through the 20-locale i18n resources.

Surface today: **14 tools** by default, 17 with background generation on.

## Safety model

- Off by default; enabling is an explicit user action in Settings.
- Loopback-only listener; no remote access.
- Logging is opt-in, written per app launch (reset at startup), viewable and
  clearable in the settings pane.
- File writes from tools require absolute paths and refuse to replace existing
  files unless `overwrite` is passed.

## Proposed delivery plan (4 PRs, smallest reviewable steps)

1. **Server core + docs family + settings UI** — server, registry, session host,
   docs bridge/tools, Settings pane, i18n (largest PR).
2. **Slides family** — headless `create_pptx` + visible deck session (shares the
   `slides:apply-txn` code path, so undo/journal/autofit all apply).
3. **Sheets family** — headless `create_xlsx` + visible grid session via a
   renderer bridge; explicit-path save reuses the existing sidecar pipeline
   (new optional `targetPath`/`overwrite` fields; the write path itself is
   unchanged).
4. **`read_pdf`** — headless pdfium text extraction + format-registry entry
   (smallest PR).

Every PR would keep to the repo conventions: English-only code and docs, unit
tests per family plus a full-surface acceptance test over real HTTP, e2e specs
next to the existing ones in `e2e/`, and round-trip safety on save paths noted
per PR.

## Questions before I open the PRs

1. Where should the architecture/design document live? This repo has no
   `planning/` directory — I can keep the design doc out of the PRs, or move it
   into `docs/` if you want it in-tree.
2. Protocol emphasis: I implemented both Streamable HTTP and a stdio bridge —
   is one of them preferred as the documented default for the mcp.json example?
3. Is the Settings-pane capability matrix (docs/slides/sheets/pdf rows) fine to
   include in PR 1, or should UI land in a separate follow-up?

ZKelvins99:
Closes #338 (proposal; motivation and the maintainer's reply are there).

Agents asked to produce Word documents today either write `python-docx` (no layout engine, poor fidelity) or drive Office over COM (slow, fragile, platform-bound). GenOffice already ships a correct, fast docx engine — but only its in-app AI panel can reach it. This exposes that capability to Claude Desktop, Cursor and any MCP client.

This is PR 1 of 4. It lands the server plus the Word family; Slides, Sheets and PDF text reading follow, each as an increment on top of this one.

## What's here

**Server.** `apps/shell/src/main/mcp/mcp-server.ts` runs an MCP server inside the shell main process. It speaks Streamable HTTP (`/mcp`) and legacy SSE (`/sse` + `/messages`), and `scripts/mcp-stdio-bridge.js` bridges stdio-only clients to it. Dependencies are injected through `configureMcpRuntime`, so the module never imports `index.ts` back and is testable without Electron.

**Word.** `create_session {family:"docx"}` opens a real docs tab; `insert_content` / `replace_blocks` / `apply_ops` / `read_document` drive it through a renderer command bridge (`docs:mcp-command` / `mcp-result`), reusing the in-app AI's restricted-HTML parser, op executor and save pipeline. The user watches each step in the app rather than an agent writing bytes behind the UI — one of the two directions you called out.

**Headless generation** (`create_docx` / `read_docx`) delegates to the bundled `genoffice` CLI rather than reimplementing conversion, so there is one engine, not two. Off by default; the "background generation" setting exposes it.

**Settings** gains an MCP pane: enable switch, port, log toggle, and a capability summary. The server is **off by default**.

## Why a session instead of editing files directly

`create_session` / `save_session` share one lifecycle across families, so the agent sees two tools instead of six near-identical ones. One session is active at a time, and content tools refuse when the active family is not theirs, so a target is never ambiguous. A later PR adds the same pair for slides and sheets.

## Security model

- Binds `127.0.0.1` only, and validates `Host` and `Origin` on every endpoint. The server has no authentication, so it must not be reachable off-host; request without an `Origin` (curl, a local CLI) is allowed.
- Off by default; nothing listens until the user enables it.
- Tools never silently overwrite: an explicit path that already exists is refused unless `overwrite:true`, and paths must be absolute. Interior writes go through the same `atomicWriteFile` the editors use.
- The CLI is spawned with `ELECTRON_RUN_AS_NODE=1` on the app's own runtime.

## Checks

- `npm run typecheck` (all workspaces)
- `npm run lint`
- `npm run licenses`
- `npm run check:english-comments`
- `npm run fixtures` (no diff)
- `npm run build:all`
- `apps/shell/tests/mcp/**` + `apps/shell/tests/settings-mcp.test.ts` — 71 tests covering the transport handshake (against the real SDK client and raw JSON-RPC), tool registration, schema validation, the session host, the format registry, the storage layer and the settings pane
- `e2e/mcp-visible-doc.spec.ts` — the visible-document flow against a real built app (local only; CI does not run Playwright)

Rebased onto current `main` as requested (was ~50 commits behind at the time of the proposal; rebased cleanly, and the branch is now even with `main`).

One known local-only failure, unrelated to this PR and reproducible on a clean `main` on Windows: `apps/shell/tests/headless-export.test.ts` (12 tests, Windows path handling).

## Where docs live

I kept the design/decision notes out of this PR: they are still in my fork under `planning/`, which upstream does not have. Happy to move them somewhere you prefer (`docs/`, a PR description, or drop them) — tell me which and I will follow up.


merrick-2002:
Merged — thank you for the careful implementation, the test coverage and the clear write-up. Looking forward to the follow-up PRs.

PR #408, opened 2026-09-15
Let MCP agents drive the visible Slides deck

ZKelvins99:
Follow-up to #406 (merged), which wired the MCP server and the Word family. This adds the second editor family from the proposal in #338: the visible Slides deck.

**Driving the deck the user has open**

`create_session` now accepts `family: "pptx"` and opens a real slides tab. New tools:

- `apply_slide_ops` — draw/format/arrange elements through the same transaction engine the app's own `slides:apply-txn` uses;
- `read_deck` — compact model readout (durable ids, EMU geometry, text) so the agent can address what it drew;
- `save_session` writes the tab to an explicit path with no dialog.

Slides sessions live in the main process keyed by the tab's webContents id, so unlike docs there is no renderer round-trip: `applySessionTxn` is now the single funnel shared by the IPC handler and the MCP bridge, which keeps the two paths behaviorally identical (same validation, atomicity/rollback, undo history and op journal).

One bug this surfaced and fixes: `scheduleDeckBroadcast` early-returns for a single-window session on the assumption that an originating renderer will apply its own IPC return value. MCP has no originating renderer, so the canvas stayed blank after a successful transaction. The bridge now pushes `slides:deck-changed` to the session's own tab; multi-window sessions keep the existing scheduled broadcast. `e2e/mcp-visible-deck.spec.ts` covers it by asserting the canvas fingerprint actually changes after an edit, which fails on the pre-fix build.

**Headless generation**

`create_pptx` follows the #406 pattern: it parses the outline into ops and delegates to the bundled `genoffice` CLI, so there is still one engine rather than a second converter in the shell.

Checked: `npm run typecheck`, `npm run lint`, `npm run licenses`, `npm run check:english-comments`, `npm run fixtures`; `apps/shell/tests/mcp` plus `settings-mcp` (82 tests); all app builds; and the `mcp-visible-doc` / `mcp-visible-deck` e2e specs locally.


ZKelvins99:
Pushed `c822eda` addressing the review findings:

- `applySessionTxn` sets `metaDirty` after a successful transaction, so archive-only ops (`setNotes`, a theme commit) no longer leave the session looking clean — closing the tab could have discarded the edit.
- `saveSessionDeckTo` creates the target's parent directory and drops the untitled crash-recovery entry, mirroring `slides:save-as`.
- `read_deck` caps text at 300 characters and table output at three rows, the same shape the CLI's `describeDeck` produces, with `rows`/`textTruncated` markers when a cap bites.
- Group children are marked `coordinates: "local-to-group"` instead of having the group's child frame reported as slide-space EMU.
- `apply_slide_ops` returns the transaction summary without the renderer's render tree. That tree carries base64 image data, so an image-heavy deck was being serialized into the agent's context on every op; the canvas push still uses it.
- A JSON outline's `bullets` entries now render as bullets (a bare string gets a character bullet) while `paragraphs` stays plain text.

Checked: `npm run typecheck`, `npm run lint`, prettier on the changed files; `apps/shell/tests/mcp` (83), `apps/slides` (744), and the `mcp-visible-deck` e2e spec, which now also asserts the render tree is not echoed.

merrick-2002:
#409 is merged, so this one needs a rebase onto main before it can go in. The only conflict is the `get_app_info.formats` assertion in `apps/shell/tests/mcp/mcp-http-surface.test.ts`: with background off, #409 now returns `[]`, so keep that expectation and assert `pptx` in the background-on case. Source files merge cleanly. Ping here once it's pushed.

ZKelvins99:
Rebased onto `main` (with #409) and pushed. `mergeable` is true again and both checks are green.

The only conflict was the one you named, in `apps/shell/tests/mcp/mcp-http-surface.test.ts`:

- kept `expect(infoJson.formats).toEqual([])` for the background-off case, and added the `families` assertion #409 introduced — with generation off the `pptx` family reports `{ save: ['pptx'] }` with no `generate`, the same rule `docx` follows;
- moved the background-on expectation into the headless-generators test, asserting `['docx', 'pptx']` once `create_docx`/`create_pptx` are registered.

Source files merged cleanly, as you said. Ready for another look.

merrick-2002:
Merged — thank you! The visible-deck bridge reusing `applySessionTxn` as the single funnel, plus the single-window `deck-changed` push, is exactly the shape we wanted. Nice work on the follow-up round too.

PR #409, opened 2026-09-15
Fix the review findings on the MCP server

ZKelvins99:
Follow-up to #406, which was merged before this review pass landed. These are the concrete issues I could reproduce from the findings on that PR, all in the docs family and the stdio bridge. No behavior changes beyond what each bullet describes.

**Correctness**

- `read_docx` now passes `docs read --full`. The CLI clips every block to a 200-character preview by default, so a long paragraph came back silently truncated despite the tool's documented contract. I reproduced it against the bundled CLI: 359 characters in, 200 out.
- `resolveOutputPath` rejects a requested path whose extension disagrees with the generated format. Previously `report.pdf` had `.docx` appended, so the tool wrote `report.pdf.docx` and returned a path the caller never asked for.
- `docs:save-to` no longer deletes a pre-existing file when a teardown rollback races an overwrite. The rollback assumed the target was freshly created (true for `docs:save-new`, false here), so a tab closing mid-write destroyed the previous version instead of leaving it in place.
- The stdio bridge never answers a notification, even when forwarding it fails. An unsolicited id-less error line on stdout reads as a protocol violation to a stdio client.

**Session integrity**

- The docs bridge only accepts a command's reply from the tab the command targeted. Request ids are sequential and guessable, so any other docs tab could previously resolve another tab's pending command; it also now rejects in-flight commands when the target tab is destroyed, rather than making the agent wait out the 120-second timeout.
- `apply_ops` routes through `executeTool` instead of calling `executeOps` directly, so an external batch gets the same stale-index guard and numbering context as the built-in agent. Two symptoms this fixes: block indexes written against a stale read could target the wrong blocks, and `setList` received no numbering id.
- The stdio bridge caches the client's handshake and replays it on a reconnected session. A reconnect creates a new, uninitialized server session, so every tool call after a transient disconnect failed as "not initialized" while the client believed it was ready. The replayed `initialize` response is suppressed so the client does not see a duplicate.

**Accuracy and cost**

- `get_app_info` no longer advertises generation the client cannot reach: `create_docx` is hidden when background generation is off, so `formats` and the per-family `mcp.generate` now follow that setting.
- `read_document` is described as what it returns — a clipped `index|type|text` block list for targeting edits — rather than "blocks with current formatting".
- `readTail` reads a bounded suffix instead of the whole append-only log. The settings pane polls it every two seconds, so the old implementation's main-process work grew with total log size.
- An invalid port edit reverts to the port the main process last confirmed rather than a hard-coded 3093, which could disagree with a running server, and a valid edit persists while the server is off.

Checked: `npm run typecheck`, `npm run lint`, `npm run check:english-comments`, prettier on the changed files; `apps/shell/tests/mcp` plus `settings-mcp` (73 tests), and the `mcp-visible-doc` e2e spec. The stdio-bridge notification test and the `read_docx --full` argv assertion both fail on the pre-fix code.


merrick-2002:
Merged, thanks for the thorough follow-up on the review findings.

PR #410, opened 2026-09-15
fix(docx): fold field code split across paragraphs with single-quoted fldCharType

aniruddhaadak80:
## Summary

- What changed? fieldStackAfter in packages/docx-engine/src/parse-fields.ts now accepts single-quoted w:fldCharType values, mirroring the already-merged single-quote handling in field-balance.ts and eq-field.ts. fieldLabel end/begin detection is quote-agnostic too.
- Why is this change needed? Third-party single-quote producers emit w:fldCharType='begin'/'separate'/'end'. The paragraph folder missed them, so IF/DATE field code split across paragraphs never joined: the first marks stayed visible and the tail showed only its own runs instead of the joined result.

## Related issue

Closes #

## Validation

- [x] npm run lint
- [x] npm run typecheck (docx-engine workspace; repo-wide typecheck has a pre-existing pdf2docx pdf-lib interop failure on Windows, untouched by this change)
- [x] npm test (focused: field-display 49 passed, field-balance 8 passed; full docx-engine 1327 passed with 1 parallel-load flake in deep-nested-table that passes in isolation)
- [ ] npm run format:check (Prettier runner fails to spawn on this Windows checkout with EINVAL; change follows existing formatting)

New regression test fails without the fix (markers [false,false,false,false]) and passes with it ([true,true,false,false] plus joined text Left {It is not).

List any checks not run and explain why: format:check not runnable locally on Windows (see above); CI will verify.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks!

PR #411, opened 2026-09-15
fix(shell): validate raw rename name before trimming so trailing-space gate holds

aniruddhaadak80:
## Summary

- What changed? The Home renameFile IPC handler now validates the raw requested name via a new isValidRawRenameName helper before trimming, instead of trimming first and then validating.
- Why is this change needed? Trimming first silently turned 'report ' into 'report' and renamed to a different name than requested, making the trailing-space rejection added in the previous rename-gate change unreachable via IPC. Leading-whitespace names had the same silent-rename surprise. Now surrounding-whitespace input is rejected with the existing localized errBadName gate.

## Related issue

Closes #

## Validation

- [x] npm run lint
- [x] focused vitest: apps/shell/tests/home-rename-validation.test.ts 4 passed (1 pre-existing illegal-chars, 1 same-file, 1 reserved/trailing-dot, 1 new raw-whitespace)
- [ ] npm run typecheck (shell workspace typecheck reports only pre-existing failures in untouched files: xlsx-gateway build types, pdf-lib interop, MCP SDK imports; no errors in rename-validation.ts or the edited handler)
- [ ] npm run format:check (Prettier runner fails to spawn on this Windows checkout with EINVAL; change follows existing formatting)

List any checks not run and explain why: full npm test not run for this renderer-only main-process gate; focused suite covers it. CI will verify.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks!

PR #412, opened 2026-09-15
fix(docs): exclude supplementary private-use and variation selectors from asian char count

aniruddhaadak80:
## Summary

- What changed? The ASIAN_RE surrogate-pair range in apps/docs/src/renderer/word-count.ts narrows from D840-DBFF to D840-D8BF, so only CJK Extensions B-H (U+20000-U+3134F) count as Asian characters. The header comment now documents the bound.
- Why is this change needed? The wider range counted supplementary private-use (planes 15-16, e.g. U+F0000/U+100000) and the variation selectors supplement (e.g. U+E0100) as CJK words, inflating the Words figure versus Word. This is the invited narrowing follow-up to the Southeast Asian/full-surrogate-range word-count change.

## Related issue

Closes #

## Validation

- [x] `npm run lint`
- [x] Focused vitest: apps/docs/tests/word-count.test.ts 8 passed (7 pre-existing + 1 new PUA/variation-selector case)
- [ ] `npm run typecheck` (docs word-count.ts itself is clean; repo-wide run on this Windows checkout reports pre-existing pdf-lib/gateway failures in untouched files)

New test fails without the fix (PUA counted as 1) and passes with it (0), while Ext-B/Ext-H fixtures still count as 1.

List any checks not run and explain why: full `npm test` not run for this stats-only change; focused suite covers it. CI will verify.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks!

PR #413, opened 2026-09-15
fix(pdf): fold AI-tool search queries the same way as the search index

aniruddhaadak80:
## Summary

- What changed? `searchText`, `locateOccurrence`, and `countOccurrences` in apps/pdf/src/renderer/ai/tools.ts now fold the query with the shared `foldCase` helper instead of `toLowerCase()`, matching how the search index itself is built (`search.ts` uses `foldCase`). The snippet window in `searchText` now slices by the folded query length.
- Why is this change needed? `toLowerCase()` grows dotted capitals (`İ` U+0130 → `i̇`, two chars) while `foldCase` is length-preserving by design, so the three AI-tool helpers disagreed with `searchInIndex`: `edit_text`/`insert_text`/note anchoring reported "not found" for text the index had, counts came back 0, and snippet offsets drifted. This completes the case-folding fix from the earlier search change, which left these three sites untouched.

## Related issue

Closes #

## Validation

- [x] `npm run lint`
- [x] Focused vitest: apps/pdf/tests/ai-tools.test.ts dotted-capital case passes (149 tests in file; new `edit_text` İmir test fails without the fix with `isError: true` and passes with it)
- [ ] `npm run typecheck` (untouched-file pre-existing failures on this Windows checkout: pdf-lib interop, xlsx-gateway build types; no errors in the edited files)

List any checks not run and explain why: full `npm test` not run for this focused AI-tool change; focused suite covers it. CI will verify.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks!

PR #414, opened 2026-09-15
fix(html): escape apostrophes in single-quoted style attributes on set_style

aniruddhaadak80:
## Summary

- What changed? `set_style` in apps/html/src/renderer/document/ops.ts now escapes the serialized style value according to the attribute's existing quote style, mirroring the `set_attr` branch fixed earlier: double-quoted keeps `escapeAttr`, single-quoted escapes `&` and `'` (not `"`).
- Why is this change needed? For a document like `<div style='color:red'>`, a `set_style` op with an apostrophe value (`font-family: a'b`) previously emitted `style='font-family: a'b'`, terminating the attribute early and corrupting the round-trip (remainder parses as bogus attributes). Double-quoted originals were safe; single-quoted plus `'` was not.

## Related issue

Closes #

## Validation

- [x] `npm run lint`
- [x] Focused vitest: apps/html/tests/ops.test.ts single-quoted cases pass (pre-existing `set_attr` ampersand test plus the new `set_style` apostrophe test; new test fails without the fix showing the raw `'` breakout, passes with `&#39;`)

List any checks not run and explain why: full `npm test` not run for this focused op change; focused suite covers it. CI will verify.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks!

PR #416, opened 2026-09-15
fix(pdf2docx): anchor arabic leader detection to line end and cap page digits

aniruddhaadak80:
## Summary

- What changed? `hasDotLeaderRun` in packages/pdf2docx/src/analyze/toc.ts now accepts an arabic page number after a dot leader only when the digits run to the end of the line (gather + rest-blank check, capped at 4 digits), mirroring the roman branch right below it. `TOC_LINE_RE` and the roman-branch token test adopt the same `{1,4}` arabic / `{1,6}` roman bounds as the leaderless `PAGENUM_UNIT_RE`.
- Why is this change needed? The arabic path fired on the first digit mid-line, so `Intro .... 12 apples` inflated the leader-line counts used for index-page classification while `TOC_LINE_RE` (anchored) rejected it — the two gates disagreed and form/report pages misclassified as index pages. Unbounded digits also converted `Report ..... 20240` with `pageNumber: "20240"` while the leaderless path correctly rejects 5+ digits as years/zips. Miss-rather-than-misfire hardening in the exact spirit of the earlier digit-cap change.

## Related issue

Closes #

## Validation

- [x] `npm run lint`
- [x] Focused vitest: packages/pdf2docx/tests/toc.test.ts 15 passed (13 pre-existing + 2 new); both new tests fail without the fix (`Intro .... 12 apples` counted as leader, `20240` converted with pageNumber) and pass with it

List any checks not run and explain why: full `npm test` not run for this focused analyzer change; focused suite covers it. CI will verify.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks!

PR #417, opened 2026-09-15
fix(pdf): sanitize fields in encodeStyle so crafted styles cannot shift key fields

aniruddhaadak80:
## Summary

- What changed? `encodeStyle` in apps/pdf/src/renderer/color-runs.ts now validates each field with the same gates `decodeStyle`/`patchStyle` use (hex-only color, charset-only font id, finite in-range size) instead of joining raw values into the `|`-separated key.
- Why is this change needed? The earlier style-key hardening covered `decodeStyle` + `patchStyle` but left `encodeStyle` raw, so a crafted value passed straight to it (e.g. `font: "a|b"`) stored a separator-bearing key (`|a|b|||`) that decodes shifted (`{font: "a"}` with data loss), and non-finite sizes (`NaN`/`Infinity`) round-tripped as non-decimal payloads. This closes the write-side half with the same validators, defense in depth.

## Related issue

Closes #

## Validation

- [x] `npm run lint`
- [x] Focused vitest: apps/pdf/tests/color-runs.test.ts 22 passed (21 pre-existing + 1 new); new test fails without the fix (`|a|b|||` stored) and passes with it (dropped to base)

List any checks not run and explain why: full `npm test` not run for this focused style-key change; focused suite covers it. CI will verify.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks!

PR #418, opened 2026-09-15
fix(sheets): shift whole-row spans on copy/fill like whole-column spans

aniruddhaadak80:
## Summary

- What changed? `offsetFormulaRefs` (copy/fill semantics) in packages/xlsx-gateway/src/domain/formula-shift.ts now shifts whole-row spans (`2:4`) on fill-down via the existing `ROW_SPAN_RE`, mirroring the whole-column-span pass on fill-right. `$`-anchored components stay pinned and off-grid spans become `#REF!`, exactly like the column pass.
- Why is this change needed? The structural path (`shiftFormulaRefs`, fixed for row spans earlier) and the copy/fill path diverged: `REF_RE` needs a column component, so `=SUM(2:4)` copied down stayed `=SUM(2:4)` instead of becoming `=SUM(3:5)` the way `=SUM(B:B)` already becomes `=SUM(C:C)` on fill-right. Excel shifts both axes; the row axis was left stale by accident.

## Related issue

Closes #

## Validation

- [x] `npm run lint`
- [x] New vitest cases in apps/sheets/tests/fill-range.test.ts (shift, `$`-pin, fill-right no-op, off-grid `#REF!`); verified locally by compiling the module standalone and asserting 9 cases (this Windows checkout cannot resolve the `@genoffice/xlsx-gateway` workspace subpath — a pre-existing environment issue also visible in the shell typecheck — so CI runs the vitest file)
- [ ] `npm run typecheck` (same pre-existing workspace-resolution failures in untouched files)

Pre-fix `offsetFormulaRefs('=SUM(2:4)', 1, 0)` returns `=SUM(2:4)` (stale); post-fix `=SUM(3:5)`. Pre-existing column-span and plain-ref behaviors re-asserted unchanged.

List any checks not run and explain why: full `npm test` not run; focused coverage above plus CI. CI will verify.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks!

PR #420, opened 2026-09-15
fix(markdown): serve sibling images of filesystem-root documents

aniruddhaadak80:
## Summary

- What changed? The md-asset:// serve gate in apps/markdown/src/main/markdown-main.ts now uses a new exported `isInDocDir` helper (in the electron-free asset-lifecycle.ts) that builds the directory prefix without doubling a trailing separator. Root-level directories (`/`, `C:\`) already end in one, so `dir + sep` produced `//` / `C:\\` and 403d every sibling image.
- Why is this change needed? This is the main-side half of the filesystem-root sibling-image fix: the renderer side already requests `md-asset:///a.png` for `/note.md`, but the main-process gate rejected it, so every image in a root-level document stayed broken. Nested documents are unaffected; prefix-attack directories (`/docs-evil` vs `/docs`) stay rejected and the directory itself is never served.

## Related issue

Closes #

## Validation

- [x] `npm run lint`
- [x] New apps/markdown/tests/md-asset-root.test.ts 3 passed (root siblings allowed on POSIX + Windows, nested allow/reject incl. prefix attacks, directory itself never served); pre-existing asset-lifecycle 18 + slash-and-image 26 suites still green

Old inline logic verified to reject `/a.png` under `/` and `C:\a.png` under `C:\` (403); new helper allows both.

List any checks not run and explain why: full `npm test` not run for this focused gate change; coverage above plus CI. CI will verify.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks!

PR #422, opened 2026-09-15
fix(docx): read uppercase checkbox checked values as checked

aniruddhaadak80:
## Summary

- What changed? Checkbox state reading is now case-insensitive in both places: the shared `isOn` helper in packages/docx-engine/src/checkbox-control.ts lowercases before comparing, and `checkboxStateOf` in packages/docx-engine/src/parse-props.ts reuses that helper instead of its own case-sensitive copy.
- Why is this change needed? `w14:val="TRUE"` / `w:val="ON"` read as unchecked, so the editor drew ☐ for a checked box (and write-back derived state from the wrong glyph). This mirrors the accepted on/off falsy-list parity used elsewhere; absent value still means checked.

## Related issue

Closes #

## Validation

- [x] `npm run lint`
- [x] Focused vitest: field-display uppercase-checked cases (FORMCHECKBOX `ON` + sdt `TRUE`) — both fail without the fix (☐ drawn) and pass with it (☒ drawn)

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks!

PR #423, opened 2026-09-15
fix(docx): read single-quoted TOC tab stop values and leaders

aniruddhaadak80:
## Summary

- What changed? `tocLeaderOf` in packages/docx-engine/src/parse-fields.ts now accepts single-quoted `w:val` / `w:leader` on `w:tab` stops, completing the single-quote parity family for TOC handling.
- Why is this change needed? A single-quoted `<w:tab w:val='right' w:leader='dot'/>` matched zero right stops (so the leader fell through to the style) or matched the stop but read `leader: 'none'`, rendering dot leaders as a blank gap. Verified: pre-fix the fixture yields `tocLine` with no leader, post-fix with `leader: 'dot'`.

## Related issue

Closes #

## Validation

- [x] `npm run lint`
- [x] Focused vitest: new single-quoted tab-stop case — fails without the fix, passes with it

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks!

PR #424, opened 2026-09-15
fix(docx): honor single-quoted xml:space preserve on part roots

aniruddhaadak80:
## Summary

- What changed? `partXmlSpacePreserve` in packages/docx-engine/src/parse-props.ts now accepts single-quoted `xml:space='preserve'` on part roots (`w:document`/`w:hdr`), not just double-quoted.
- Why is this change needed? PDF-to-DOCX converters set `xml:space` once on the root; with single quotes the inherited preserve scope was missed, so every edge space in the part trimmed and words glued (`HelloWorld`). Display-only one-line parity fix in the whitespace-fidelity family.

## Related issue

Closes #

## Validation

- [x] `npm run lint`
- [x] New unit test in wt-raw-whitespace.test.ts — fails without the fix, passes with it (`"preserve"`/`'preserve'` true, `"default"` false)

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks!

PR #425, opened 2026-09-15
fix(file-parse): include footnote and endnote text in docx extraction

aniruddhaadak80:
## Summary

- What changed? `docxToText` in packages/file-parse/src/docx.ts now appends `## Footnotes` / `## Endnotes` sections with each note's text after the block tree, since footnotes/endnotes live outside the block tree.
- Why is this change needed? AI attachment text silently lost footnote/endnote content that Word shows and search should index — the parser already returned `doc.footnotes`/`doc.endnotes` but the flattener never read them. Same extraction-completeness class as the earlier pptx-tab and lowercase-xlsx fixes.

## Related issue

Closes #

## Validation

- [x] `npm run lint`
- [x] New test builds a docx with footnotes/endnotes parts — fails without the fix (body text only), passes with it (both details present)

List any checks not run and explain why: full `npm test` not run for this focused extraction change; focused suite covers it. CI will verify.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks!

PR #427, opened 2026-09-15
fix(ai-provider): flush TextDecoder at SSE stream end

aniruddhaadak80:
## Summary

- What changed? `sseLines` in packages/ai-provider/src/protocols/shared.ts now calls `decoder.decode()` once more after the stream ends to flush bytes still buffered inside `TextDecoder` under `{ stream: true }`.
- Why is this change needed? A stream cut inside a multibyte char (dropped connection) left the partial bytes buffered forever: the tail vanished silently. With the standard final flush the truncation surfaces as the replacement mark instead of disappearing — this affects all three streaming protocols, which all consume `sseLines`.

## Related issue

Closes #

## Validation

- [x] `npm run lint`
- [x] New test streams `data: 中` plus one dangling lead byte — fails without the fix (tail dropped), passes with it (replacement mark); neighboring sse-frame-tolerance suite still green

List any checks not run and explain why: full `npm test` not run for this focused streaming change; focused suites cover it. CI will verify.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks!

PR #428, opened 2026-09-15
fix(pdf): drop non-integer page targets in printPdf

aniruddhaadak80:
## Summary

- What changed? `printPdf` in apps/pdf/src/renderer/print.ts now requires integer page targets (`Number.isInteger`) alongside the existing range check.
- Why is this change needed? A float like `1.5` passes `>= 1 && <= numPages` but makes pdf.js `getPage` throw in the measure pass, which has no catch — the exception aborts the entire print job and `window.print` never fires. Now floats are dropped like out-of-range entries and the valid targets print.

## Related issue

Closes #

## Validation

- [x] `npm run lint`
- [x] New test `printPdf(doc, [1.5, 2, 99])` renders page 2 only — fails without the fix (float fetched), passes with it

List any checks not run and explain why: full `npm test` not run for this focused validation change; focused suite covers it. CI will verify.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thanks!

PR #432, opened 2026-09-16
Let MCP agents drive the visible Sheets grid

ZKelvins99:
Follow-up to #406 and #408 (both merged). This adds the third editor family from the proposal in #338 — motivation is in that issue: **letting an agent drive the editor the user has open, and reaching clients that cannot run a shell at all.**

**Driving the grid the user has open**

`create_session` now accepts `family: "xlsx"` and opens a real spreadsheet tab. New tools:

- `apply_sheet_ops` — edit the live Univer grid through the same executors the built-in AI uses (`readCellsImpl`, `planFromOps` + `applyChangePlan`, `handleSave`), so external edits land in the edit journal and undo history exactly like in-app ones;
- `read_sheet` — workbook overview (sheet ids/names/extents) plus cell values and formulas, so the agent can address what it wrote;
- `save_session` writes the tab to an explicit path with no dialog.

Unlike the slides bridge, a sheets workbook lives in the **renderer**, so this one needs a request/response channel into the tab — the same plumbing as the docs bridge (`sheets:mcp-command` / `sheets:mcp-result` / `sheets:mcp-ready`). Ops are validated with `workbookOperationSchema` before they run, matching the AI path, so a missing `sheetId` is reported as such rather than surfacing as "Unknown sheet: undefined".

Like the app's own "new spreadsheet", a real blank `.xlsx` is written up front: the sidecar save pipeline needs an on-disk workbook, and the in-memory demo grid cannot be saved.

**Saving to an explicit path** adds an entry point to the existing sidecar pipeline rather than a second writer — the save request gains `targetPath` and `overwrite` so the dialog is skipped. The write path itself is unchanged.

**Headless generation**

`create_xlsx` follows the #406/#408 pattern: it stages the row matrix and delegates to the bundled `genoffice` CLI, so there is still one engine rather than a second converter in the shell. Formula cells get their values computed through the sidecar, which the CLI already does.

**Checks**

`npm run typecheck`, `npm run lint` (0 errors), `npm run licenses`, `npm run check:english-comments`, `FORMAT_BASE_REF=upstream/main npm run check:theme-colors`, `npm run check:skill-version`, `npm run fixtures` (no diff), prettier on the changed files; `apps/shell/tests/mcp` + `settings-mcp` (92 passed); `apps/sheets` (2647 passed; the 21 failures are pre-existing on this Windows host — Rust sidecar and path-handling cases — and reproduce identically on a clean `upstream/main`); all app builds; and `mcp-headless`, `mcp-visible-doc`, `mcp-visible-deck`, `mcp-visible-sheet` e2e locally (4 passed).

One conflict-marker note for reviewers: rebasing onto `main` after #408 merged touched only `get_app_info.formats` in `apps/shell/tests/mcp/mcp-http-surface.test.ts`, which now follows the #409 semantics — with generation off the list is empty, and the sheets family reports `{ save: ['xlsx'] }` with no `generate`, the same rule docs and slides follow.


ZKelvins99:
CI on the first run flagged a real defect in the new test, now fixed in this commit:

`apps/shell/tests/mcp/sheets-tools.test.ts` used a Windows-only literal (`D:/nope/blocked.xlsx`) as the "save target already exists" sentinel. On Linux that path is not absolute, so the tool refused it with `path must be absolute` before the clobber guard was ever reached. The fake bridge and the call now use an absolute path in the test's temp dir and match on the file name, so the assertion tests the overwrite policy on every platform.

The e2e job then hit an infrastructure flake after all 73 tests passed (an Electron process did not exit during worker teardown, `sheets-edit-save`); it passed on the re-run and both checks are green now. Nothing in `apps/` changed between those two runs — only the unit-test file above.

merrick-2002:
Merged, thank you! Third editor family in the line, and the renderer request/response channel mirrors the docs bridge nicely. Looking forward to the last part of #338.

PR #438, opened 2026-09-16
fix(sheets): decode BOM-less UTF-16 CSV instead of keeping NUL garbage

aniruddhaadak80:
## Summary

- What changed? `decodeCsvBuffer` in packages/xlsx-gateway/src/gateway/csv-import.ts now sniffs BOM-less UTF-16 before the strict-UTF-8 attempt: any meaningful NUL presence (impossible in UTF-8/legacy CSV text) routes to UTF-16, with the NUL-heavy side picking LE vs BE.
- Why is this change needed? BOM-stripped UTF-16 passed strict UTF-8 (NUL is valid UTF-8), so decoding stopped there and downstream got NUL-interleaved garbage with zero delimiters — columns lost. Verified: pre-fix `a\0,\0b\0…`, post-fix clean text for LE/BE, ASCII and CJK alike, while plain UTF-8/GBK/Shift_JIS paths are untouched.

## Related issue

Closes #

## Validation

- [x] `npm run lint`
- [x] New vitest cases (ASCII LE/BE + CJK LE without BOM); sheets suites cannot execute on this Windows checkout (pre-existing workspace-subpath resolution failure), so the real module was verified via standalone execution: 7/7 assertions pass post-fix, NUL garbage confirmed pre-fix
- [ ] `npm run typecheck` (same pre-existing workspace-resolution failures in untouched files)

List any checks not run and explain why: full `npm test` not run; coverage above plus CI. CI will verify.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Merged, thank you! The NUL-ratio sniff is a clean way to catch BOM-less UTF-16 before the UTF-8 pass.

PR #433, opened 2026-09-16
fix(pdf): prevent extra blank pages when printing

GoodScholar:
## Summary

PDF printing could turn a single portrait page into two sheets and a single landscape page into three sheets when printer margins were enabled. The hidden editor left its full-height root in the print layout, and landscape bitmaps could overflow the printable height.

Hide the editor root during printing, reset the screen-only document height and overflow, and fit page images proportionally within the printable area. Insert page breaks only between images. Add Chromium PDF-output regression coverage for portrait/landscape, single/multiple pages, and zero/nonzero margins, checking both page counts and image painting on every sheet.

## Related issue

Closes #430

## Validation

- [x] `npm run format:check`
- [ ] `npm run lint`
- [ ] `npm run typecheck`
- [ ] `npm test`

Scoped checks passed on upstream main (`de78d0e`):

- `npx playwright test --config e2e/playwright.config.ts e2e/pdf-print-pagination.spec.ts --reporter=list` — 8 passed.
- `npm run test -w @genoffice/pdf -- tests/print.test.ts` — 13 passed.
- `npx eslint e2e/pdf-print-pagination.spec.ts`
- `npx tsc --noEmit --skipLibCheck --esModuleInterop --target es2022 --module commonjs e2e/pdf-print-pagination.spec.ts`
- `git diff --check`

Repository-wide lint, typecheck, and tests were not run because this change is limited to PDF print CSS and its standalone regression test. Physical printer and native macOS print-dialog verification were not performed.

## Screenshots or recordings

Not applicable to the editor UI. Automated PDF output checks reproduce the reported 2-page portrait / 3-page landscape output before the fix and verify one content-bearing page afterward.

## Contributor checklist

- [x] The change is focused and does not include unrelated reformatting or refactoring.
- [x] User-facing strings use the existing i18n resources. (No user-facing strings added.)
- [x] File open/save changes include an appropriate round-trip or fidelity test. (No file open/save changes.)

PR #440, opened 2026-09-16
Let MCP agents read PDF text

ZKelvins99:
This is the fourth and last piece of the proposal in #338: **reading**, alongside the three visible editors already merged (#406 docs, #408 slides, #432 sheets).

**read_pdf**

`read_pdf` returns the page count, the document info dictionary (title/author) and per-page text with page sizes, optionally limited to a page range (`"3"` or `"1-5,8"`). Extraction reuses the pdfium WASM instance already loaded in the main process — the same one the text editor uses — walking textpages without mutating the document. A page with an empty text layer comes back `hasTextLayer:false`, so an agent can tell a scanned page from a text one instead of silently receiving nothing. Output is capped (80k chars) with a `truncated` flag; pages the caller did not ask for never count as truncation.

**Read-only is the final design.** The pdf app is a viewer and PDF editing is not on the roadmap, so the registry exposes this family as `mcp: { read: 'pdf' }` with no `save` and no `generate` — MCP never advertises a capability it cannot deliver (`get_app_info` reports it that way too). The tool is headless and session-free, so unlike the three editor families it needs no `create_session` and is registered with the visible tools.

**Two pdfium details worth recording**, both handled in `apps/pdf/src/main/read-text.ts`: in this build `FPDFText_GetText` takes **four** arguments (the buffer-size parameter is dropped) and its return value counts the trailing NUL, which must be stripped; and `FPDF_GetMetaText` takes the tag as an **ASCII bytestring** while the value comes back as **UTF-16LE**.

**Motivation** is in #338 — the maintainer asked for "letting an agent drive the editor the user has open, and reaching clients that cannot run a shell at all". This is the read half of the second point.

**Checks**

`npm run typecheck`, `npm run lint` (0 errors), `npm run licenses`, `npm run check:english-comments`, `FORMAT_BASE_REF=upstream/main npm run check:theme-colors`, `npm run check:skill-version`, `npm run fixtures` (no diff), prettier on the changed files; `apps/shell/tests/mcp` + `settings-mcp` (99 passed), `apps/pdf` (758 passed — the 1 failure is the pre-existing Windows-only `generated-output.test.ts` path case, which reproduces on a clean `upstream/main`); built `apps/pdf` and `apps/shell`; and all five `mcp-*` e2e specs pass locally (`mcp-read-pdf` plus the four from the earlier PRs).

One note for review: rebasing this branch onto the post-#432 `main` left two unrelated hunks from the old pre-merge snapshot in shared files (`headless-cli.ts`, `index.ts` — a deleted dead `createXlsxViaCli` helper and a moved function). Both files are back to their `main` state here, so the diff is exactly the `read_pdf` increment.


ZKelvins99:
CI note: `test` is green; `e2e` is red, but with the **same nine failures that `main` already has** — this branch does not introduce them.

I checked the e2e run on `main` itself (`a4d8e1a`, job 104693001542): 83 passed / 9 failed, and this PR is 84 passed / **the same 9** — the extra pass is the new `mcp-read-pdf` spec. All five `mcp-*` specs pass here, including the new one.

The nine split into two groups, both untouched by this PR (the diff is only `read_pdf`):

- 8 × `e2e/pdf-print-pagination.spec.ts` — it calls `chromium.launch()` and dies with `Executable doesn't exist at .../chromium_headless_shell-1234/...`. The spec came in with #433; the e2e job never runs `npx playwright install`, so the browser it needs is absent. Nothing to do with `read_pdf` — it fails before any app code loads.
- 1 × `e2e/markdown-tab.spec.ts:142` — `expect(.doc-editor strong).toHaveCount(0)` got 1 after undo; also failing on `main`.

So the actionable part for this PR is only the `read_pdf` increment. Whether to add `npx playwright install chromium` to the e2e step (or scope that spec to skip when the browser is missing) looks like a separate call — happy to open it as its own small PR if you want it.

merrick-2002:

PR #441, opened 2026-09-16
fix(pptx): read and strip single-quoted showMasterSp

aniruddhaadak80:
## Summary

- What changed? All four `showMasterSp` ("hide background graphics") touchpoints in packages/pptx-engine now accept single-quoted values: slide parse (`parse.ts`), inherited-decoration lookup (`index.ts`), layout-view guard (`master-edit.ts`), and the save-path strip in `patchSlideShowMasterSpXml` (`generate.ts`).
- Why is this change needed? With single-quoted `showMasterSp='0'`, the parser missed the flag (master/layout decorations rendered when PowerPoint hides them), and toggling visibility left the old attribute in place while appending a second `showMasterSp="0"` — a duplicate attribute that fails schema validation and triggers PowerPoint repair. Verified: pre-fix parse yields no `masterSpHidden` and the strip is a no-op; post-fix both behave like the double-quoted forms.

## Related issue

Closes #

## Validation

- [x] `npm run lint`
- [x] New `hidden-slide.test.ts` cases (parse both quote forms + strip/normalize) — fail without the fix, pass with it

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:

PR #443, opened 2026-09-16
fix(pptx): honor true/True table style flags and merges

aniruddhaadak80:
## Summary

- What changed? Table `xsd:boolean` attributes in packages/pptx-engine/src/parse.ts now go through a case-insensitive `xsdBool` helper: the six `tblPr` style flags, `hMerge`/`vMerge` in both table and cell paths, and `tblPr rtl` (previously lowercase-only).
- Why is this change needed? PowerPoint writes `"1"` but third-party writers emit `"true"`/`"True"`; those read as false, so styled tables lost their header row, banding, merges, and RTL — a whole-table visual regression. Verified: pre-fix `firstRow="true" bandRow="True"` renders the header cell with the band tint; post-fix with the solid accent header + banded rows.

## Related issue

Closes #

## Validation

- [x] `npm run lint`
- [x] New `table-style.test.ts` case — fails without the fix, passes with it

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:

PR #445, opened 2026-09-16
Let MCP agents work with the open documents

ZKelvins99:
Closes the gap left by the editor-session tools: an agent could only reach the **blank tab its own session opened**. `read_document` / `read_sheet` / `read_deck` are bound to that session, and `read_docx` reads disk — so a document the *user* is editing was unreachable over MCP. This adds one tool, `open_documents`, with three actions:

- **`list`** — every editor tab with its `id`, `type`, `path`, `title`, whether it has ever been saved, and whether it has unsaved changes.
- **`read`** — one document's **live** content, unsaved edits included.
- **`close`** — closes one, **saving it first by default**; `unsaved:"discard"` drops the changes instead. An untitled dirty document is saved into the default save folder before the tab goes away.

A document is identified by **path**, or by the **tab id** from `list` when it has never been saved. Opening the same file twice (or editing before a first save) still resolves to exactly one document.

**PDF is list-only.** The pdf app is a viewer, so an open PDF has no readable or savable state here: `read` and `close` refuse it with the reason, and the registry keeps advertising the family read-only. Text extraction for a PDF on disk is a separate tool.

**Most of this is wiring, not new machinery.** The docs/sheets/slides bridges are addressed by webContents id, so they drive any open tab, not just the ones a session created — and slides registers its session per renderer on open, so a deck the user opened is already reachable. What had to be added:

- the tab manager's **per-family dirty query** (five families answer synchronously in the main process; docs has to ask its renderer) and `filePath` on `TabSummary`, which was on `TabRecord` but never surfaced;
- a **prompt-free close**. The existing `closeTab` runs the interactive Save / Don't Save / Cancel dialog, which must not appear in a flow the user did not start, so the plain removal step is now its own method;
- for **markdown and html**, which had no bridge at all: a live-read channel plus a dialog-free save. The save reuses each app's own pipeline by pointing the view's save target at the destination first, so asset rewrites and the owned-asset manifest still run — the "untitled first save" path would otherwise open a dialog.

One detail worth recording: the read channel **retries on an interval until the renderer answers**. A request that arrives before the editor has mounted its listener is dropped silently by Electron, which is exactly how the first e2e run failed (it timed out rather than returning stale text). The renderer stays silent while it is still loading, so the main process keeps polling instead of failing on a document that is merely not ready — the same shape as the shell's own control channel.

**Checks**

`npm run typecheck`, `npm run lint` (0 errors), `npm run check:english-comments`, `FORMAT_BASE_REF=upstream/main npm run check:theme-colors`, `npm run check:skill-version`, prettier on the changed files; `apps/shell/tests/mcp` (101 passed, 16 of them new), `apps/markdown` + `apps/html` (451 passed); built markdown/html/shell; and the full `npm run test:e2e` suite (59 passed, 5 skipped) including three new specs — one lists/reads/closes a Markdown file opened through the app's own routing, one does the same for a .docx through the docs bridge, and one types into the live editor then asserts the edit reaches disk only when the close saves it.


merrick-2002:

PR #451, opened 2026-09-16
feat(mcp): edit an open document from the content tools, with audit fixes

ZKelvins99:
Fixes a live regression on `main` and then answers the second half of what you asked for in #338 — *"letting an agent drive the editor the user has open."*

`open_documents` (from #445) can already **list, read and close** the user's tabs, but the content tools could still only reach the blank tab their own `create_session` opened, so an agent had no way to *change* the document the user was looking at.

## `read_pdf` is missing on `main` today

`app-mcp.ts` imports `createPdfTools` but never calls it: #445 dropped the `...createPdfTools()` line when it merged `main` into its branch — only the **import** came back, because the merge conflicted in the import block but not in the return array. The tool is therefore never registered, while `mcpStatus().capabilities` still advertises `'pdf'`, so the settings pane tells users PDF is available and no client can call it.

This is why `apps/shell/tests/mcp/mcp-http-surface.test.ts` is **3/4 failing on `main`** — the assertion `expected [ 'apply_ops', …(13) ] to deeply equal [ 'apply_ops', …(14) ]` is the missing `read_pdf`. Typecheck cannot catch it (`noUnusedLocals: false`) and the only e2e that would is Playwright. Restoring the call turns those three green.

## Editing an open document

The content tools (`insert_content`, `replace_blocks`, `apply_ops`, `read_document`, `apply_slide_ops`, `read_deck`, `apply_sheet_ops`, `read_sheet`) now take an optional `document` argument: a tab id or path, exactly the values `open_documents {action:"list"}` returns. That redirects the edit to the tab the user has open, with no session involved — every family bridge is keyed by webContents id, so a resolved tab needs no `create_session`, and the session's own blank tab is never replaced.

Mutating calls switch to that tab and bring the window forward first, so the user watches the change land rather than discovering it later. Reads deliberately leave the UI alone: an agent sweeping several tabs for context should not make the view jump along. Showing the tab is a courtesy, so a UI failure there is logged and the edit still goes through.

Resolution reuses `open_documents`' own matching rules (exact tab id first, then resolved path) and refuses a document belonging to another family, so both surfaces produce the same guided errors.

## Other fixes

- **Untitled html could not be closed.** `close` resolved the save extension through the generation registry, which only covers the three families with a headless `create_*` tool, so an untitled html document threw `family "html" has no MCP generation format`. Reachable in one step: `File > New > HTML`, type, then `open_documents {action:"close"}` (default saves first). It now reads the family's own save format.
- **A discard left slides' crash-recovery copies behind**, so the next open offered to restore edits the caller had explicitly dropped. The interactive "Don't Save" clears both; the MCP path cleared neither. (docs and sheets were checked too and are consistent with their own interactive behaviour — only slides diverged.)
- **`save_session` reported a failed write as success.** Sheets returns the renderer's `SaveOutcome {ok:false}` as data, which was passed through before ending the session — the agent was told it saved, and left with no session to retry against. It now raises a tool error and keeps the session open. Drivers also disagree on the success shape (docs `{ok,path}`, slides `{path}`), so the shared tool normalizes `ok: true`.
- **A tab that failed to become ready stayed unaddressable for life.** Both renderer readiness polls were bounded (docs: 400 tries at 50ms; sheets: a 20s interval that cleared itself), and readiness is announced exactly once and never re-derived — so a tab whose renderer was still booting when the window closed looked fine in the UI while every command timed out in the shell, reachable on a cold dev start. Both now slow down rather than stop, and the shell-side waits drop to 24s so the caller gets the specific reason instead of the MCP client's own bare timeout.
- **`create_session` could orphan a tab and a file**: a failed sheets session left both the tab and its backing `.xlsx` behind, with nothing in the MCP surface able to clean up either.
- **The sheets bridge was missing the two guards #409 gave docs**: replies were accepted by guessable `requestId` from any tab, and a destroyed tab left its in-flight commands waiting out the full 120s. Same fix as docs.
- **markdown/html live reads stranded each other**: one waiter slot per tab meant a second concurrent read overwrote the first's resolver and left it to hit its 30s timeout.
- **A malformed request body was answered 500 "internal error"**, which reads as a GenOffice fault and hides the real problem; it is now the JSON-RPC parse error it is (400/-32700, 413 when oversized).
- **Docs error text pointed at a tool that does not exist.** The shared executors name `get_document_context`; on the MCP surface the same readout is `read_document`, so the hint sent agents chasing a tool they could not call. `insert_content` also accepted any `afterBlockIndex` — the executor clamps an out-of-range value to the last block, so a mistyped index inserted somewhere unintended while `replace_blocks` rejected the same mistake.
- **A formula's value was unreadable and unsaved.** The text lands synchronously but the result is computed asynchronously, so an external caller reading the cells it just wrote saw `value: null`, and `apply_sheet_ops` answering "ok" while `read_sheet` shows nothing is indistinguishable from a silent failure. The batch result now carries the values once the engine settles, and they are remembered so a later save writes a real cached `<v>` (read back rather than taken from the recalc overlay, which skips the cells a batch just wrote). Workbooks the gateway creates from scratch also now carry a minimal `xl/styles.xml`: Excel tolerates its absence, but the Rust sidecar's formula engine fails the whole import with `specified file not found in archive`, so a blank workbook plus a formula write produced a file the app could open and the CLI could not.
- The `read_pdf` page-count probe read page 1 in full and then re-read it (now a zero-budget probe), and a char budget spent exactly on a page boundary claimed a complete read while pairing `hasTextLayer:true` with empty text.

## Checks

- `npx vitest run --root apps/shell tests/mcp/ tests/settings-mcp.test.ts --no-file-parallelism` — **143 passed** (was 112 passed / 3 failed on `main`)
- `npm run typecheck` for shell, docs, sheets, slides, pdf, markdown, html, xlsx-gateway
- `npm run lint` (0 errors), `npm run check:english-comments`, `npm run licenses`, prettier on changed files
- app suites: docs 2371, sheets 2703, markdown 269, html 182, slides 826, pdf 758, shell 419

Every fix has a test that fails against the old code.

Pre-existing Windows-only failures — reproducing on a clean `upstream/main`, untouched here: `apps/shell/tests/headless-export.test.ts` (12), `apps/sheets/tests/promote-file-atomically.test.ts` (2), `apps/pdf/tests/generated-output.test.ts` (1), and `apps/docs/tests/protect-dialog.test.ts` (1, flaky under full-suite load — passes in isolation).


merrick-2002:
Merged, thank you — the document targeting plus the audit fixes (and restoring read_pdf on main) landed together.

PR #453, opened 2026-09-16
feat(mcp): edit an open document from the content tools

ZKelvins99:
> **Stacked on #451** (the MCP audit fixes), which is still open. The first commit (`fix(mcp): repair the read_pdf regression…`) belongs to that PR — this one builds on it and should land after it. Once #451 merges, the diff here shrinks to the seven commits below.

Answers the second half of what you asked for in #338: *"letting an agent drive the editor the user has open."* `open_documents` (from #445) could already **list, read and close** the user's tabs, but the content tools could still only reach the blank tab their own `create_session` opened — so an agent had no way to *change* the document the user was looking at.

## Editing an open document

The content tools (`insert_content`, `replace_blocks`, `apply_ops`, `read_document`, `apply_slide_ops`, `read_deck`, `apply_sheet_ops`, `read_sheet`) now take an optional `document` argument: a tab id or path, exactly the values `open_documents {action:"list"}` returns. That redirects the edit to the tab the user has open, with no session involved — every family bridge is keyed by webContents id, so a resolved tab needs no `create_session`, and the session's own blank tab is never replaced.

Mutating calls switch to that tab and bring the window forward first, so the user watches the change land rather than discovering it later. Reads deliberately leave the UI alone: an agent sweeping several tabs for context should not make the view jump along. Showing the tab is a courtesy, so a UI failure there is logged and the edit still goes through.

Resolution reuses `open_documents`' own matching rules (exact tab id first, then resolved path) and refuses a document belonging to another family, so both surfaces produce the same guided errors.

## Fixes in this PR

- **`save_session` reported a failed write as success.** Drivers disagree on the success shape — docs returns `{ok, path}`, slides only `{path}` — so an agent checking `ok` across families read the slides save as a failure; the shared tool now normalizes it to `ok: true`.
- **A tab that failed to become ready stayed unaddressable for life.** Both renderer readiness polls were bounded (docs: 400 tries at 50ms; sheets: a 20s interval that cleared itself). Readiness is announced exactly once and never re-derived, so a tab whose renderer was still booting when the window closed looked fine in the UI while every command timed out in the shell — reachable on a cold dev start. Both now slow down rather than stop (50ms for 20s, then 1s), and the shell-side waits drop to 24s so the caller gets the specific reason instead of the MCP client's own bare timeout.
- **`create_session` could orphan a tab and a file.** A failed sheets session left both the tab and its backing `.xlsx` behind, with nothing in the MCP surface able to clean either up; the tab is now closed and the file removed.
- **A malformed request body was answered 500 "internal error"**, which reads as a GenOffice fault and hides the real problem. It is now the JSON-RPC parse error it is (400 / -32700, 413 when oversized).
- **Docs error text pointed at a tool that does not exist.** The shared executors name `get_document_context`; on the MCP surface the same readout is `read_document`, so the hint sent agents chasing a tool they could not call. `insert_content` also accepted any `afterBlockIndex` — the executor clamps an out-of-range value to the last block, so a mistyped index inserted somewhere unintended while `replace_blocks` rejected the same mistake. It is now validated against the live block count.
- **A formula's value was unreadable and unsaved.** The text lands synchronously but the result is computed asynchronously, so an external caller reading the cells it just wrote saw `value: null` — and `apply_sheet_ops` answering "ok" while `read_sheet` shows nothing is indistinguishable from a silent failure. The batch result now carries the values once the engine settles, and they are remembered so a later save writes a real cached `<v>`. They are read back rather than taken from the recalc overlay, which skips the very cells a batch just wrote. Workbooks the gateway creates from scratch also now carry a minimal `xl/styles.xml`: Excel tolerates its absence, but the Rust sidecar's formula engine fails the whole import with `specified file not found in archive`, so a blank workbook plus a formula write produced a file the app could open and the CLI could not.

## Checks

- `npx vitest run --root apps/shell tests/mcp/ tests/settings-mcp.test.ts --no-file-parallelism` — **143 passed** (was 124 before this branch; +19)
- `npm run typecheck` for shell, docs, sheets, slides, pdf, markdown, html, xlsx-gateway
- `npm run lint` (0 errors), `npm run check:english-comments`, `npm run licenses`, prettier on changed files
- app suites: docs 2371, sheets 2703, plus markdown 269 / html 182 / slides 826 / pdf 758 in the earlier run

Pre-existing Windows-only failures, untouched and reproducing on a clean `upstream/main`: `apps/shell/tests/headless-export.test.ts` (12), `apps/sheets/tests/promote-file-atomically.test.ts` (2), `apps/pdf/tests/generated-output.test.ts` (1), and `apps/docs/tests/protect-dialog.test.ts` (1, flaky under full-suite load — passes in isolation).


ZKelvins99:
Consolidated into #451 — both PRs are now a single review. The commits here were fast-forwarded onto #451's branch (`fix/mcp-audit`), which now carries the full stack (8 commits, +1367/-92 in 32 files), so nothing is lost and #451's description covers all of it. Closing this one to keep it to a single PR.

PR #455, opened 2026-09-17
True PDF redaction utility (fixes #295)

### LINKED ISSUE #295: Add Redact Feature
### App

GenOffice PDF (pdf)

### Problem

Could you add a redact/blackout feature that allows users to permanently obscure sensitive content (text, images, or areas) in documents?

If it's available, I couldn't find it.

### Proposed solution

_No response_

alfirus:
Fixes #295 — Add Redact Feature.

**Problem:** annotation-only black boxes leave the underlying text searchable, selectable, and copyable. A proper fix must remove text runs + image pixels under the area, not just paint over it.

**What this adds** (`tools/pdf-redact/`, 3 files):
- `redact_pdf.py` — true redaction via PyMuPDF `add_redact_annot` + `apply_redactions(IMAGE_PIXELS / LINE_ART_REMOVE_IF_TOUCHED / TEXT_REMOVE)`
  - `--search "phrase"` redacts every match (recommended for text)
  - `--rect "x,y,w,h"` explicit areas, repeatable; `--page N` targeting, repeatable
  - working `--interactive` mode; clean save with `garbage=4` (no page duplication)
- `test_redact.py` — end-to-end test: builds a PDF, redacts via rect + search paths, asserts redacted text is gone from extraction, surrounding text preserved, page count unchanged
- `README.md` — usage, coordinates (top-left, points), security notes

**Verified:** `python test_redact.py` → PASS (secret removed, public text kept, 1→1 pages). CLI `--rect` and `--search` both exercised.

**Scope note:** this is a standalone Python utility under `tools/` (where the repo already keeps Python tooling). Full integration into the PDF app UI / pdf-lib pipeline is a follow-up — happy to take direction from maintainers on where that should live.


merrick-2002:
Thanks for taking a run at #295, and for verifying that a redaction has to remove the underlying text runs and image pixels rather than paint over them. That is exactly the bar we want.

We are not going to merge this one, though. It is a standalone Python script: it does not run inside the GenOffice PDF app, the CLI or the MCP server, so a user who installs GenOffice still has no redact. It also pulls in a Python runtime and PyMuPDF (AGPL), and the project has no Python at runtime today. Closing the PR and keeping #295 open.

If you would like to build the real thing, we would welcome a PR that does it inside the app with the PDFium engine we already ship:

- Engine: `apps/pdf/src/main/text-edit.ts` and `image-edit.ts` drive `@embedpdf/pdfium` (wasm). `FPDFPage_RemoveObject`, `FPDFPageObj_GetBounds`, `FPDFText_GetLooseCharBox`, `FPDFImageObj_SetBitmap` and `FPDFPage_GenerateContent` are already exported and used there.
- Save path: `apps/pdf/src/main/save-pdf.ts` is the single entry point (`applySaveRequest`). PDFium edits have to land before pdf-lib touches the bytes; there is a comment explaining why.
- UI: a redact tool next to the existing draw tools in `apps/pdf/src/renderer`, area drag or text selection, black box burned into the page content.
- Acceptance: text objects under the box removed, image pixels under the box blanked, a test that reopens the saved file and asserts the text is no longer extractable, no page count change.

Please comment on #295 with a short design before starting so we can catch anything early.

PR #435, opened 2026-09-16
feat(markdown): add lossless source-backed editing

GoodScholar:
## Summary

- Add a lossless Markdown document session that preserves source boundaries, frontmatter, CRLF/BOM details, protected raw fragments, and save/undo history.
- Add whole-document CodeMirror source mode with syntax highlighting, folding, line numbers, selection synchronization, and safe visual/source transitions.
- Guard protected-source edits behind explicit confirmation and keep AI/source patches synchronized with the visual editor.
- Improve visual editing behavior for empty blocks, heading markers, block gutters, drag handles, and source-editor layout.
- Integrate the latest upstream Markdown sidebar, spellcheck, image-viewer, and bounded-tokenizer changes while preserving lossless source provenance.

## Related issue

No linked issue.

## Validation

- [x] `npm run format:check`
- [x] `npm run lint` (0 errors; existing non-blocking warnings remain outside this change)
- [x] `npm run typecheck`
- [x] `npm test`
- [x] `npm test -w @genoffice/markdown` — 27 files, 495 tests passed
- [x] Electron E2E — all tests passed
- [x] Sheets compatibility gate

All required checks completed successfully on commit `ba5359d` in [upstream CI run 35061897139](https://github.com/genspark-ai/genoffice/actions/runs/35061897139).

## Screenshots or recordings

Not included; source-editor interactions and visual states are covered by the Electron E2E suite.

## Contributor checklist

- [x] The change is focused and does not include unrelated refactoring.
- [x] User-facing strings use the existing i18n resources.
- [x] File open/save changes include round-trip and fidelity coverage.

PR #458, opened 2026-09-17
feat(markdown): add opt-in unchanged-source round trips

GoodScholar:
Opening and saving an unchanged Markdown document through the legacy editor normalizes its source. This first, default-off stage preserves the original UTF-8 text when the editor document and envelope still match the loaded snapshot, including after undo. Edited documents continue to use the existing serializer.

This replaces the previous large editor replacement in this PR with a small foundation, following the staged direction requested in #435. The existing editor schema, typing/Enter handling, AI editing, and default serialization remain in place.

- Opt in with `localStorage.setItem('mdapp.experimentalRoundTrip', '1')` in the Markdown renderer, then reload. Remove the key and reload to disable; the flag is off by default and also falls back to off when storage is unavailable.
- Save and MCP reads share the serializer. Comparisons happen only when serializing, with no added work per keystroke.
- Save As image rewrites return the actual persisted text so an unchanged source snapshot can be rebased safely; concurrent edits do not get captured as saved content.
- Tests exercise every currently tracked Markdown file (73 files, including README variants) through the legacy load path and compare unchanged output byte-for-byte. No files are skipped. This is the current repository corpus, not a claim to reproduce the maintainer's exact earlier 78-file dataset.

Validation: 360 Markdown unit tests, 3 Chromium renderer tests (flag off/on, edits and undo, save/MCP consistency, repeated Save As with HTML and images), Markdown typecheck/build, targeted ESLint and Prettier checks. The browser tests are added to CI.

Scope: this is unchanged-document source preservation, not a full lossless serializer for edited documents. It does not change or claim to fix the existing schema's bold/code validation, nor ship the source editor, projection/session engine, protected-block editing, block splitting, or performance work. Those remain separate follow-up stages as requested by the maintainer. No unrelated CSS, CONTEXT, planning documents, or PDF changes are included.

The complete previous implementation and regression fixes are preserved at [`GoodScholar:codex/markdown-lossless-full-backup-35c37a5`](https://github.com/GoodScholar/genoffice/tree/codex/markdown-lossless-full-backup-35c37a5) as the source for those follow-ups. This PR remains on the original `codex/markdown-lossless-editing` branch; history has not been rewritten.


merrick-2002:
Merged — thank you!

This is exactly the kind of staged follow-up we hoped for after #435: a small, self-contained foundation instead of a wholesale editor replacement. A few things we particularly appreciated while reviewing:

- The serializer stays out of the typing path and only compares at save / MCP-read time, so there is no per-keystroke cost.
- `writtenText` coming back from Save As is a thoughtful detail — rebasing the snapshot on what was actually persisted, rather than what the renderer sent, closes a real gap after image rewrites.
- The test corpus walks every tracked Markdown file in the repository byte-for-byte, with no skips, and the browser spec covers flag on/off, edits, undo, and repeated Save As. That made the review straightforward.
- Scope discipline: no unrelated CSS, docs, or planning files this time. Much easier to take.

The flag stays opt-in for now while we shake it out alongside the rest of the Markdown save path; we will follow up here when it is on by default. Thanks again for sticking with this — looking forward to the next piece.

PR #461, opened 2026-09-17
fix(pptx-engine): honor the 1904 date system in chart categories

### LINKED ISSUE #460: Chart category dates ignore the 1904 date system
### App
GenOffice Slides (slides)

### Version and OS
Source revision 6f8a892; Windows, Node 22.23.2. Reproduced in the pure TypeScript pptx-engine unit tests.

### What happened?
Chart category dates ignore the chart-level 1904 date system (`c:chartSpace/c:date1904`). Every numeric category is converted with the 1900-system epoch (serial minus 25569), so on a 1904-system workbook every date label renders about 1,462 days early.

Minimal reproduction: parse a line-chart part whose categories are serials 37261/37262 with format m/d/yyyy. Without date1904 the labels are 1/5/2002 and 1/6/2002 (correct). Add `<c:date1904 val=1/>` to the chartSpace and parse again.

Expected: 1/6/2006 and 1/7/2006.

Actual: still 1/5/2002 and 1/6/2002 — the flag is never read.

### Logs / screenshots
A regression suite added before the fix produced 21 failing chart cases (flag spellings, serial boundaries, every category-chart plot type, combos, sparse caches, save round-trips). The 1900-system behavior is unchanged when the flag is absent or false.

aniruddhaadak80:
## Summary

- `parseChartXml` now reads the chart-level `<c:date1904>` flag and threads it through shared category formatting (including combos that take categories from a later plot). A present flag with a missing `val` defaults to true per the OOXML schema; `1`/`true`/`on` (any case, either quote style) select the 1904 epoch, everything else keeps the existing 1900 behavior.
- 1904-system serials use the 24107 epoch offset instead of 25569, fixing date labels that previously rendered about 1,462 days early on 1904-system workbooks. Non-date numeric and string categories are untouched.
- Added 30 regression cases: flag spellings (1904 on / 1900 preserved), serial boundaries and formats, all 11 category-chart plot types, combo fallback, sparse caches with non-numeric values, and byte-preserving no-op plus unrelated-edit saves with the flag set.

## Related issue

Closes #460

## Validation

- [x] `npm run format:check` (direct Prettier check on both changed files passes; `git diff --check` clean)
- [x] `npm run lint` (scoped ESLint on both changed files, zero warnings)
- [x] `npm run typecheck` (`npm run typecheck -w @genoffice/pptx-engine` passes)
- [x] `npm test` (`chart.test.ts`: 101 passed; full `@genoffice/pptx-engine` suite: 986 passed, 17 skipped, 4 failed — all 4 are 5s timeouts on heavy `01_standard_business.pptx` fixture tests in unrelated files and each passes in isolation, i.e. Windows parallel-load flakes, not assertion failures)

The regression suite failed 21 cases before the fix and passes fully after it.

## Screenshots or recordings

Not applicable. This changes parsed chart category labels, not editor UI.

## Contributor checklist

- [x] User-facing strings use the existing i18n resources. No new UI strings.
- [x] File open/save changes include an appropriate round-trip or fidelity test. No-op and unrelated-edit saves assert chart bytes are preserved.


merrick-2002:
Merged, thank you! The epoch handling and the byte-preserving round-trip tests were exactly what this needed.

PR #464, opened 2026-09-17
fix(file-parse): keep empty shared-string cells empty and preserve malformed-ref text

aniruddhaadak80:
﻿## Summary

- What changed? `packages/file-parse/src/xlsx.ts`: `cellText` now treats an empty, whitespace-only, non-numeric, or out-of-range shared-string `<v>` as empty instead of coercing it via `Number('') === 0` into `shared[0]`; `columnIndex` now returns `-1` when a cell ref has no leading column letters, and the row assembler appends such cells in document order instead of writing `cells[-1]`.
- Why is this change needed? AI attachment indexing over `.xlsx` was silently corrupting rows: any `<c t="s">` with a missing/empty `<v>` inherited the first shared string, and any malformed `@_r` (for example `r="1"`) dropped its text entirely. Both are real-world producer quirks and caused wrong answers with no error signal.

## Related issue

Closes #

## Validation

- [x] `npx prettier --write` on changed files (repo `npm run format` wrapper fails on Windows with `EINVAL` on `prettier.cmd`; direct Prettier reports unchanged)
- [x] `npm run typecheck -w @genoffice/file-parse`
- [x] `npm test -w @genoffice/file-parse` (46 passed, including 2 new regression tests)

List any checks not run and explain why: full-repo `npm test` / `npm run lint` / root `typecheck` not run for this isolated engine change; package-scoped equivalents pass.

## Screenshots or recordings

Not applicable.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test (new `xlsxToText` regression tests for empty/out-of-range shared refs and malformed cell refs).


merrick-2002:
Merged, thank you! Backporting to the internal tree now.

PR #465, opened 2026-09-17
fix(ai-search): parse gsk JSON when trailing logs follow the payload

aniruddhaadak80:
## Summary

- What changed? `parseGskOutput` in `packages/ai-search/src/gsk.ts` now tries the longest valid JSON block from each candidate start line and shrinks past trailing log lines, instead of parsing everything from the start line to end-of-output.
- Why is this change needed? When the `gsk` CLI appends lines such as `[INFO] done` after the JSON payload, the previous slice-to-end parse always failed and surfaced as `No JSON found` even though a valid payload was present. This caused flaky search/image/project failures.

## Related issue

Closes #

## Validation

- [x] `npx tsc --noEmit` in `packages/ai-search`
- [x] `npx vitest run tests/gsk.test.ts` (28 passed, including 3 new trailing-noise cases)

List any checks not run and explain why: full-repo suite not run for this isolated parser change; package-scoped test/type/lint pass.

## Screenshots or recordings

Not applicable.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test (not applicable; parser-only change with regression tests).


merrick-2002:
Thanks. The trailing-log case is real, but the fix regresses the default output: `gsk` pretty-prints (`JSON.stringify(x, null, 2)`), and scanning from the end for the first line that starts with `{` or `[` picks up the last inner element of an array of objects, so the parser returns that element instead of the root object. Please anchor on the outermost opening line (first `{` or `[` at column 0) and trim the log lines after the matching close, or try `JSON.parse` from each candidate start and take the earliest one that succeeds.

merrick-2002:
Merged, thanks. We adjusted the parser on our side to prefer the outermost object, since gsk pretty-prints by default and the inner-element match could win on arrays.

PR #468, opened 2026-09-17
fix(docx-engine): clamp protection spinCount to prevent digest DoS

aniruddhaadak80:
## Summary

- What changed? Added `DEFAULT_SPIN_COUNT = 100000` and `MAX_SPIN_COUNT = 1000000` plus an exported pure helper `resolveSpinCount(raw: unknown)` in `packages/docx-engine/src/protection.ts` (re-exported from `src/index.ts`). `verifyProtectionPassword` now fails fast on a missing salt and on spin counts above the maximum before any `Subtle.digest` work; `hashProtectionPassword` validates through the same helper. Added `packages/docx-engine/tests/protection-spincount.test.ts`.
- Why is this change needed? The untrusted `w:cryptSpinCount` value from document XML previously fed directly into the iterated SHA-512 loop, so a hostile file could force millions of digest calls and hang open (CPU DoS).

## Related issue

Closes #

## Validation

- [x] `npx tsc --noEmit` in `packages/docx-engine`
- [x] `npx vitest run tests/protection-spincount.test.ts` (8 passed) plus `write-protection`, `hostile-input`, `protected-text-edit` (30 passed)

List any checks not run and explain why: full-repo suite not run for this isolated engine change; package-scoped checks pass.

## Screenshots or recordings

Not applicable.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test (digest-free clamp tests plus existing protection round-trips).


merrick-2002:
Thanks. Two things before merge: `verifyProtectionPassword` now throws on a missing salt or an oversized `spinCount`, but the callers in `App.tsx` (around line 1943) and `ProtectDialog` do not catch, so a document with an unusual protection block turns a wrong-password result into an unhandled error. Please return `false` there instead of throwing. Also `resolveSpinCount` clamps while `verify` rejects; pick one policy for both (clamping is friendlier to legacy files).

aniruddhaadak80:
Both points addressed in 28c6853, verified locally (33/33 protection tests pass, package typecheck clean): 1) verifyProtectionPassword no longer throws on missing salt or oversized spinCount - both return false, which is the existing wrong-password path in App.tsx submitModifyPwd and ProtectDialog (both unguarded, as you noted). 2) One policy for untrusted counts, with one deliberate deviation to flag: creation/normalization still clamps via resolveSpinCount, but verify fail-closes WITHOUT hashing above MAX instead of clamp-verifying. Reason: a clamped verify burns up to a million digests just to return false - I measured it past the 20s vitest timeout. A stored count above MAX can only come from a hostile or hand-crafted file, so skipping the digests is fail-closed and equivalent in outcome. Happy to switch to literal clamp-verify if you prefer the symmetry. The e2e red on the last run is an Electron harness startup error (startupData null, not part of any test) - unrelated to this diff.

merrick-2002:
Merged, thanks!

PR #471, opened 2026-09-17
fix(agent-core): redact unquoted secrets and sanitize tool outputs

aniruddhaadak80:
## Summary

- What changed? Extended `sanitizeAgentPayload` in `packages/agent-core/src/loop.ts` to also redact unquoted secret assignments (`password=`, `secret_key:`, `api_key=`, plus `secret`/`private_key` hyphen/underscore variants) with a bounded linear-match value pattern, and added sanitize-on-store for tool outputs so secrets read from files are redacted before entering history. The `onToolExecuted` event still carries the original output for local UI display. Added tests in `sanitize-payload.test.ts` and new `sanitize-tool-output.test.ts`.
- Why is this change needed? The previous patterns only matched quoted values and only sanitized the user message, so `password=hunter2` in a file read via a tool result was re-sent to the remote model verbatim every turn (privacy leak).

## Related issue

Closes #

## Validation

- [x] `npx tsc --noEmit` in `packages/agent-core`
- [x] `npx vitest run tests/sanitize-payload.test.ts tests/sanitize-tool-output.test.ts` (13 passed) plus `tests/loop.test.ts` (54 passed)

List any checks not run and explain why: full agent-core suite not run; scoped sanitize files plus loop regression pass.

## Screenshots or recordings

Not applicable.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test (not applicable; redaction unit tests).


merrick-2002:
Thanks for the thorough pass. We are going to pass on redacting tool outputs before they enter history: the model edits the user's document from those outputs, so a redacted value gets written back into the file as `[REDACTED_SECURE_TOKEN]`, and the unquoted `secret:` pattern matches ordinary prose. Redacting only the user's own input is deliberate. If you want to pursue this, an opt-in that redacts the transcript persisted to disk (not what the model sees) is a scope we could take.

merrick-2002:
Closing this one as discussed above: redacting tool outputs before the model sees them writes placeholders back into the user's document, and the bare `secret:` pattern hits ordinary prose. The redaction-of-user-input path stays as is. If you want to revisit, an opt-in that redacts only the transcript persisted to disk is a scope we would take as a new PR. Thanks for the rest of the batch, which is all merged.

PR #472, opened 2026-09-17
fix(markdown): allow formatting around inline code

GoodScholar:
Markdown such as **`bold code`** or [`linked code`](https://example.com) parses into text with both formatting and code marks. The default code schema excludes all other marks, leaving an invalid document and causing `RangeError: Invalid collection of marks for node text: bold,code` when the document is validated.

Allow inline code to coexist with surrounding formatting and links. The code extension stays at the same pinned version, with an explicit dependency. Code contents remain literal. This is the opening/schema-fix stage following #435 and merged #458; it does not reintroduce the source editor or projection engine.

Regression coverage opens, validates, edits, saves and reopens code inside bold, italic, strike and links. The existing repository corpus now also requires every loaded document to pass schema validation, alongside byte-identical unchanged serialization. Before the fix, 27 tests failed (4 focused cases and 23 current corpus files); after the fix, all 73 current Markdown files pass. These counts refer to the current corpus, not the maintainer's earlier 78-file dataset.

Validation: 365 Markdown unit tests, 27 Chromium renderer tests, Markdown typecheck/build, targeted ESLint and formatting checks. The browser coverage adds 20 real-keyboard Enter cases across headings, paragraphs, bullet/ordered/task lists, at middle/end positions, with the opt-in flag on and off. It checks that subsequent input lands in the split block and that both text and document structure survive save/reopen.

Four performance cases use 127,000-byte and 317,000-byte fixtures with the flag on/off. Each requires 20 actual inserted characters and an insertion-command-to-next-animation-frame p95 below 100 ms. This measures editor transaction/frame scheduling, not end-to-end hardware keyboard latency. The existing CI browser step runs these gates. This is a targeted matrix, not a reproduction of the maintainer's original 81/573 edit dataset.

These tests protect the current legacy editing path. Full source-aware rewriting of edited documents remains outside this PR; no claim is made that the complete lossless editor has been restored.


merrick-2002:
Thanks for keeping this rebased. The Code mark `excludes: ''` fix is right (the `**\`x\`**` RangeError is real). One thing before merge: the `insertion-to-frame p95 stays below 100 ms` assertion in `roundtrip-serializer.spec.ts` will flake on shared CI runners. Please log the number instead of asserting on it, or put the threshold behind an env flag that is off in CI.

merrick-2002:
Merged, thanks!

PR #474, opened 2026-09-17
feat(docs): separate East Asian and Latin font settings

### LINKED ISSUE #462: Feature request: Separate East Asian and Western/Latin font settings in Docs
### App

GenOffice Docs (docs)

### Problem

Please support separate font settings for East Asian (Chinese, Japanese, and Korean) text and Western/Latin text within the same selection, paragraph, or style, similar to Microsoft Word's Asian text font and Latin text font controls.

A common requirement for bilingual documents and academic papers is to use SimSun (宋体) for Chinese and Times New Roman for English and numerals. Users should be able to apply this pair to an entire mixed-language paragraph without selecting and formatting every language segment separately. Relying on missing-glyph fallback does not express this requirement reliably: a Chinese font may also contain Latin glyphs, so the English text can retain the Chinese font's Latin design.

Related: #335 already mentions separate Chinese and English fonts as part of a broader set of word-processing improvements. This request focuses on that feature and its expected behavior. #86 concerns discovering installed fonts; this request concerns assigning different available fonts by script.

### Proposed solution

- Add separate **East Asian font** and **Western/Latin font** selectors to the document font settings, with both choices applicable to the same mixed-language selection.
- Make both settings available in paragraph/character styles and document defaults, so a bilingual font pair can be applied consistently throughout a document.
- Changing one font setting should preserve the other setting and unrelated formatting such as size, bold, and italic.
- Preserve separate font assignments when importing, saving, and reopening DOCX files, including assignments inherited from styles or theme fonts. In WordprocessingML, this should respect the distinction between `w:rFonts` attributes such as `w:eastAsia`, `w:ascii`, and `w:hAnsi`, and their theme equivalents, rather than flattening them into a single font family.

For example, select the whole paragraph:

> 中文测试 English text 123

Set the East Asian font to **SimSun (宋体)** and the Western/Latin font to **Times New Roman**. Chinese characters should render in SimSun, while English letters and ASCII digits render in Times New Roman. Editing either font selector should change only its intended text, and the result should survive a DOCX save/reopen cycle in GenOffice and Microsoft Word.

Manually formatting each language segment is a possible workaround, but it becomes tedious and error-prone in long documents and when text is edited or pasted. Explicit settings for both font groups would make bilingual document formatting much more practical.

GoodScholar:
## Summary

Mixed-language selections can now use independent East Asian and Western/Latin font settings. For example, applying SimSun and Times New Roman to `中文测试 English text 123` keeps Chinese in SimSun and English letters/digits in Times New Roman, without formatting each language segment separately.

- Add two explicit font selectors with selection, document-default, paragraph-style, and character-style scopes. Resolve inherited values per slot and display mixed selections independently.
- Preserve unrelated formatting and untouched font/theme slots during DOCX save/reopen, including East Asian-only edits that inherit their Latin font. Keep complex-script fonts unchanged.
- Route East Asian text to its assigned font even when the Latin font contains CJK glyphs, including the read-only split pane. Coordinate asynchronous style previews with saving and document switches.

## Related issue

Closes #462

## Validation

- [x] Prettier check on all changed files.
- [x] ESLint on changed TypeScript files: no errors; one existing `App.tsx` hook-dependency warning.
- [x] `npm run typecheck -w @genoffice/docs`
- [x] `npm run test -w @genoffice/docs`
- [x] `npm run test -w @genoffice/docx-engine`
- [x] Regression coverage for independent slots, mixed/inherited values, style/default persistence, editor-to-DOCX round trips, subsequent typing, and split-pane rendering.

Validation was rerun after integrating with current upstream `main`, retaining its lazy-media/full-file save handling. Unrelated application suites and repository-wide lint/typecheck were not run. Microsoft Word manual interoperability verification remains outstanding.

## Screenshots or recordings

Manual Chromium verification confirmed that both selectors and the scope control are visible, and that changing East Asian text to KaiTi followed by Latin text to Arial leaves the CJK text in KaiTi. No recording is attached.

## Contributor checklist


merrick-2002:
Thank you, this is a substantial piece of work and it holds up well: the docs and docx-engine suites are green, the ribbon selectors write the slots explicitly, untouched `rFonts` slots (including `cs` and the theme attributes) survive save and reopen, and the East Asian routing does not catch Cyrillic, Arabic or Hebrew. Three things need to change before we can merge:

1. **Per-keystroke cost of `script-fonts.ts`.** The decoration set is rebuilt for the whole document on every transaction (the cache is keyed by `doc`, which is new after each edit). Measured: 16 ms per key at 2k paragraphs, 94 ms at 5k, 1.27 s at 20k. Please map the previous set through `tr.mapping` and recompute only the textblocks the transaction touched.
2. **The right-click font dialog (`ContextMenu.tsx`) still routes by the name regex and writes only `font`.** With the new `explicitEa` precedence in `marks.ts`, a run with `eastAsia=SimSun` that gets KaiTi from the dialog renders SimSun, the ribbon shows SimSun, and the saved file has KaiTi. The dialog should write the same slots as the ribbon (or become the two selectors).
3. **The "(Body)" ribbon entry used to clear the explicit run font so it inherits again.** It now writes the resolved name explicitly, and falls back to a hardcoded `'SimSun'` when the document has no East Asian default, so an English document gets SimSun written into its runs. Please restore the clear/inherit action and drop the hardcoded fallback.

Two smaller suggestions: serialize `onFontSettings` (await the pending promise before reading the refs, otherwise two overlapping calls overwrite each other), and filter latent/hidden styles out of the scope `select` and reset the scope when the document changes.

merrick-2002:
Merged, thanks. We are following up on our side with the per-keystroke decoration cost on very long documents and the context-menu font dialog.

Issue #340, opened 2026-09-13
3 new Features Request


krishnavel93:
### App

GenOffice Docs (docs)

### Problem

hi sir, i have 3 new suggestions
1. add complete math (katex or mathjax support) in markdown 
2. add export to image / png option in file menu
3. In pdf text edit tool instead of making the bold originally, change or add the border colour of words or text selected to black (default)... it will look like bold virtually, at the same time text formatting wont get break... this feature available in pdfgear pdf editor and pdnob pdf editor

i hope u add this feature soon. Thanks.

merrick-2002:
Thanks a lot for the detailed suggestions — all three are well taken.

1. **Math in Markdown**: KaTeX is already built in. Inline `$...$` and block `$$...$$` LaTeX render in the Markdown app, and there is an "Insert formula" entry in the `/` slash menu. If a specific expression does not render for you, please paste it here and we will look into it.
2. **Export as image / PNG**: Slides already has "Export as Images…" in the File menu; we will add the same option to Docs (and Markdown).
3. **PDF text edit bold via stroke**: agreed — keeping the original font and drawing a thin stroke in the text color preserves the glyph widths and the layout, instead of swapping to a bold font file. We will implement this.

We will get these done as soon as we can and update here when they ship.

krishnavel93:
sir when typing math working but when paste from gemini ai or claude ai the math text not get rendered it remains as plain text when pasting..pls fix that.actually this same isues is a long time problem in marktext software too but there only block math alone not working when get pasted.

merrick-2002:
All three are now in v0.10.915: KaTeX math in Markdown (typed, and since the follow-up fix, pasted `\[ \]` / `\( \)` math from AI chats is normalized to `$` delimiters), File → Export as Images in Docs, and the PDF text tool renders bold as a synthetic stroke so the original font is left intact. Closing; thanks for the suggestions, and open a new issue if any of them misbehaves.

PR #476, opened 2026-09-17
fix(markdown): add paginated PNG export with resilient image loading

GoodScholar:
Add Markdown File → Export as Images, producing one 192 DPI PNG per PDF page in a dedicated new folder under the selected directory. Export writes are bound to the requesting renderer, never overwrite existing pages, and incomplete sessions are cleaned up on failure. The UI shows localized progress and errors.

Broken or CSP-blocked images no longer abort the export: each image decode failure is tolerated, consistent with the existing print path. Editor-only separator images are excluded from printed content.

This PR now contains only PNG export. Formula paste was split into #492, which independently targets main and addresses the Wikipedia fallback duplication and plain-text parsing review findings. Refs #340.

Validation on this branch:
- 393 Markdown unit tests passed.
- 7 browser tests passed, including successful export, cancellation, and write failure with a broken image in the document.
- Real Electron File-menu PNG export passed: PNGs decode, every page contains content, the local image survives, and page count matches PDF.
- Markdown build passed. Markdown typecheck and changed-file lint passed before splitting; existing Shell build/typecheck results remain applicable because the split does not alter Shell code.
- Independent review verified the decode failure fix.

Full monorepo checks were not run. The unrelated, pre-existing ordered-list paste limitation (`1. Energy: \(E=mc^2\)` remains literal) is documented in #492.

CI follow-up: synced with current main and fixed the inherited html2docx typecheck failure by narrowing optional page-background screenshot IDs before constructing `Set<string>`. Full workspace typecheck now passes locally; 9 IR-validation tests pass. The full local unit-test run stops in existing remote-image tests because local DNS resolves the test host to private/benchmark addresses rejected by the SSRF guard. Latest GitHub CI: both `test` and `e2e` passed.


merrick-2002:
Thanks. Both halves are solid: the export session design (dedicated folder, writes bound to the requesting renderer, no overwrite, cleanup on failure) is tighter than what we had, and the pasted-math heuristics pass the false-positive cases we tried (`$5 and $10`, `$HOME`, inline code, fences, table cells). Three things need to change before merge:

1. **PNG export fails on any broken image.** `App.tsx` awaits `Promise.all(imgs.map((img) => img.decode()))` without a catch, so one dead link or CSP-blocked remote image aborts the whole export. PDF export has no such limit. Use `.decode().catch(() => {})` as the other call site in the same file already does.
2. **Wikipedia MathML pastes twice.** `annotation.closest('math')` replaces only the `<math>` element and leaves the sibling `img.mwe-math-fallback-image-inline/display`, so the result is a math node plus a duplicate SVG image. Treat `.mwe-math-element` as the carrier or drop the fallback image.
3. **Formula detection changes how the rest of the paste is parsed.** In `clipboardTextParser`, when any formula is found the whole text goes through full Markdown parsing, otherwise the default plain-text path; `# Title` becomes a heading only if the same paragraph also contains `$x^2$`. Please convert only the formula tokens and leave the rest on the default path.

Known gap worth noting in the description: plain text starting with `1. ` is taken by the ordered-list paste handler before your parser runs, so `1. Energy: \(E=mc^2\)` stays literal. That is pre-existing.

Would you consider splitting this into two PRs? The PNG export can merge as soon as (1) is fixed; the math paste can iterate on its own.

merrick-2002:
Merged, thanks!

PR #477, opened 2026-09-18
fix(mcp): keep computed numbers out of the sheet text path, and address sheets by name

ZKelvins99:
Fixes two defects in the MCP sheet surface, reported from a real session. Both are visible to an agent that reads back what it just wrote; the first also reaches the saved file.

## A repeating formula came back as a string

`=1/3` reported `"0.333333"`, `read_sheet` showed the same, and the saved `.xlsx` carried `t="str"` with the truncated digits. An A/B against the bundled CLI proved the engine was never wrong — `genoffice sheet apply` writes `<v>0.3333333333333333</v>` for the same formula, and the sidecar returns exactly that number.

The value was the view model's **rendered text**. Two interceptor layers rewrite it: number formats turn dates/percentages into display strings, and General re-renders a number to fit the column's width (`numfmt-fix.ts` `formatGeneral` — which is also why the poisoned precision varied with the file: 8 decimals in a wide column, 6 in a narrow one). The reader reported `value` and dropped the model value behind it, so anything machine-facing received display text. The save path already preferred the raw value, but it reads through this same reader, so the protection never saw it.

- `lazyCellReader`'s model value now travels with the read (`rawValue`).
- `modelCellValue` (univer-sync.ts) picks the model value, keeping the rendered text only for the cached-value fallback — where the engine errored and the file's cache is deliberately the better display.
- The apply response, `read_sheet`, and the save path all use it. `read_sheet` keeps the rendered text as `display`, so a caller can still see what the user sees.

On disk this now matches the CLI exactly: `<c r="B1"><f>1/3</f><v>0.3333333333333333</v></c>`, no `t="str"`.

## Worksheet ids do not survive a reopen

Sheet ids are minted per session: closing and reopening the same file turned `jg_nIT-Asw-tL3tvR3f4q` into `sheet-3`, and an op holding the old id failed with `Unknown sheet` — or, worse, would write into whatever sheet next carried that id. The CLI has always addressed sheets by name, which is stable and is what the user sees on the tab.

The MCP ops now accept `sheet` / `sourceSheet` / `targetSheet` (the CLI's vocabulary) alongside the id fields, resolved in the renderer where the open workbook's sheets are known. `read_sheet` takes `sheet` too. An unknown name fails the whole batch before anything applies and lists the real sheets, rather than dropping the reference and letting the op fall back to the active sheet and write into the wrong place. An id in the name slot is still accepted, so nothing that worked stops working.

## The edit is visible

`apply_sheet_ops` now switches to the sheet it is editing and scrolls to the batch's first cell — the jump the find bar performs on a hit — so an edit to a sheet the view is not showing does not land silently off-screen.

**Checks**

- typecheck clean across all workspaces; lint 0 errors
- sheets unit suite: 246 files / 2732 tests pass (two pre-existing failures on main stay unrelated: `promote-file-atomically`, `xlsx-recalc` — both reproduce on a clean tree)
- shell MCP suite: 139 pass
- new `e2e/mcp-sheet-values.spec.ts` drives the built app over Streamable HTTP: `=1/3` applies as `1/3`, reads back as `1/3`, saves at full precision without `t="str"`; a two-sheet workbook built by the CLI is edited by name, the named sheet gains focus, and an unknown name is refused naming the available sheets
- `check:english-comments`, prettier on changed files: clean

Note on the report's other observations: the `fullCalcOnLoad` self-healing in Excel is unchanged, and the `sheetId` churn it describes is addressed by the name vocabulary above rather than by making ids stable.


merrick-2002:
Merged, thank you! Great write-up — reading the model value instead of the rendered text was exactly the right cut, and the name-based sheet addressing lines up with the CLI.

PR #482, opened 2026-09-18
fix(html2docx): validate IR, abort image wait, restore safely

aniruddhaadak80:
## Summary

- What changed? Added exported `normalizeIr`/`raceWithAbort` in `packages/html2docx/src/convert.ts` (re-exported from `src/index.ts`): the in-page extractor IR is validated (rejects non-arrays, null items, missing/empty `type`, malformed `shotId`) with a descriptive `Error` instead of a later `TypeError`; the caller `AbortSignal` is raced into the per-image wait so abort during slow decode rejects immediately; screenshot `finally` restores are isolated per block so one failure cannot skip the rest or leave the page mutated. Added `packages/html2docx/tests/ir-validation.test.ts`.
- Why is this change needed? Compromised or malformed page JS could crash conversion, abort during image load still waited out the timer, and a single restore failure aborted remaining restores and left the page mutated for later shots.

## Related issue

Closes #

## Validation

- [x] `npx tsc --noEmit` in `packages/html2docx`
- [x] `npx vitest run tests/ir-validation.test.ts` in `packages/html2docx` (9 passed)

List any checks not run and explain why: full-repo suite and end-to-end browser suite not run; package-scoped pure unit tests pass.

## Screenshots or recordings

Not applicable.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test (not applicable; converter robustness with unit tests).


merrick-2002:
Merged, thank you for hardening the extractor boundary. We are fixing the `Set<string>` typecheck slip on our side and it will land with the next sync.

aniruddhaadak80:
Fixed in 77f59c0: the pageBgShotIds construction mapped node.shotId (optional on IrNode) after a truthiness filter, so tsc still saw (string|undefined)[] and raised TS2769 against Set<string>. Narrowed with as string post-filter. Verified with npm run typecheck on packages/html2docx locally - clean. CI re-running on the new head.

aniruddhaadak80:
Update with wider significance: the TS2769 this fixes is red on mainline itself right now (packages/html2docx/src/convert.ts builds Set<string> from optional shotId without narrowing - verified on current main). That is why sibling PRs like #467 fail typecheck on files they never touch. Landing this unblocks the test job repo-wide, not just here. Fix verified locally with the package typecheck; CI run on 77f59c0 pending.

aniruddhaadak80:
Correction to my two comments above: this PR merged at 07:14 WITHOUT the narrowing fix (my 77f59c0 push landed on the merged branch afterward and changed nothing on main). Mainline is still red at convert.ts:317, which is why sibling PRs keep failing typecheck. Follow-up opened as #499 with exactly that one-line fix, verified locally. Sorry for the confusion.

PR #488, opened 2026-09-18
fix(dialogs): add role, focus, and escape semantics

aniruddhaadak80:
## Summary

- What changed? Gave Stamp, Properties, Cutout, Crop dialogs (PDF) and the Print dialog (Slides) proper dialog semantics: `role="dialog"`, `aria-modal="true"`, and `aria-label` from an already-used title key (`stampTitle`, `propsTitle`, `imageCutout`, `imageCrop`, `appPrintTitle` — no new strings). Added Escape-to-close where missing (Stamp, Properties), focus-first-field on mount with return-focus on unmount everywhere, and `role="alert"` on the Cutout/Crop error spans to match `PasswordDialog`. Added `apps/pdf/tests/dialog-a11y.test.ts` (3 tests: semantics, Escape/backdrop close, focus management).
- Why is this change needed? Screen-reader users never entered dialog mode on these dialogs, tab could escape past the backdrop, and keyboard users had no consistent Escape path — while the sibling `PasswordDialog` already did all of this correctly.

## Related issue

Closes #

## Validation

- [x] `npx tsc --noEmit` (apps/pdf shows only pre-existing untouched `src/main` pdf-lib export errors; Slides reports nothing for `PrintDialog`)
- [x] `npx vitest run tests/dialog-a11y.test.ts` in `apps/pdf` (3 passed)

List any checks not run and explain why: full-repo suite not run; scoped checks pass. No CSS touched so the theme-color gate is unaffected.

## Screenshots or recordings

Not applicable.

## Contributor checklist

- [x] User-facing strings use the existing i18n resources (existing keys only).
- [x] File open/save changes include an appropriate round-trip or fidelity test (not applicable; accessibility semantics with component tests).


merrick-2002:
Merged, thank you for bringing these dialogs in line with PasswordDialog.

PR #492, opened 2026-09-18
fix(markdown): convert pasted formulas while preserving plain text

GoodScholar:
Pasted formulas can remain literal even when typed formulas render correctly. Convert only formula tokens in plain-text clipboard content, preserving the default paragraph and active-mark behavior for surrounding text. Recover TeX from rich KaTeX and Wikipedia MathML carriers without retaining duplicate visual fallback images, including display formulas split across HTML line breaks.

Split from #476 so PNG export and formula paste can be reviewed and merged independently. Both PRs target main. Refs #340.

Validation on this branch:
- 411 Markdown unit tests passed, including 23 formula clipboard cases covering Wikipedia inline/display fallback removal, plain-text heading/link syntax, code protection, and display-math paragraph boundaries.
- 14 browser tests passed, including native clipboard paste and save/reopen.
- Markdown typecheck, changed TypeScript ESLint, formatting, and diff checks passed.
- Independent review completed; the extra-empty-paragraph finding was fixed and rechecked.

Known pre-existing limitation: plain text beginning with `1. ` is intercepted by the ordered-list paste handler before this parser runs. For example, `1. Energy: \(E=mc^2\)` remains literal. This PR does not change that handler.

Clipboard tests use fixed format fixtures rather than live authenticated AI service pages. Full monorepo checks were not run.

CI follow-up: synced with current main and fixed the inherited html2docx typecheck failure by narrowing optional page-background screenshot IDs before constructing `Set<string>`. Full workspace typecheck now passes locally; 9 IR-validation tests pass. The full local unit-test run stops in existing remote-image tests because local DNS resolves the test host to private/benchmark addresses rejected by the SSRF guard. For commit `4e05564`, the full Linux E2E job passed (8m57s), including the previously failing worksheet-navigation case. The independent full-workspace `test` job is still running; its prior run passed.

Sheets E2E follow-up: background range loading queues active-sheet restoration callbacks. An older callback could undo a later explicit MCP navigation to that same sheet, producing the intermittent Summary/Data 2024 failure. Explicit MCP focus now invalidates previous restoration callbacks for that workbook; subsequent background loads retain the existing protection against stealing focus. The original helper was extracted without changing its other behavior.

Validation: deterministic fake-timer regression failed before the fix and passed afterwards; 20 focused tests, Sheets typecheck and changed-file lint passed (three pre-existing React-hook warnings). The rebuilt Electron MCP sheet suite passed three consecutive runs (6 tests). Local E2E used the installed GenOffice native sidecar because Rust is unavailable locally; Linux CI builds the sidecar from source. Independent code review found no blocking issues.


merrick-2002:
Merged, thanks. The Sheets focus fix bundled here is welcome too; we will keep an eye on it in the MCP suite.

PR #499, opened 2026-09-18
fix(html2docx): narrow optional shotId for pageBgShotIds set

aniruddhaadak80:
## Problem\nMainline typecheck is red: packages/html2docx/src/convert.ts builds Set<string> from IrNode.shotId (optional) without narrowing, so tsc raises TS2769. Every PRs test job fails on files it never touches (e.g. #467, #468).\n\n## Solution\nAssert string post-filter; the truthiness filter guarantees presence.\n\n## Impact\nUnblocks the test job repo-wide. No behavior change: identical runtime values, type-level only.\n\n## Evidence\n- npm run typecheck on packages/html2docx locally: clean\n- Follows up #482, which merged before this line could ride along


merrick-2002:
Thanks for catching this. The same fix landed through #521 (type predicate variant), so main typechecks again and this one is now redundant. Closing as superseded; your other PRs from this batch have all been merged.

PR #505, opened 2026-09-19
fix(agent-core): guard streamText cap and extract errors

aniruddhaadak80:
## Summary

- What changed? `streamText` in `packages/agent-core/src/stream-text.ts` now normalizes `maxChars` to a finite value greater than zero (fallback `200000` for `NaN`/`Infinity`/`0`/negative) and routes `partialOrEmpty`, `onDelta` progress, and `onDone` through a `safeExtract` helper that catches `extract` throws, keeps the raw reply as fallback text, and carries the error message into the outcome. Progress-listener throws are likewise ignored so the stream still settles. Added `packages/agent-core/tests/stream-text-guard.test.ts`.
- Why is this change needed? The size cap never fired when `maxChars` was `NaN` and tripped immediately when `<= 0`; worse, any `extract` throw escaped the transport callback and left the returned promise unsettled forever (hang).

## Related issue

Closes #

## Validation

- [x] `npx tsc --noEmit` in `packages/agent-core` (clean)
- [x] `npx vitest run` in `packages/agent-core` (92 passed, including 5 new guard tests)

List any checks not run and explain why: full-repo suite not run; scoped to `packages/agent-core`.

## Screenshots or recordings

Not applicable.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test (not applicable; stream reliability with unit tests).


merrick-2002:
Merged, thanks!

PR #506, opened 2026-09-19
fix(pptx-engine): bound slide id allocation scan

aniruddhaadak80:
## Summary

- What changed? `nextSlideId` in `packages/pptx-engine/src/slide-ids.ts` no longer spreads the used-id set into `Math.max` (stack overflow on decks with hundreds of thousands of slides); it computes the max with a loop. Out-of-range attributes (below 256 or above 2147483647, invalid per ST_SlideId) are ignored instead of forcing the ceiling scan, and the ceiling fallback now throws a descriptive `slide id space exhausted` error instead of looping forever when no id is free. Added `packages/pptx-engine/tests/slide-ids-guard.test.ts`.
- Why is this change needed? A crafted `presentation.xml` with extreme id attributes could stall or crash slide insertion (DoS); allocation must stay fast and bounded.

## Related issue

Closes #

## Validation

- [x] `npx tsc --noEmit` in `packages/pptx-engine` (clean)
- [x] `npx vitest run tests/slide-ids-guard.test.ts tests/slide-id-range.test.ts` in `packages/pptx-engine` (7 passed, including the existing ceiling-fallback tests)

List any checks not run and explain why: full-repo suite not run; scoped to `packages/pptx-engine`.

## Screenshots or recordings

Not applicable.

## Contributor checklist

- [x] User-facing strings use the existing i18n resources (error string is developer-facing only).
- [x] File open/save changes include an appropriate round-trip or fidelity test (allocation covered by unit plus existing round-trip tests).


merrick-2002:
Merged, thanks!

PR #507, opened 2026-09-19
fix(pdf2docx): ignore non-finite inputs in median

aniruddhaadak80:
## Summary

- What changed? `median` in `packages/pdf2docx/src/geometry.ts` now ignores non-finite inputs (`NaN`/`Infinity`/`-Infinity`); when nothing finite remains it returns `0`, matching the existing empty-list behavior so all current `|| 12` fallbacks keep working. Extended the median block in `packages/pdf2docx/tests/geometry.test.ts`.
- Why is this change needed? Geometry inputs come from raw PDF numbers, so corrupt metrics slip in. The old code let `Infinity` flow straight through (the `|| 12` fallbacks catch `NaN`/`0` but not `Infinity`), poisoning every downstream consumer at once — e.g. an infinite column gap that never splits, or an infinite font size. Callers without any fallback (baselines, pitches) got `NaN` instead.

## Related issue

Closes #

## Validation

- [x] `npx vitest run tests/geometry.test.ts tests/toc.test.ts` in `packages/pdf2docx` (25 passed)
- [ ] Full `typecheck` — `tsc --noEmit` in `packages/pdf2docx` reports only pre-existing `pdf-lib` CJS/ESM interop errors in untouched `tests/helpers/fixtures.ts` (junctioned `node_modules` artifact; zero errors in changed files)

List any checks not run and explain why: full-repo suite not run; scoped checks pass.

## Screenshots or recordings

Not applicable.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test (aggregate hardening with unit tests).


merrick-2002:
Merged, thanks!

PR #508, opened 2026-09-19
fix(file-parse): resolve backslash and above-root pptx targets

aniruddhaadak80:
## Summary

- What changed? `resolveTarget` in `packages/file-parse/src/pptx.ts` now normalizes Windows-style backslash separators to forward slashes before resolving, and explicitly clamps `..` segments at the zip root so a hostile `../../..` chain cannot read as a deeper traversal. Extended `packages/file-parse/tests/office-formats.test.ts` with backslash and above-root regression cases.
- Why is this change needed? Decks produced by Windows tools can emit `slides\slide1.xml`-style targets, which previously missed lookup entirely and silently dropped the slide from AI indexing. Above-root chains now resolve deterministically to the zip root instead of relying on implicit pop-on-empty behavior.

## Related issue

Closes #

## Validation

- [x] `npx tsc --noEmit` in `packages/file-parse` (clean)
- [x] `npx vitest run tests/office-formats.test.ts` in `packages/file-parse` (31 passed, including the pinned missing-part skip behavior)

List any checks not run and explain why: full-repo suite not run; scoped to `packages/file-parse`.

## Screenshots or recordings

Not applicable.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test (resolution covered by regression tests; existing skip semantics preserved).


merrick-2002:
Merged, thanks!

PR #509, opened 2026-09-19
fix(docx-engine): base-26 chart columns and hole size default

aniruddhaadak80:
## Summary

- What changed? Exported `colLetter` from `packages/docx-engine/src/chart.ts` (via `src/index.ts`) and fixed it to proper base-26: series columns start at B, so index 24 → Z, 25 → AA, 26 → AB (previously plain charCode arithmetic emitted `[`, `\` past Z). Doughnut `holeSize` parsing now falls back to the spec default 50% on unparseable values instead of zeroing (which silently dropped the setting from the model). Added `packages/docx-engine/tests/chart-cols.test.ts` plus an invalid-holeSize case in `chart-parse-model.test.ts`.
- Why is this change needed? Any chart with 26+ series generated workbook refs Word repairs or drops (`Sheet1!$$1`-style garbage), and a corrupt `holeSize` silently changed the model.

## Related issue

Closes #

## Validation

- [x] `npx tsc --noEmit` in `packages/docx-engine` (clean)
- [x] `npx vitest run tests/chart-cols.test.ts tests/chart-parse-model.test.ts tests/chart-insert.test.ts` in `packages/docx-engine` (41 passed; new tests fail on the old code)

List any checks not run and explain why: full-repo suite not run; scoped to `packages/docx-engine`.

## Screenshots or recordings

Not applicable.

## Contributor checklist

- [x] File open/save changes include an appropriate round-trip or fidelity test (27-series build plus parser round-trip included).


merrick-2002:
Merged, thanks!

PR #512, opened 2026-09-19
fix(cli): harden MCP http host and file handling

aniruddhaadak80:
## Summary

- What changed? `packages/cli/src/mcp/http.ts`: `baseUrlOf` no longer trusts client-controlled `X-Forwarded-Host/Proto` by default (poisoned headers made the server hand out attacker-host download URLs); honored only with explicit `GENOFFICE_TRUST_PROXY_HEADERS=1`. Upload byte counting moved into a pipeline `Transform` (a separate `data` listener races the pipeline for chunks). `handleDownload` maps `stat` failures to 404 and destroys the response on read-stream errors instead of piping unguarded. Upload `decodeURIComponent` falls back to `safeName` sanitization (201) instead of 500. `packages/cli/src/mcp/files.ts`: `remoteName` uses the same `decodeSafe` fallback; the redirect loop fails fast on non-http(s) targets. Added `packages/cli/tests/mcp-http-harden.test.ts`.
- Why is this change needed? Host-header poisoning is a URL-integrity security issue, and malformed input plus download races surfaced as 500s/hangs instead of clean 4xx/fallbacks.

## Related issue

Closes #

## Validation

- [x] New-test logic executed green against the real modules via an isolated Node harness (13 checks: poisoned host ignored + opt-in honored, `%ZZ` fallbacks, redirect protocol gate, missing/raced download → 404, oversize → 413)
- [ ] Full `vitest`/`typecheck` in `packages/cli` — the worktree's junctioned `node_modules` predates the CLI workspace links (`@modelcontextprotocol/sdk` missing; even untouched base tests fail identically there). `tsc` reports zero errors in changed files; CI with a fresh install runs the new suite

List any checks not run and explain why: see above; will watch CI and fix any red immediately.

## Screenshots or recordings

Not applicable.

## Contributor checklist

- [x] User-facing strings use the existing i18n resources (no user-facing strings).
- [x] File open/save changes include an appropriate round-trip or fidelity test (upload→download round-trip asserted in the new tests).


merrick-2002:
Merged, thanks. When bringing this in we kept the proxy-header opt-in but documented it, and the loopback default stays; the upload, download and decode fixes land as is.

PR #533, opened 2026-09-20
fix(ai-provider): cap streamed tool-argument buffers per tool

aniruddhaadak80:
## Summary

- Adds a per-tool streamed argument buffer cap (`MAX_TOOL_JSON_CHARS`, 512k) shared by the OpenAI-compatible and Anthropic protocols: exceeding it aborts the turn with a guided error instead of growing `pending.json` without bound.
- Previously a malicious or broken endpoint streaming argument fragments forever never tripped the idle watchdog (bytes keep arriving) and could OOM the process.
- Adds regression tests on both protocols: endless 1KB-fragment streams abort with the cap error, fragment counts stay bounded, and nothing is emitted.

## Related issue

None.

## Validation

- [x] `npm run format:check` (touched files, prettier clean)
- [ ] `npm run lint`
- [x] `npm run typecheck` (`npx tsc --noEmit` clean for the package; full check via CI)
- [x] `npm test` (`packages/ai-provider/tests/stream.test.ts`: full file green)

- Package lint and the full workspace matrix are left to CI.

## Screenshots or recordings

Not applicable.

## Contributor checklist


merrick-2002:
Thanks — a per-tool buffer cap is the right guard where the idle watchdog cannot see, merged.

PR #539, opened 2026-09-20
fix(cli): cap --ops input size before read and parse

aniruddhaadak80:
## Summary
- What changed? Added MAX_OPS_BYTES 16MB cap in packages/cli/src/ops-input.ts with stat pre-check for files and byte check for stdin, throwing resource_limit CliError.
- Why? readOpsInput read arbitrary GB file/stdin then JSON.parse downstream OOMs. MCP http already caps at 32MB; ops path had none.

## Related issue
Closes #

## Validation
- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test
Added packages/cli/tests/ops-input.test.ts (oversized rejects with resource_limit, small passes). Full CI will verify.

## Screenshots or recordings
Not applicable.

## Contributor checklist
- [x] File open/save changes include an appropriate round-trip or fidelity test. (Not applicable - input guard only.)


merrick-2002:
Thanks for the update — the test fix is good, but the cap itself is still the issue for us: ops JSON legitimately inlines pictures as `data:image/...;base64` (`addPicture.bytes`), and four or five phone photos already exceed 16 MB. Elsewhere we allow a single 20 MB image and the MCP HTTP server accepts 32 MB bodies, so a 16 MB total would reject decks the app itself produces. The stdin check also runs after the full read, so it doesn't prevent the OOM it describes. Closing; if a bound is wanted here it should track `MAX_TRANSFER_BYTES`.

merrick-2002:
Reopening — on reflection this is worth landing with a different bound rather than dropping. What would make it mergeable:

- Raise the cap to match `MAX_TRANSFER_BYTES` (256 MB) so decks that inline pictures as `data:image/...;base64` in `addPicture.bytes` keep working, and derive it from that constant instead of a new literal.
- For stdin, enforce the cap while accumulating chunks (stop reading once the running total passes it) rather than after the full read, so the check actually prevents the allocation it describes.

The test fix you already pushed is good as is. Happy to merge once those two are in.

merrick-2002:

PR #544, opened 2026-09-20
feat(docs): let the AI see pictures inside the document via analyze_media

psvanzyl:
## Problem
In Docs, the agent reads the document as a block list of text. An embedded image block renders as `0|image|` with no content, so prompts like "describe the picture on this page" produce: *"I can see that an image is there, but its picture content isn't visible to me as text."*

The shared `analyze_media` tool (`packages/ai-search`) already exists and is wired up in **Slides** (main IPC → preload → skill tool, gated on the configured media provider). Docs has the same media settings (`media.analysisProvider`, incl. BYOK/OpenAI-compatible custom endpoints) but no way to call it.

## Change
Ports the Slides wiring to Docs:
- `packages/ai-search`: `loadMediaReference` now accepts `data:` URLs — images pasted into a document live as base64 data URLs in the image block, so this is the only way to hand them to the media provider.
- `apps/docs` main / preload / shared ipc: expose `ai:analyze-media` (mirrors the Slides handler).
- `apps/docs` skill + tools: new `analyze_media` tool — takes `blockIndex` (resolved from the image block's stored src) or a `url`, plus `requirements`; gated on `mediaAnalysisAvailable` exactly like `generate_image`.
- `apps/docs` protocol: an image block with no text preview now renders as `(picture, no text — use analyze_media with blockIndex N to see it)` so the model discovers the tool instead of reporting blindness.
- i18n(ai): `aiSumAnalyzeMedia` / `aiSumAnalyzeMediaDone` added to all 21 locale shards.

## Test
- `npm run typecheck` — clean for `@genoffice/ai-search` and `@genoffice/docs`.
- `npm run test -w @genoffice/ai-search` — 66 passed; `-w @genoffice/docs` — 257 files / 2493 tests passed.
- `eslint` clean on all touched files.
- Manual E2E: patched build against a local OpenAI-compatible vision endpoint; image-in-document prompt returned a correct description via the new tool.


merrick-2002:
Merged — thank you, @psvanzyl! Nice port of the Slides analyze_media wiring over to Docs; the lazy-media handling was a good catch.

PR #545, opened 2026-09-20
feat: add dynamic AI header and provider/model badge across all apps

alfirus:
## Summary

Adds dynamic AI provider/model display to the AI panel header and provider badge on assistant messages across all 5 apps (docs, html, markdown, slides, pdf).

## Changes

- **New component:** `ProviderModelBadge` in `packages/ui/src` - renders a small badge showing the active AI provider and model name
- **Dynamic header:** AI panel header now shows the current provider name and model (e.g. "OpenAI gpt-4o") instead of static "Genspark"
- **Provider badge on messages:** Assistant chat messages include a `ProviderModelBadge` for consistent provider attribution
- **All 5 apps updated:** docs, html, markdown, slides, and pdf AiPanel components

## Files changed

- `packages/ui/src/ProviderModelBadge.tsx` (new)
- `packages/ui/src/index.ts` (export added)
- `apps/docs/src/renderer/ai/AiPanel.tsx`
- `apps/html/src/renderer/ai/AiPanel.tsx`
- `apps/markdown/src/renderer/ai/AiPanel.tsx`
- `apps/pdf/src/renderer/ai/AiPanel.tsx`
- `apps/slides/src/renderer/ai/AiPanel.tsx`


merrick-2002:
Thank you for putting this together, @alfirus — the shared `ProviderModelBadge` component is tidy and the five-app sweep was clearly done with care.

We're going to pass on this one, and the reason is product rather than code: the panel header is where the brand mark and title live, and we want to keep that fixed regardless of which provider is configured. Per-message attribution is an interesting idea, but to be correct it needs the provider/model recorded on the message at the time it was generated (otherwise switching providers relabels the whole history), and that's a broader change we'd rather design on our side first.

One small heads-up for future renderer work: import shared provider tables from `@genoffice/ai-provider/browser` — the root entry pulls in Node modules and breaks the Docs/Slides production build.

Closing this for now. If provider attribution lands on our roadmap we'll reference this PR. Thanks again for contributing.

PR #546, opened 2026-09-20
fix(file-parse): cap attachment input bytes and extracted text length

aniruddhaadak80:
## Summary
- What changed? Added MAX_ATTACHMENT_BYTES 20MB pre-stat check and MAX_EXTRACTED_CHARS 200k truncation with marker in packages/file-parse/src/parse.ts; all office/text/pdf paths truncate.
- Why? parseFileToText read arbitrary-size attachments and joined all rows/slides with no cap; giant log or 1M-row xlsx OOMs AI ingest and blows model context.

## Related issue
Closes #

## Validation
- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test
Added truncation unit + large-file parse tests in text-formats.test.ts. Full CI will verify.

## Screenshots or recordings
Not applicable.

## Contributor checklist
- [x] File open/save changes include an appropriate round-trip or fidelity test. (Not applicable - ingest caps only, small files byte-identical.)


merrick-2002:
Thanks for the series today — most of it is in. We're going to close this one rather than merge it, for two reasons on our side:

- Every caller of `parseFileToText` (docs, sheets, slides, html attachment paths) already gates at `ATTACHMENT_MAX_BYTES` (50 MB) with a localized error, so the 20 MB check here lowers the effective limit and surfaces an unlocalized message.
- The attachment reader IPC is deliberately paged (`offset` / `maxChars`, returns `totalChars`) so the model can read a long document in slices. Truncating the extracted text at 200k chars inside the shared parser silently cuts everything past roughly page 100 of a dense PDF.

If you want to bound extraction time, the useful shape would be a page budget the caller can see and continue from, not a hard cut inside the parser.

merrick-2002:
Reopening — bounding extraction time is a real need, so let's shape this so it fits how the callers use the parser:

- Drop the 20 MB byte check here: all four callers already gate at `ATTACHMENT_MAX_BYTES` (50 MB) with a localized message, so the shared parser shouldn't lower it.
- Instead of truncating the extracted text to 200k chars inside `parseFileToText`, expose a page/segment budget the caller controls and can continue from (the attachment reader IPC already pages with `offset` / `maxChars` / `totalChars`, so a hard cut inside the parser silently loses everything past it).
- Surface "stopped at page N of M" in the result rather than an English error string, so the apps can localize it.

If that shape works for you, #568 can be folded in here as the pdf side of the same budget. Happy to merge once it's reworked.

merrick-2002:
Closing this for now to keep the queue tidy. The review above still stands: drop the 20 MB byte check (all four callers already gate at `ATTACHMENT_MAX_BYTES`), and replace the 200k-char hard cut inside `parseFileToText` with a page/segment budget the caller controls and can continue from, surfacing "stopped at page N of M" in the result. If you rework it along those lines, please open a new PR (or reopen this one) and we will pick it up promptly. Thanks for the work.

PR #548, opened 2026-09-20
fix(sheets): bound CSV import size before conversion

aniruddhaadak80:
## Summary
- What changed? Added MAX_CSV_IMPORT_BYTES 32MB pre-stat check in sheets open path before readFile+decodeCsvBuffer+csvToXlsxBuffer, plus cap test.
- Why? Unbounded readFile on attacker-controlled CSV (email/double-click) OOMs main before sidecar limits; brings parity with ATTACHMENT_MAX_BYTES and image caps.

## Related issue
Closes #

## Validation
- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test
Added csv-import-cap.test.ts. Full CI will verify.

## Screenshots or recordings
Not applicable.

## Contributor checklist
- [x] File open/save changes include an appropriate round-trip or fidelity test. (Guard only, no behavior change under cap.)


merrick-2002:

chatjesus:
Public-sample regression inputs for this fix and #553 (macOS 15.6.1 arm64):

- Microsoft OfficeComprehensionBenchmark Stop_Data_2019_to_2022.csv (109,365,743 bytes) caused `RangeError: Invalid string length` at the older importer's worksheet-XML Array.join during local testing. Independently reproduced with a 4-GiB Node heap.
- Food_Inspections_modified.csv (324,775,585 bytes) caused an Electron main-process V8 heap-out-of-memory / SIGTRAP in local testing without an artificial Electron heap cap.

These observations are not a test of a current packaged GenOffice release. At upstream commit 84d5b8fb9349434cc293a8e5c10cc3b13035c07b, the file-open path has this 32-MiB pre-read guard, plus #553 row/column limits. Both fixtures exceed the byte cap; this is regression evidence for the existing mitigation, not an allegation that the current guarded desktop path still crashes on these files.

Public sample manifest: https://huggingface.co/datasets/microsoft/OfficeComprehensionBenchmark/tree/0104ac2f88348ecf1148781b85fdb6a189b59671

Suggested regression: call the actual file-open handler and verify oversized files are rejected before readFile/conversion, with the app alive and a clear size-limit message. The existing cap test mirrors the constant/predicate without exercising that handler. No current-upstream desktop binary was run for this follow-up.

PR #558, opened 2026-09-20
fix(cli): reject negative and fractional deck page indices

aniruddhaadak80:
## Summary
- What changed? Hardened pageInRange in deck.ts to require integer >=0, plus range test.
- Why? -1 passed (wrote 00.json), 1.5 passed; both corrupt staged deck flow.

## Related issue
Closes #

## Validation
- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test
Added deck-page-range.test.ts. Full CI will verify.

## Screenshots or recordings
Not applicable.

## Contributor checklist


merrick-2002:
Thanks for checking this path. Both callers of `pageInRange` declare `index: z.number().int().nonnegative()` in their MCP tool schema, and the SDK validates arguments before the handler runs, so negative and fractional indices can't reach it. The added test also exercises a local `inRange` copy rather than `pageInRange`. Closing as not needed.

PR #567, opened 2026-09-21
fix(pptx-engine): land non-finite font sizes on lower bound

aniruddhaadak80:
## Summary
- What changed? szAttr now maps non-finite pt to the 1pt lower bound (exported for tests) with regression tests.
- Why? Math.max/min propagate NaN, so a non-finite font size serialized as sz=NaN corrupt OOXML; sibling clampInt already documents lower-bound landing.

## Related issue
Closes #

## Validation
- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test
Added szattr-bounds.test.ts. Full CI will verify.

## Screenshots or recordings
Not applicable.

## Contributor checklist
- [x] File open/save changes include an appropriate round-trip or fidelity test. (Validation-only, prevents corrupt sz.)


merrick-2002:

PR #568, opened 2026-09-21
fix(file-parse): stop pdf page iteration above truncation cap

aniruddhaadak80:
## Summary
- What changed? pdfToText breaks page iteration once raw chars exceed 1M (MAX_PDF_RAW_CHARS), far above the 200k downstream truncation cap. Invariant test added.
- Why? A 10k-page PDF paid full pdfjs text-content cost even though output is truncated anyway; early break keeps parsed output byte-identical while bounding CPU.

## Related issue
Closes #

## Validation
- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test
Added cap-invariant test. Full CI will verify.

## Screenshots or recordings
Not applicable.

## Contributor checklist
- [x] File open/save changes include an appropriate round-trip or fidelity test. (Output identical by construction; small PDFs untouched.)


merrick-2002:
Closing together with #546: the test imports `MAX_EXTRACTED_CHARS`, which only exists in that PR, so the `test` job is red on its own, and the premise (downstream truncation making later pages unobservable) doesn't hold on main — the paged attachment reader serves those pages. The early `break` would silently drop the tail of a 300+ page PDF. Thanks for looking at this path though; the pdf extraction is worth bounding, just as a page budget the caller controls.

PR #569, opened 2026-09-21
fix(docx-engine): sanitize non-finite ink geometry before writing

aniruddhaadak80:
## Summary
- What changed? anchoredInkRunXml sanitizes non-finite px inputs (fallback 1 for extents, 0 for offsets) since Math.max/min propagate NaN. Test added.
- Why? A NaN measurement would serialize as cx=NaN corrupt OOXML; offsets had no floor at all.

## Related issue
Closes #

## Validation
- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test
Added NaN-geometry test. Full CI will verify.

## Screenshots or recordings
Not applicable.

## Contributor checklist
- [x] File open/save changes include an appropriate round-trip or fidelity test. (Validation-only; finite inputs byte-identical.)


merrick-2002:

PR #575, opened 2026-09-21
fix(project-store): bound pending opening messages with early flush

aniruddhaadak80:
## Summary
- What changed? Opening-message buffer capped at 200 per chat; overflow materializes the JSONL file early instead of growing memory. Test added.
- Why? Thousands of pre-first-reply user messages accumulated unboundedly in memory (each up to 200k chars). Early flush preserves every message with correct seq order.

## Related issue
Closes #

## Validation
- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test
Added pending-cap.test.ts (250 msgs, order + count). Full CI will verify.

## Screenshots or recordings
Not applicable.

## Contributor checklist


merrick-2002:

PR #586, opened 2026-09-21
feat: discover models for custom OpenAI-compatible endpoints

3DAV-Works:
## Summary

- **What changed?** The `custom` provider's model field in Settings → AI Model
  now offers a picker of the models the endpoint actually serves, instead of a
  free-text box you have to type a model id into. The main process asks the
  configured base URL for `GET {baseUrl}/models` and folds the ids it returns
  into the settings screen's provider catalog, which already switches that row
  from a text box to a dropdown whenever the entry has models.
- **Why is this change needed?** `codex` already discovers its models live,
  through the app-server's `model/list`. The `custom` provider — the
  `needsBaseUrl` slot, pointed at OMLX, Ollama, LM Studio, vLLM or OpenRouter —
  did not, so a user had to know and type exact model ids, and a typo shows up
  only as a failed chat turn.

Notes on the design, for review:

- **Vendor-blind.** One request to the one endpoint every OpenAI-compatible
  server implements. Nothing sniffs for Ollama or LM Studio, and a server that
  does not answer simply keeps the free-text box.
- **Silent probe, not a chat turn.** It deliberately does not go through
  `aiFetch`: that wrapper retries through the Electron rescue fetch and warns on
  a network failure, which is right for a turn the user is waiting on and wrong
  for a background probe of a local server that is very often just not running.
  It does keep the user agent `aiFetch` would have added, since a gateway that
  refuses Node's bare `node` would otherwise leave the picker silently empty
  while chat itself worked. Refused, unreachable, slow, oversized, non-JSON and
  oddly shaped answers all return an empty list, and `listCustomModels` never
  throws.
- **No default model is guessed.** `/models` has no `isDefault` marker, and real
  lists mix embedding and conversion models in with the chat ones, so the first
  entry can be actively wrong. `defaultModel` stays `''` and the picker opens
  with nothing selected. A model the user typed by hand is pinned to the top of
  the list so it never disappears.
- **The settings screen writes the catalog, never the settings.** That is the
  existing rule for `refreshCodexModels`, and it matters here: `updateConfig`
  rebuilds the settings object from the render that captured it, so a write
  landing from a reply that resolves seconds later would revert whatever the
  user typed in the meantime. Two tests pin exactly that.
- One debounced effect (400 ms), keyed on the catalog entry's `needsBaseUrl`
  flag rather than on the literal `custom` id, so it follows the slot rather
  than the name and stays a no-op while any other provider is selected. React's
  own cleanup drops a superseded reply.
- 5 s timeout via the existing `createStreamWatchdog`. The body is read chunk by
  chunk against a 2 MB cap and the stream is cancelled once it is passed, since
  a chunked answer declares no length and the declared-length check alone would
  let an endpoint stream unbounded into memory. At most 2000 ids are kept; ids
  past that are dropped, with no UI to raise the limit in this PR.
  `Authorization: Bearer …` is sent only when a key is configured, since an
  empty `Bearer ` is a malformed credential that strict servers reject and local
  servers usually need no key at all.
- The return type reuses `CodexModelCatalog`, the `{ models, defaultModel }`
  shape the preload and `HomeApi` already speak, to keep the diff small. Happy
  to follow up with a neutral `ModelCatalog` alias if you would prefer one.

## Related issue

None; this follows on from #229 (allow keyless custom endpoints for local
models), which made keyless local endpoints usable but still required the model
id to be typed by hand.

## Validation

- [x] `npm run format:check` — "All matched files use Prettier code style!"
- [x] `npm run lint` — 0 errors, 13 warnings, all pre-existing in `apps/sheets`
      and `apps/slides` and none in a file this PR touches
- [x] `npm run typecheck` — clean
- [x] `npm test -w @genoffice/ai-provider` — Test Files 20 passed (20), Tests
      270 passed (270), including 22 new cases in
      `packages/ai-provider/tests/custom-models.test.ts`
- [x] `npm test -w @genoffice/shell` — Test Files 46 passed (46), Tests 482
      passed (482), including 20 new cases in
      `apps/shell/tests/settings-custom-models.test.ts`

List any checks not run and explain why: the root `npm test` across every
workspace was not run; this PR touches only `@genoffice/ai-provider`,
`@genoffice/shell` and one IPC registration in the docs main process, and both
affected suites are green in full.

The provider tests use verbatim `/models` bodies recorded from real servers
(Apple MLX/OMLX, Ollama 0.20.5, LM Studio) plus the real 401 body OMLX returns
with no key, so the parsing is pinned against wire shapes that actually occur.

## Screenshots or recordings

Settings → AI Model → Custom shows a model picker once a base URL answers
`/models`, and the free-text box otherwise.

The Custom provider on a local OpenAI-compatible endpoint, with the Model row as a picker:

Settings → AI Model → Custom, model row shown as a picker

The same picker open, listing the ids the server returned (chat, vision and embedding models side by side, which is why no default is guessed):

The model picker open, listing the ids returned by /models

## Contributor checklist

- [x] User-facing strings use the existing i18n resources. (No new strings: the
      picker/text-box switch already exists and is driven by `meta.models.length`.)
      (Not applicable — no file format is touched.)


merrick-2002:

Issue #166, opened 2026-08-29
本地模型功能一直报：由于工具调用次数限制，我需要高效。


Childish1jin:
### App

GenOffice Docs (docs)

### Problem

<img width="468" height="1170" alt="Image" src="https://github.com/user-attachments/assets/91968888-c3be-473e-90c7-17e75f27f0f2" />

merrick-2002:
Thanks for the report and the screenshot — this is fixed in [v0.8.667](https://github.com/genspark-ai/genoffice/releases/tag/v0.8.667).

We found three underlying problems, all of which hit local/BYOK models hardest:

- When a run hit the tool-call turn limit, the internal "turn limit reached" note was left in the conversation history forever, so every later request kept telling the model that tools were forbidden. Weaker models then echo that note verbatim — which is exactly what your screenshot shows. The note is now removed once the run finishes.
- There was no protection against a model repeating the exact same turn over and over; runs now detect a degenerate loop and stop instead of spinning.
- Some OpenAI-compatible local servers resend the full tool name on every streaming chunk, which corrupted tool names (e.g. `read_fileread_file`), made every call fail, and left the model looping on the errors. This is now handled.

Please update to v0.8.667 and let us know if you still see the loop with your local model.

merrick-2002:
Closing this as the fixes shipped in [v0.8.667](https://github.com/genspark-ai/genoffice/releases/tag/v0.8.667) (see the comment above for details). If you still hit the loop after updating, feel free to reopen with the details and we'll take another look.

PR #587, opened 2026-09-21
fix(docs): report matched-but-unchanged blocks honestly in AI command summary

3DAV-Works:
## Summary

- **What changed?** The docs AI assistant builds a one-line summary after every
  `apply_ops` batch. When nothing was changed, it said *"No matching blocks; the
  document was not changed."* even in the case where blocks **were** matched and
  simply already had the requested state. `summarize()` in
  `apps/docs/src/renderer/ai/ops.ts` now distinguishes the two cases: if the
  batch changed nothing, skipped nothing as protected, but matched blocks, it
  returns a new string `aiCmdNoneUnchanged` instead. The key is added to all 20
  docs AI locale shards, and three vitest cases cover the new branch, the
unchanged `aiCmdNone` branch, and a multi-op batch with tracked deletions.

- **Why is this change needed?** The old sentence was a lie in the common
  "already done" case, and the model acts on it. Ask it to *make the headings
  bold* in a document whose headings are already bold and it is told there were
  no matching blocks, so it concludes its selector was wrong and retries with a
  different one — sometimes several times. That wastes turns, and on smaller
  local models it can loop. Telling it the blocks matched and were left
  unchanged ends the exchange. The new sentence deliberately states only what
  happened, not why: `matched` means different things per op (for example
  `setImageProperties` counts non-image blocks as matched, and `setList`
  refuses headings), so any claim about the blocks "already being" in the
  requested state would be false for some of them.

Before and after, for "make the headings bold" on a document whose two headings
are already bold:

| | Summary the model is shown |
| --- | --- |
| Before | `No matching blocks; the document was not changed.` |
| After | `2 matching block(s) were left unchanged; the document was not changed.` |

A genuinely non-matching selector still yields the original `aiCmdNone` string,
and the protected-block case (`aiCmdNoneSkipped`) is untouched.

One subtlety worth flagging: `OpResult` carries an optional `skippedDeleted`,
counting targets skipped because their text is a pending tracked deletion.
Those are subtracted from each result's matched count, because such a block was
untouchable rather than left unchanged on its own merits, and the subtraction is
clamped at zero per result. The clamp matters: `deleteBlocks` is the only op
whose `matched` includes its tracked-deleted targets, while `findReplace` and
`setMatchedFont` count *hits* that `matched` already excludes, so an unclamped
sum could go negative and resurrect the very message this PR removes.

## Related issue

Related: #166 (closed)

## Validation

- [x] `npm run lint` — `✖ 13 problems (0 errors, 13 warnings)` (all 13 warnings
      pre-existing on `main`, none in the touched files)
- [x] `npm run typecheck` — clean, exit 0
- [x] `npm test -w @genoffice/docs` — `Test Files  258 passed (258)` /
      `Tests  2496 passed (2496)`

List any checks not run and explain why: the root `npm test` across every
workspace was not run; the change is confined to `apps/docs`, so the docs
workspace suite was run in full instead.

The new tests bite: with the `summarize()` change reverted and the locale keys
left in place, `ai-noop-summary.test.ts` fails with
`expected 'No matching blocks; the document was …' to contain '2 matching block(s) were left unchang…'`,
and removing just the per-result clamp fails the multi-op tracked-deletion case
the same way.

## Screenshots or recordings

Not applicable — text-only change to an assistant status string.

## Contributor checklist

      (Not applicable — no docx read or write path is touched.)


merrick-2002:

PR #588, opened 2026-09-21
fix(sheets): repair cross-sheet references IronCalc imports unquoted

3DAV-Works:
## Summary

- **What changed?** A worksheet named with an `&` — `R&D`, `P&L`, both common
  in finance — loses every reference to it when the workbook is loaded for
  recalculation. The sidecar now repairs those references after import, and
  the command line no longer describes a formula it could not parse as one
  using an unsupported function.

  Given a workbook whose `Summary!C23` is `='R&D'!C6`, where the R&D sheet's
  C6 is 59.5:

  | | Before | After | Excel and the app's grid |
  | --- | --- | --- | --- |
  | `sheet read --where formula` | `0` | `59.5` | `59.5` |

  `Summary!C24` is `SUM(C16:C23)`, so the wrong value was also silently
  feeding a total. Renaming the sheet to `RD` made everything correct, and
  every other cross-sheet reference in the same file, plus a SUMIF, was
  already right.

- **Why is this change needed?** The cause is in IronCalc, the engine the
  sidecar pins at 0.7.1. Its `name_needs_quoting` in `expressions/utils` is a
  denylist of `()'$,;-+{}` and space that omits `&`,
  so `quote_name` returns `R&D` bare. Because the importer converts every
  formula to R1C1 through it (`ironcalc/src/import/worksheets.rs:312-315`),
  the file's correct `'R&D'!C6` is stored as `R&D!C6`, which the lexer then
  reads as `R`, `&`, `D!C6`.

  IronCalc has since fixed this on its main branch
  (ironcalc/IronCalc#1422, merged 2026-09-17, switching to Excel's allowlist),
  but no published crate carries it yet: the latest release, 0.8.3, still has
  the denylist, and this repository is on 0.7.1. The repair below bridges
  that gap and can be deleted once the pinned IronCalc includes the fix; the
  `ampersand_sheet_references_resolve_across_sheets` test will keep passing
  either way.

  The fix here puts the quotes back. After `load_from_xlsx`, any formula
  whose parsed node is a parse error and whose stored R1C1 still carries a
  bare `Name!` for a sheet of this workbook is requoted and re-parsed with a
  sidecar-owned parser in R1C1 mode; the result replaces the stored node and
  text only if it parses cleanly. This restores what the file said rather
  than guessing at it: the text is the engine's own R1C1, the sheet is in the
  same workbook, text literals and already-quoted names are copied through
  untouched, and the longest matching name wins so `R&D` cannot truncate
  `R&D Notes`.

  A formula the requote cannot rescue no longer takes the file's cached value
  either. That substitution exists for external-workbook references, whose
  source really is unreachable; for a sheet in the same file a producer that
  writes no cached value (openpyxl writes an empty `<v/>`) imports as zero,
  so the old behaviour reported a number nobody computed. Those cells now
  keep their `#ERROR!`. External-workbook references keep their cached
  values exactly as before.

  Separately, the CLI treated both `#NAME?` and `#ERROR!` as "functions the
  local engine does not evaluate". Only `#NAME?` means that; `#ERROR!` means
  the formula would not parse. They now get separate warnings,
  `formulas_not_cached` and `formula_errors`.

## Related issue

None in this repository. Upstream engine fix: ironcalc/IronCalc#1422
(merged 2026-09-17, unreleased as of 2026-09-21).

## Validation

- [x] `npm run lint` — `13 problems (0 errors, 13 warnings)`; all 13 are
      pre-existing warnings in `apps/sheets` and `apps/slides` renderer files
      this branch does not touch.
- [x] `npm run typecheck` — clean.
- [x] `npm run native:test -w @genoffice/sheets` —
      `test result: ok. 192 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out`
- [x] `npm test -w @genoffice/cli` —
      `Tests  1 failed | 256 passed | 27 skipped (284)`
- [x] `cargo fmt --check` on the sidecar crate — clean, and `cargo clippy`
      reports nothing new in the changed code.

The one CLI failure is `tests/cli.test.ts:120`, "open fails with exit code 4
when no app binary is available". It fails identically with this branch's
changes stashed, so it is pre-existing and depends on how the app binary is
discovered on the machine.

Full `npm test` at the repository root was not run: it builds and exercises
every workspace including the Electron apps, and this branch touches the
xlsx sidecar and the CLI, whose suites are both above.

New tests in `apps/sheets/native/xlsx-engine/src/recalc.rs`, all against a
hand-built workbook written the way openpyxl writes one:

- `ampersand_sheet_references_resolve_across_sheets` — the 59.5 Excel gives.
- `every_reference_in_one_formula_is_requoted` — `'R&D'!A1+'R&D'!A2`, so the
  repair walks the whole formula rather than stopping at the first match.
- `an_irreparable_local_reference_is_not_pinned_to_an_invented_zero` — a
  malformed formula naming an in-workbook sheet keeps its `#ERROR!`.
- `quoting_covers_every_bare_reference_and_leaves_the_rest_alone` — the
  quoting rules, including text literals, external-workbook spellings and
  longest-name-wins.
- `external_link_formulas_keep_their_cached_values` — unchanged, still green.

Also checked by hand on the real 14-sheet, 43-formula openpyxl-written
workbook this was found in, by pointing the shipped CLI at the rebuilt
sidecar with `XLSX_SIDECAR_PATH`: the `'R&D'!C6` cell reads the value Excel
shows instead of 0, the `SUM` over it is correct, and every other value in
the sheet is unchanged. A CLI-created workbook with the same sheet
name, which previously read `#ERROR!`, now computes as well.

## Screenshots or recordings

Not applicable.

## Contributor checklist

- [x] User-facing strings use the existing i18n resources. (The two strings
      changed are CLI warning messages, which the CLI emits in English
      alongside a machine-readable `code`, as the surrounding code does.)
- [x] File open/save changes include an appropriate round-trip or fidelity
      test — see the list above.


merrick-2002:

PR #589, opened 2026-09-21
fix(ai-search): narrow data URL match groups (un-breaks sheets typecheck)

3DAV-Works:
## Summary

- **What changed?** `readMediaRef` in `packages/ai-search/src/media-tools.ts`
  read `m[1]` and `m[2]` straight off a `RegExp` match. It now destructures
  the two groups and guards both before use.
- **Why is this change needed?** `apps/sheets` typechecks with
  `noUncheckedIndexedAccess`, so on `main` since #544 `npm run typecheck`
  fails there with two `TS2532: Object is possibly 'undefined'` errors at
  `media-tools.ts` lines 114 and 118, and CI on `84d5b8f` is red. Behaviour
  is unchanged: a ref that does not match still throws the same error, and
  one that does decodes the same bytes and mime.

## Related issue

None. Follows #544; same shape as #521.

## Validation

- [x] `npm run typecheck -w @genoffice/sheets` — clean (was 2 errors)
- [x] `npm run typecheck -w @genoffice/ai-search` — clean
- [x] `npm test -w @genoffice/ai-search` — Test Files 5 passed (5), Tests 67 passed (67)
- [x] `prettier --check` on the touched file — clean

List any checks not run and explain why: the root `npm run lint` and full
`npm test` were not run for a one-file, six-line change; the touched
package's own suite and the two typechecks that matter are above.

## Screenshots or recordings

Not applicable.

## Contributor checklist

- [x] User-facing strings use the existing i18n resources. (No strings changed.)


merrick-2002:
Thanks for catching this. `main` went green with the 2026-09-21 sync, which carried the same destructuring fix for `readMediaRef`, so this branch is now conflicting and no longer needed. Closing — appreciate the quick fix.

PR #590, opened 2026-09-21
fix(cli): stop writing the styles part twice in new workbooks

3DAV-Works:
## Summary

- Every workbook written by `genoffice create --type xlsx` currently opens in
  Microsoft Excel (Mac and Windows) only after the repair dialog: "We found a
  problem with some content in '<name>.xlsx'. Do you want us to try to recover
  as much as we can?" Values survive the repair, but the prompt appears on
  every file.
- The cause is a duplicated styles part. Since #451 the blank workbook the CLI
  starts from (`blankXlsxBuffer` in `packages/xlsx-gateway/src/gateway/csv-import.ts`)
  ships `xl/styles.xml`, its `[Content_Types].xml` Override and an `rId3` styles
  Relationship. `blankWorkbook` in `packages/cli/src/formats/xlsx.ts` still
  appended the same three unconditionally, so the package ended up with two
  Relationships carrying `Id="rId3"` and two Overrides for `/xl/styles.xml`.
  Duplicate Relationship Ids and duplicate Override PartNames are invalid under
  the Open Packaging Conventions, which is what puts Excel into file-level
  repair.
- `blankWorkbook` now adds the part, the Override and the Relationship only
  when the blank does not already carry them, and picks the Relationship Id as
  max + 1 instead of the literal `rId3`. That is the pattern
  `addDefaultStylesheet` already uses in the xlsx gateway. The additions were
  kept as a guard rather than removed, so the CLI still writes a valid package
  if the blank ever loses its stylesheet again.
- Adds two tests that unzip a created workbook (single-sheet and multi-sheet)
  and assert that Relationship Ids and Override PartNames are unique, that
  exactly one styles Relationship and one styles Override exist, and that
  `xl/styles.xml` is present.

## Related issue

Related: #451 — no issue has been filed for this regression; happy to open one
if you would prefer the report tracked separately.

## Validation

- [x] `npm run lint` — `✖ 13 problems (0 errors, 13 warnings)`; all 13 warnings
      are pre-existing `react-hooks/exhaustive-deps` warnings in renderer files
      under `apps/docs`, `apps/pdf`, `apps/sheets` and `apps/slides`, none of
      which this change touches.
- [x] `npm run typecheck` — clean.
- [x] `npm test -w @genoffice/cli` — `Tests 1 failed | 258 passed | 27 skipped (286)`.
      The one failure is `tests/cli.test.ts:120`, which expects exit code 4 when
      `GENOFFICE_APP_BIN` points at a missing app binary. It fails identically on
      unmodified `main` in this environment, and does not touch xlsx.
      `tests/xlsx.test.ts` on its own: `Tests 9 passed | 13 skipped (22)` (the
      skips require the xlsx sidecar binary, which is not built here).
- [x] The new tests were confirmed to bite: with the `blankWorkbook` change
      reverted and the tests kept, both fail on the duplicate, e.g.
      `expected [ 'rId1', 'rId2', 'rId3', 'rId3' ] to deeply equal [ 'rId1', 'rId2', 'rId3' ]`.
- [x] Manual check in Excel for Mac (which shares its file parser with Excel
      for Windows; the repair log is Excel's own file-level validation), on a three-sheet
      workbook built from JSON by the CLI: before the change, opening it raised
      the repair prompt; removing exactly the two duplicated entries by hand made
      it open cleanly; after the change, a freshly created workbook opens with no
      prompt and correct values.

Not run: the full root `npm test`, since the change is confined to the CLI
package.

## Screenshots or recordings

Not applicable.

## Contributor checklist

- [x] User-facing strings use the existing i18n resources. (No user-facing strings changed.)


merrick-2002:

PR #596, opened 2026-09-21
fix(docs): cap compare LCS matrix with linear fallback

aniruddhaadak80:
## Summary
- What changed? compareParagraphs uses a linear multiset-join fallback past a 4M-cell LCS budget. Tests extended.
- Why? The LCS matrix is quadratic: comparing two large documents allocated hundreds of millions of cells (renderer OOM/freeze). Small-document output is unchanged.

## Related issue
Closes #

## Validation
- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test
Extended compare.test.ts. Full CI will verify.

## Screenshots or recordings
Not applicable.

## Contributor checklist
- [x] File open/save changes include an appropriate round-trip or fidelity test. (Not applicable - compare view only.)


merrick-2002:
Thanks — the LCS matrix can indeed blow up on long documents. Two changes before we can take it: (1) the fallback kicks in at 4M cells (≈2000×2000 paragraphs), which real documents hit, and the unordered multiset diff loses `changed` pairing (edited paragraphs show as delete+insert). Please switch the matrix to `Uint16Array` rows (or a rolling two-row variant) and raise the budget to ~50M cells; (2) keep the current algorithm as the primary path, use the fallback only past that budget, and return an `approximate`/`truncated` flag so the UI can say so.

merrick-2002:
Closing this for now to keep the queue tidy. The review above still stands: keep the current LCS as the primary path on `Uint16Array` rows (or a rolling two-row variant) with a budget around 50M cells, use the fallback only past that budget, and return an `approximate`/`truncated` flag for the UI. The branch also conflicts with main now. If you rework it along those lines, please open a new PR (or reopen this one after rebasing) and we will pick it up promptly. Thanks for the work.

PR #597, opened 2026-09-21
fix(ui): cap findInText result count

aniruddhaadak80:
## Summary
- What changed? findInText stops collecting past 2000 offsets (exported MAX_FIND_RESULTS). New test.
- Why? A single-character find-all over a large document returned millions of offsets (renderer OOM/freeze on highlight rendering).

## Related issue
Closes #

## Validation
- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test
Added find-text.test.ts. Full CI will verify.

## Screenshots or recordings
Not applicable.

## Contributor checklist
- [x] File open/save changes include an appropriate round-trip or fidelity test. (Not applicable - find UI only.)


merrick-2002:
Thanks. Capping at 2000 hits per paragraph puts the limit at the wrong layer: markdown `replaceAll` would only replace the first 2000 occurrences and the count would be wrong, while the real cost is in the decoration layer. Please move the cap to the total in `findMatches` (with a `truncated` marker the UI can show) and make `replaceAll` loop until no matches remain rather than relying on one capped pass.

merrick-2002:
Closing this for now to keep the queue tidy. The review above still stands: the cap belongs on the total in `findMatches` (with a `truncated` marker the UI can show), and `replaceAll` needs to loop until no matches remain instead of relying on one capped pass. If you rework it along those lines, please open a new PR (or reopen this one) and we will pick it up promptly. Thanks for the work.

PR #603, opened 2026-09-21
fix(agent-core): validate IPC stream chunk shape and size

aniruddhaadak80:
## Summary
- What changed? IPC transport truncates delta/reasoning text to 256k and drops tool-calls missing string id/name or a non-object input. Tests added.
- Why? The main-to-renderer bridge is a trust boundary: a malformed giant delta OOMed the transcript and a malformed tool-call poisoned the tool loop.

## Related issue
Closes #

## Validation
- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test
Extended electron-transport.test.ts. Full CI will verify.

## Screenshots or recordings
Not applicable.

## Contributor checklist
- [x] File open/save changes include an appropriate round-trip or fidelity test. (Not applicable - transport only.)


merrick-2002:
Thanks. Looking at where these chunks come from: the main→renderer stream carries our own provider adapters' output (docs-main / sheets-main / slides ai-ipc forward `AgentToolCall` verbatim), so this is not an untrusted boundary. Truncating a delta or dropping a tool call here would silently corrupt content or stall the agent loop instead of surfacing a bug, and the validator also accepts `input === undefined` although the type requires it. Closing as not-a-fix — if you find a path where a malformed chunk actually reaches the renderer, a repro would be very welcome.

PR #604, opened 2026-09-21
fix(slides): allowlist hyperlink schemes at set-link, follow, and export

aniruddhaadak80:
## Summary
- What changed? New shared safeExternalUrl (http/https/mailto only) gates all three URL layers: decodeLinkTarget drops dangerous schemes at parse, slides:set-link rejects them before deck write, slideshow follow and PDF-export hrefs filter them. Tests added.
- Why? File-authored javascript:/file: URLs previously flowed into window.open (system browser) and exported PDF /URI actions - a phishing/drive-by vector from hostile decks.

## Related issue
Closes #

## Validation
- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test
Extended export-links.test.ts (export + decode layers). Full CI will verify.

## Screenshots or recordings
Not applicable.

## Contributor checklist
- [x] File open/save changes include an appropriate round-trip or fidelity test. (Not applicable - link targets only.)


merrick-2002:
Thanks — gating `slides:set-link` and `exportLinkHref` is right. Two changes: (1) please revert the `decodeLinkTarget` change: `TextEditOverlay` calls it on every edit commit to rebuild run links, so any file-authored relative link (`../report.pdf`), scheme-less `www.x.com`, `tel:` or `ftp:` link would be silently stripped from the deck when the user edits that text box — parsing must never drop links; (2) drop the new `safeExternalUrl` in `run-link.ts` and import the one from `@genoffice/electron-utils` (the renderer already imports that package), and skip the `SlideShowView` guard — `setWindowOpenHandler` in slides-main already routes `window.open` through the suite gate. The e2e failure (`mcp-headless.spec.ts`) looks like a runner flake unrelated to this diff.

merrick-2002:

PR #611, opened 2026-09-21
fix(pdf2docx): sanitize charScale and charSpacing before OOXML emit

aniruddhaadak80:
## Summary
- What changed? Span charScale requires finite 0.5..2, charSpacing requires finite within 720pt; the rebuild emit skips non-finite values. Test added.
- Why? Hostile PDF glyph metrics flowed into w:w/w:spacing as Infinity/NaN (Infinity passes >0 checks), producing invalid OOXML and skewing fit aggregates.

## Related issue
Closes #

## Validation
- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test
Extended spans.test.ts. Full CI will verify.

## Screenshots or recordings
Not applicable.

## Contributor checklist
- [x] File open/save changes include an appropriate round-trip or fidelity test. (Covered: spans feed the rebuild path.)


merrick-2002:
Thanks. Keep the finiteness check, but the 0.5..2 window is below OOXML: `w:w` is legal from 1% to 600%, and condensed/stretched PDF text (`Tz` 30 or 300) is real — dropping it makes every rebuilt line wrap differently, which is exactly what the surrounding comment warns about. Please widen `charScale` to 0.01..6 and drop the duplicated finiteness guard on the rebuild side once spans.ts filters.

merrick-2002:

PR #616, opened 2026-09-21
fix(pptx-ops): cap unit-string lengths to valid EMU range

aniruddhaadak80:
## Summary
- What changed? parseLength clamps output to finite +/-50M EMU (returns undefined on non-finite). Tests added.
- Why? User-supplied op strings like 9999999999in produced ~9e15 EMU into xfrm and bypassed the numeric !(x>0) guards (Infinity>0 is true).

## Related issue
Closes #

## Validation
- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test
Extended units.test.ts. Full CI will verify.

## Screenshots or recordings
Not applicable.

## Contributor checklist
- [x] File open/save changes include an appropriate round-trip or fidelity test. (Not applicable - unit parse only.)


merrick-2002:
Thanks. Non-finite lengths are already rejected by `requireFinite` (registry.ts, used for offsets in element-ops / insert-ops), and xfrm is clamped to the ST_Coordinate ceiling at serialization (`emuAttr` in pptx-engine generate.ts). The 50M EMU cap is also below PowerPoint's 56in slide max (51,206,400 EMU), so `"56in"` would be silently rewritten to a wrong value. Closing — if a bound is wanted, the op validator is the place to reject out-of-range, not the unit parser.

PR #617, opened 2026-09-21
fix(docs): bound copy-image clipboard data-url decode

aniruddhaadak80:
## Summary
- What changed? New copy-image-guard module: data URLs capped at ~15MB with raster-subtype allowlist (lazy-media ids still pass), display-size overrides clamped finite. Handler uses both. Tests added.
- Why? The clipboard bitmap path base64-decoded unbounded renderer payloads in the main process (decode bomb) and accepted any data: subtype; meta dims passed Infinity into pasted  attrs.

## Related issue
Closes #

## Validation
- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test
Added copy-image-guard.test.ts (electron-free module). Full CI will verify.

## Screenshots or recordings
Not applicable.

## Contributor checklist
- [x] File open/save changes include an appropriate round-trip or fidelity test. (Not applicable - clipboard only.)


merrick-2002:
Thanks — the subtype allowlist and the finite-dimension clamp are good. The 15 MB cap is too low: the renderer re-encodes protected images via `canvas.toDataURL('image/png')` and freshly pasted photos are data URLs, so a 4000×3000 PNG easily exceeds it — and because the renderer has already `preventDefault()`ed, a `false` here leaves the clipboard empty with no message. Please raise `MAX_COPY_IMAGE_DATA_URL` to at least the 40 MB Slides uses (`element-clipboard.ts`), ideally 64 MB.

merrick-2002:

PR #620, opened 2026-09-21
fix(slides): bound external-text paste paragraph expansion

aniruddhaadak80:
## Summary
- What changed? External clipboard text is capped (48k chars, 2000 paragraphs) via pasteParagraphs before building the addElement op. Tests added.
- Why? OS clipboard text is attacker-influenced: a multi-MB paste previously expanded unbounded into IPC paragraphs and engine content.

## Related issue
Closes #

## Validation
- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test
Extended clipboard-actions.test.ts. Full CI will verify.

## Screenshots or recordings
Not applicable.

## Contributor checklist
- [x] File open/save changes include an appropriate round-trip or fidelity test. (Not applicable - paste path only.)


merrick-2002:
Thanks. PowerPoint has no paste limit and the OS clipboard is the user's own, so silently truncating at 48 000 chars / 2000 paragraphs loses content with no signal. Please either remove the truncation or make it generous (≥ 1M chars / 50k paragraphs) *and* surface a status message when it trims.

merrick-2002:

PR #621, opened 2026-09-21
fix(slides): cap TIFF decode dimensions before pixel allocation

aniruddhaadak80:
## Summary
- What changed? tiffToPng checks header dims (64MP / 16384px per side) before decodeImage/toRGBA8 and only decodes the chosen in-budget page. Test added.
- Why? A crafted ppt/media TIFF with giant IFD dims previously allocated w*h*4 bytes (tens of GB) in the main process; it now fails closed to the blank placeholder.

## Related issue
Closes #

## Validation
- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test
Extended tiff-decode.test.ts (synthetic giant-IFD fixture). Full CI will verify.

## Screenshots or recordings
Not applicable.

## Contributor checklist
- [x] File open/save changes include an appropriate round-trip or fidelity test. (Not applicable - display transcode only; original bytes preserved.)


merrick-2002:
Thanks — checking the IFD header before decoding and decoding a single page is a real improvement (the current code decodes every IFD). Two changes: (1) 64 MP rejects real scans (a 600 dpi A3 page is ~69 MP); please raise to ~256 MP (1 GB RGBA) and 32767 px/side; (2) if `ifds[0]` is over budget the function returns null immediately instead of falling through to a smaller in-budget page, contradicting the loop's `continue` — initialise `page = null` and pick the first in-budget IFD.

merrick-2002:

PR #658, opened 2026-09-21
fix(docs): cap save-merged-pdf fragments and part size

### LINKED ISSUE #624: docs: save-merged-pdf accepts unbounded base64Parts (OOM via IPC)
## Problem
The docs:save-merged-pdf handler iterates renderer-supplied base64Parts with no array check or size cap. A compromised renderer (or hostile AI input) can OOM the main process or hang the merge with thousands of parts.

## Evidence
- apps/docs/src/main/docs-main.ts:4096 handler entry
- apps/docs/src/main/docs-main.ts:4115 for (const b64 of base64Parts) with no validation

## Suggested fix
Require Array.isArray, cap part count (e.g. <=200) and per-part bytes, reject otherwise with an i18n error before PDFDocument.load.

## Acceptance
- Oversized/non-array input is rejected without loading pdf-lib; normal merges unaffected.

aniruddhaadak80:
## Summary
- What changed? docs:save-merged-pdf requires a bounded fragment array (1..200 parts, each <=50MB chars) via a new electron-free merged-pdf-guard module; violations return errFileTooLarge before pdf-lib loads. Tests added.
- Why? The renderer-supplied list was iterated unchecked: a hostile payload could OOM the main process or hang the merge.

Closes #624.

## Related issue
Closes #624

## Validation
- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test
Added merged-pdf-guard.test.ts. Full CI will verify.

## Screenshots or recordings
Not applicable.

## Contributor checklist
- [x] File open/save changes include an appropriate round-trip or fidelity test. (Not applicable - merge channel only.)


merrick-2002:
Thanks. The renderer chunks print groups at `PRINT_CHUNK = 10` pages (`file-actions.ts`, 5 on the fallback path), so a 2000-page document produces more than 200 parts and PDF export would fail with a misleading 'File exceeds the size limit'. Please cap the *total* base64 size (e.g. ≤ 512 MB, matching `MAX_OPEN_BYTES`) instead of the part count, or raise `MAX_MERGED_PDF_PARTS` to ≥ 10 000; keep the `Array.isArray` / string checks.

merrick-2002:
Thanks. Closing along with #624: this channel only carries the renderer's own 10-page `printPdfBuffer` chunks, so there is no untrusted part list to cap (details there).

PR #659, opened 2026-09-21
fix(docs): validate export-pdf page size and scale

### LINKED ISSUE #625: docs: export-pdf trusts non-finite page size
## Problem
The docs:export-pdf handler divides renderer-supplied twips straight into printToPDF pageSize with no finite/range check. NaN/Infinity/0 sizes (reachable from corrupt section geometry via parseFloat) crash or hang the export.

## Evidence
- apps/docs/src/main/docs-main.ts:3963 pageSize width/height from raw twips
- apps/docs/src/renderer/file-actions.ts:1236 parseFloat path producing NaN

## Suggested fix
Number.isFinite + sane range (e.g. 1440-31680 twips), else return { ok: false, error }.

## Acceptance
- NaN/Infinity/0 sizes return a clean error without calling printToPDF.

aniruddhaadak80:
## Summary
- What changed? docs:export-pdf validates page geometry (finite 1440..31680 twips) and scale (0.1..5) via a new electron-free export-pdf-guard module, returning a clean error before printToPDF. Tests added.
- Why? Corrupt section sizes arrive as NaN/Infinity and previously crashed or hung the export inside Chromium.

Closes #625.

## Related issue
Closes #625

## Validation
- [ ] npm run format:check
- [ ] npm run lint
- [ ] npm run typecheck
- [ ] npm test
Added export-pdf-guard.test.ts. Full CI will verify.

## Screenshots or recordings
Not applicable.

## Contributor checklist
- [x] File open/save changes include an appropriate round-trip or fidelity test. (Not applicable - export channel only.)


merrick-2002:
Thanks. Three changes: (1) the lower bound must be Word's 0.1in (144 twips), not 1in — label/receipt page sizes are legal; (2) the guard reuses `errFileTooLarge` ('File exceeds the size limit') for a geometry error — please add a dedicated message or a plain English string like the sibling 'export target is not an authorized path'; (3) `docs:print-pdf-buffer` takes the same params and is left unguarded — please apply the same validation there (or share the helper). Note `printToPDF` is already inside try/catch, so NaN geometry already returns `{ok:false}` today.

merrick-2002:
Thanks. Closing along with #625: non-finite sizes cannot reach `printToPDF` (the section parser falls back to defaults) and zero/negative sizes are already rejected by Electron inside the existing try/catch (details there).

